PR #34, opened 2015-07-20
Error in 'dump_version' now correctly raised.

mr-rodgers:
This error occurred because the function wasn't correctly extracting the file extension from a filename while trying to raise an Exception, and so the descriptive error never got printed.

This fix just makes sure that the exception is correctly raised.

### DIFF
--- a/testing/test_functions.py
+++ b/testing/test_functions.py
@@ -1,5 +1,8 @@
+import os.path
 import pytest
 import pkg_resources
+import tempfile
+from setuptools_scm import dump_version
 from setuptools_scm.version import guess_next_version, meta, format_version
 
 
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt:
good find!!

 got an idea for a good test? else i'll have to make one before merging this one

mr-rodgers:
I added a test that checks for whether the DumpVersion raises ValueError (expected) on the error condition.

testing/test_functions.py
```diff
@@ -47,3 +50,18 @@ def test_format_version(version, monkeypatch, scheme, expected):
         version,
         version_scheme=vs,
         local_scheme=ls) == expected
+
+
+@pytest.fixture
+def tmpfoo(request):
```
RonnyPfannschmidt: good start,

please use the `tmpdir` fixture as base,
you can then use `tmpdir.strpath` as `root` and a constant name for the `write_to` variable

PR #35, opened 2015-07-23
update project url to reflect move to github

davidszotten:


RonnyPfannschmidt:
good find, that one was completely missed

PR #54, opened 2015-09-17
Do not use stdout for reporting wrong setuptools

### LINKED ISSUE #48: Warning "setuptools_scm functionality is degraded" over stdout
The warning "setuptools_scm functionality is degraded" in setuptools_scm/setuptools_scm/version.py is printed to stdout, I rather think that it should rather be written to stderr, so that shell scripts will not break due to unexpected content in stdout.

Would `warnings.warn(....)` be suitable?

HolgerPeters:
This PR should fix #48, instead of printing the error out to stdout, distutils warning facility is used.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,12 @@
+
+
+v1.8.0
+======
+
+* fix issue with setuptools wrong version warnings being printed to standard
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt:
well done, could you please add a changelog entry as finishing touch

Issue #62, opened 2015-11-10
Assert when installing on or after non-version git tag


c1728p9:
If I checkout an existing repo (In this case pyOCD) and create a non-version formatted tag, I get an assert when installing.  If I don't don't have a local tag then installing works correctly.  Is it a requirement when using setuptools_scm that the git repository can only have version formatted tags?

Machine: Windows 7 64 bit
setuptools-scm: Version 1.8.0
pyOCD commit: 87ca6b91477db7d458960504f2103dae9b7a319f

The sequence to reproduce is shown below.  If "git tag example_tag" is removed then the install succeeds.

```
cd c:\
git clone https://github.com/mbedmicro/pyOCD.git
cd pyOCD
git tag example_tag
virtualenv env
call env\Scripts\activate
pip install setuptools_scm
python setup.py install
```

Output:

```
Traceback (most recent call last):
  File "setup.py", line 66, in <module>
    include_package_data=True,  # include files from MANIFEST.in
  File "c:\python2.7.10_64\Lib\distutils\core.py", line 111, in setup
    _setup_distribution = dist = klass(attrs)
  File "c:\pyOCD\env\lib\site-packages\setuptools\dist.py", line 272, in __init__
    _Distribution.__init__(self,attrs)
  File "c:\python2.7.10_64\Lib\distutils\dist.py", line 287, in __init__
    self.finalize_options()
  File "c:\pyOCD\env\lib\site-packages\setuptools\dist.py", line 327, in finalize_options
    ep.load()(self, ep.name, value)
  File "c:\pyOCD\env\lib\site-packages\setuptools_scm\integration.py", line 19,in version_keyword
    dist.metadata.version = get_version(**value)
  File "c:\pyOCD\env\lib\site-packages\setuptools_scm\__init__.py", line 67, inget_version
    version = version_from_scm(root)
  File "c:\pyOCD\env\lib\site-packages\setuptools_scm\__init__.py", line 29, inversion_from_scm
    return ep.load()(root)
  File "c:\pyOCD\env\lib\site-packages\setuptools_scm\git.py", line 36, in parse
    return meta(tag, dirty=dirty, node=node)
  File "c:\pyOCD\env\lib\site-packages\setuptools_scm\version.py", line 85, in meta
    assert tag is not None, 'cant parse version %s' % tag
AssertionError: cant parse version None
```

untitaker:
Yeah, that assumption is definetly made.

[...]

PR #63, opened 2015-11-12
Fix install failure when using non-version tag

c1728p9:
When installing a package with a non-version formatted git tag
(for example 'example_tag') as the as the most recent tag then
setuptools_scm will assert that it can't parse the tag.  This patch
fixes this assert by filtering out non-version formatted tags.  By
default any tag with a '.' in the name will be interpreted as a version
tag.  Optionally, if 'exact' is set to true when calling parse then
only tags in the form of 'v_._.*' will be found.

### DIFF
--- a/setuptools_scm/git.py
+++ b/setuptools_scm/git.py
@@ -6,7 +6,7 @@ from os.path import abspath, realpath
 FILES_COMMAND = 'git ls-files'
 
 
-def parse(root):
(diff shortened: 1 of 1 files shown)


c1728p9:
This fixes #62.

RonnyPfannschmidt:
this one got fixed differently

Issue #65, opened 2015-11-27
Support git archives


Changaco:
git can inject data into archives, including tags. Here's how it could work:
1. put `use_scm_version={'git_metadata': '$Format:%D$'}` in `setup.py`
2. put `setup.py  export-subst` in `.gitattributes`
3. `setuptools_scm` extracts the version number from `git_metadata` with a regexp like `\btag: ([0-9][^,]*)\b`

RonnyPfannschmidt:
The metadata would be incomplete and broken in many situations

Also in that case a file named .git_archival.txt should be used and registered as scm backend
I'm happy to include a Variant that demonstrably works correct with random archives from github

But I will not again spend time researching and implementing it for git

RonnyPfannschmidt:
will be solved by #67 which should document the plugin that was kindly implemented and released by the original reporter

PR #66, opened 2015-11-28
Support git archives

Changaco:
Implements #65.

### DIFF
--- a/.git_archival.txt
+++ b/.git_archival.txt
@@ -1,0 +1,1 @@
+ref-names: $Format:%D$
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+.git_archival.txt  export-subst
--- a/README.rst
+++ b/README.rst
@@ -10,8 +10,8 @@ It also handles file finders for the supported scm's.
 .. image:: https://travis-ci.org/pypa/setuptools_scm.svg?branch=master
     :target: https://travis-ci.org/pypa/setuptools_scm
 
-Setup.py usage
---------------
+Usage
+-----
 
 To use setuptools_scm just modify your project's setup.py file like this:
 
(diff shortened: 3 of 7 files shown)


RonnyPfannschmidt:
rejecting as its broken exactly like i explained in #65

Changaco:
Not working for untagged commits is not the same thing as being broken.

RonnyPfannschmidt:
unless it works for pip install towards a master branch its problematic to begin with and all tags can just as easily be released

Changaco:
Supporting git archives of tagged commits is useful (see https://github.com/Changaco/python-libarchive-c/issues/23). How is it better to have no support at all instead of partial support?

RonnyPfannschmidt:
releases can just be released, but having some archvies work and some archvies not is undesirable
in particular since some of the hash urls will work and some not

if all it can ever do is work with whats supposed to be official releases, just use actual releases, those install faster, are more reliable and certain to have all metadata

even the project you references talked about just using the pypi releases

and what i really really really don't want to cause a project is solving the guaranteed to come up support request to make it work for all archives instead of some

inconsistent support for archives is bound to create suffering for downstreams - providing consistent experience is very important

Changaco:
The user didn't want to use PyPI releases, that's why I went looking for a way to include version numbers into GitHub releases. If we can't integrate support for that in `setuptools_scm` then I'll have to stick with [version.py](https://github.com/Changaco/version.py/), and/or fork `setuptools_scm` so my work here doesn't go to waste.

RonnyPfannschmidt:
due to how setuptools_scm works, you can just create a plugins that does the same,
all you need to do is register a .git_archival.txt as entrypoint in that plugin,
and then have the plugin depend on setuptools_scm and use both as setup_requires

that way you can use the functionality for tagged archives,and still have the rest work normally

RonnyPfannschmidt:
i#d even be willing to document such a plugin as additional usage vector including a description of why its not in setuptools_scm itself

Changaco:
https://pypi.python.org/pypi/setuptools_scm_git_archive

RonnyPfannschmidt:
@Changaco well done, any suggestion how to refer to it from the docs, i#m thinking of a little reorg

Issue #73, opened 2016-01-04
Test failures on EL7


opoplawski:
I'm getting the following test failures on EL7:

```
_________________________ test_next_tag[1.1-1.2.dev0] __________________________

tag = '1.1', expected = '1.2.dev0'

    @pytest.mark.parametrize('tag, expected', [
        ('1.1', '1.2.dev0'),
        ('1.2.dev', '1.2.dev0'),
        ('1.1a2', '1.1a3.dev0'),
        ])
    def test_next_tag(tag, expected):
        version = pkg_resources.parse_version(tag)
>       assert guess_next_version(version, 0) == expected

testing/test_functions.py:19: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tag_version = ('00000001', '00000001', '*final'), distance = 0

    def guess_next_version(tag_version, distance):
        version = str(tag_version)
        if '.dev' in version:
            prefix, tail = version.rsplit('.dev', 1)
            assert tail == '0', 'own dev numbers are unsupported'
            return '%s.dev%s' % (prefix, distance)
        else:
>           prefix, tail = re.match('(.*?)(\d+)$', version).groups()
E           AttributeError: 'NoneType' object has no attribute 'groups'

../../../redhat/BUILDROOT/python-setuptools_scm-1.10.1-1.el7.x86_64/usr/lib/python2.7/site-packages/setuptools_scm/version.py:96: AttributeError
_______________________ test_next_tag[1.2.dev-1.2.dev0] ________________________

tag = '1.2.dev', expected = '1.2.dev0'

    @pytest.mark.parametrize('tag, expected', [
        ('1.1', '1.2.dev0'),
        ('1.2.dev', '1.2.dev0'),
        ('1.1a2', '1.1a3.dev0'),
        ])
    def test_next_tag(tag, expected):
        version = pkg_resources.parse_version(tag)
>       assert guess_next_version(version, 0) == expected

testing/test_functions.py:19: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tag_version = ('00000001', '00000002', '*@', '*final'), distance = 0

    def guess_next_version(tag_version, distance):
        version = str(tag_version)
        if '.dev' in version:
[...]

PR #76, opened 2016-01-14
Ensure we call tag_to_version

### LINKED ISSUE #61: Ensure we strip version prefixes on old setuptools
See #59 degradion shouldn't be that bad

brianmay:
Closes #61

So I found `tag_to_version` already does the right thing. However this function is not called because `parse_version` is None. `tag_to_version` already seems to be able to cope with `parse_version` being None, so this test seems like it isn't required.

Just removing that test produced an error however because `SetuptoolsVersion` is None. I am guessing this is because I am testing based on an old version of setuptools.

Only done some basic testing so far, however it appears to work:

```
# python ./setup.py --version 
important note:

the setup of setuptools_scm is self-using,
the first execution of `python setup.py egg_info`
will generate partial data
its critical to run `python setup.py egg_info`
once before running sdist or easy_install on a fresh checkouts

pip usage is recommended

your setuptools is too old (<12)
setuptools_scm functionality is degraded
1.10.1.dev2+ng77da109
```

### DIFF
--- a/setuptools_scm/version.py
+++ b/setuptools_scm/version.py
@@ -31,6 +31,7 @@ def callable_or_entrypoint(group, callable_or_name):
 def tag_to_version(tag):
     trace('tag', tag)
     # lstrip the v because of py2/py3 differences in setuptools
+    # also required for old versions of setuptools
     version = tag.rsplit('-', 1)[-1].lstrip('v')
     if parse_version is None:
         return version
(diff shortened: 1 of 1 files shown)


brianmay:
I expect the travis tests - if they ever actually start - will fail, however due to reasons I think unrelated to this change (see #73).

RonnyPfannschmidt:
Well done

Due to upcoming travel I might not be able to put this into a release before mid next week

The comment in tag to version should be extended since its a accident that it does the right thing

brianmay:
Ok, updated the comment now.

RonnyPfannschmidt:
Thanks :)

I'll do a release after returning from travel

PR #80, opened 2016-01-31
Added Python version classifiers to setup.py

agronholm:
This should fix the project incorrectly marked on Python 3 readiness as "not Python 3 compatible yet".


RonnyPfannschmidt:
Thanks for noticing and creating the PR

The failure is due to pip8

RonnyPfannschmidt:
And I can't merge from phone, will merge at home

Issue #103, opened 2016-10-05
Invalid syntax in generated version.py 


ghost:
A package I use is distributed through Mercurial and uses setuptools_scm with setup.py like this

``` python
from setuptools import setup
setup(
    use_scm_version={
        'write_to': 'PACKAGENAME/version.py',
    },
    setup_requires=['setuptools_scm'],
#...
)
```

The resulting version.py is not valid python:

``` python
# coding: utf-8
# file generated by setuptools_scm
# don't change, don't track in version control
version = <ScmVersion 0.4.0 d=None n=None d=False x={}>
```

The issue did not occur before today, it seems that it is caused by v1.12.0 release.

untitaker:
Same issue here.

RonnyPfannschmidt:
yikes, in response i pulled the release of pypi, i am now working on a fixup

untitaker:
BTW this reordering creates the issue: https://github.com/pypa/setuptools_scm/commit/80565e3364f86f33be05d7d07be427f480f10d6a#diff-cc3eb8241d67f6487179a3c8976e128eR107

`format_version`'s rv is not used for `dump_version`.

RonnyPfannschmidt:
i am aware, i'm testing the fix and i started Hungarian notation (parsed_version vs version_string) plus type assertions

anarcat:
ouch, sorry - my bad. i assumed that since tests passed, things were okay. looks like the test suite was incomplete!

RonnyPfannschmidt:
@anarcat @untitaker please take a look at the pr

untitaker:
Fix lgtm. No tests though :P

RonnyPfannschmidt:
i#ll fix in a different manner

RonnyPfannschmidt:
the result of parse will always be a parsed version now

RonnyPfannschmidt:
fixed in #104

PR #104, opened 2016-10-05
fix issue #103 - reorder version dumping

RonnyPfannschmidt:


setuptools_scm/__init__.py
```diff
@@ -37,6 +37,7 @@ def _version_from_entrypoint(root, entrypoint):
 
 
 def dump_version(root, version, write_to, template=None):
+    assert isinstance(version, string_types)
```
anarcat, line 41: does this detect the new failure?

setuptools_scm/__init__.py
```diff
@@ -101,18 +102,20 @@ def get_version(root='.',
...
             template=write_to_template)
-        if isinstance(version, string_types):
-            return version
-        version = format_version(
-            version,
-            version_scheme=version_scheme,
-            local_scheme=local_scheme)
-        return version
+
+        return version_string
```
anarcat, line 130: i like this: much clearer than the previous implementation - i thought of doing something like that in my own PR but refrained to keep things simple...

testing/test_functions.py
```diff
@@ -52,7 +52,7 @@ def test_format_version(version, monkeypatch, scheme, expected):
 
 def test_dump_version_doesnt_bail_on_value_error(tmpdir):
     write_to = "VERSION"
-    version = VERSIONS['exact']
+    version = str(VERSIONS['exact'].tag)
```
anarcat, line 55: not sure what isthis for - would this silence possible unit test failures that could detect the fixed bug?

anarcat (commented):
looks good, but i'm wondering if the test suite would find a similar regression in the future.

testing/test_functions.py
RonnyPfannschmidt: exact.tag is a setultools version not a string

setuptools_scm/__init__.py
RonnyPfannschmidt: i did a more intensive restructure ensuring i never get a string there - the shortcut was an premature optimization, and as such the root of the evil ^^

setuptools_scm/__init__.py
RonnyPfannschmidt: the code in dump-version only works with strings, perhaps even this check is not trough enough

RonnyPfannschmidt:
@anarcat @untitaker unless you have further comments i'm about to release 1.13

untitaker:
lgtm

PR #106, opened 2016-10-10
Fix #86

### LINKED ISSUE #86: Dirty working directory not being correctly detected when tags are missing
I'm working with a "big repo" with many libraries/services. One service, "tasks," references the "model" library in its requirements.txt

```
../lib/model
```

If I run

```
$ pex -r requirements.txt -o test.pex
```

Then the generated wheel in ~/.pex/build is called `model-0.1.dev239+n6e5cad8-py2-none-any.whl`, which doesn't have the timestamp that the documentation suggests, despite the fact that the repository is dirty:

```
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

    modified:   lib/model/setup.py
    modified:   workers/tasks/requirements.txt
```

RonnyPfannschmidt:


RonnyPfannschmidt:
different git versions behave differently wrt this
_argh_

Issue #24, opened 2015-07-01
Lack of documentation


jaraco:
This project is something I am looking for, but I am missing some documentation and/or place to discuss things. Personally, I would like to help with the development and/or documentation, so I share things that I have learned about the problem `setuptools_scm` is trying to address:

Connecting SCM with versioning stuff is difficult. There are two major use cases when you want the package to know its version:
- The moment when you release (release-time) the package and 
- the moment when an unsuspecting user installs it without any knowledge of SCM at all.

During the release-time, thigs are easy. It is not that difficult to guess the descriptive version string of the package from the SCM. This is mainly because the developer has all tools he needs and the source code is tagged.

However, when one releases the package, users that are going to install it don't have the advanced tools and the package should be well-aware of which version it is (a good practice is that package `foo` has `foo.__version__` attribute that is the semver version string). So, the version has to be somehow injected to the package and the package has to be able to pick it up.

Some quick examples:
- `numpy` writes the version they infer from `git` to `numpy/version.py` at release-time. However, they keep the version number in `setup.py` along some boilerplate code.
- `setuptools` maintain a `setuptools/version.py` containing a version string in SCM
- `matplotlib` maintains version string in `lib/matplotlib/__init__.py` and has relatively convoluted code that extracts it.

Generally, there are two ways of how to solve this:
1. Store the version string somewhere and introspect it from `setup.py` using more or less convoluted code you have to write yourself.
2. Generate the version string during release-time and inject it somehow to the package in a smart way.
   The cons of this approach is that it is an unusual approach and one can forget to perform the injection step. Having it as a hook of `setup.py [sb]dist` would be so much cleaner. 
   
   Also, there would be the ordering problem - first you tag, then you release and the release gets modified ex-post when it gets the right version string. So the package should be tagged afterwards or the part with the version string should not be revision-controled.

Now, let's get to `setuptools_scm`.
AFAIK, users will be obliged to get it when they get in touch with package that requires it, which I regard as sub-optimal, but if parts of it get integrated into `setuptools`, it would dissolve this weakness.
On the other hand, it seems to me that it is introspecting `PKG-INFO`, which I think is the right way to go. In my opinion, generating `PKG-INFO` is essentially the injection step and people in charge of their package's `setup.py` just have to have good possibilities of getting it out of the file.

So what would I like to see in this package:
[...]

PR #111, opened 2016-10-14
fix mercurial for several tags on changeset ...

petrem:
…and distance only consisting of tag commits

This is an attempt to fix a situation when you have several tags on the changeset. What is more, it attempts to detect and "discard" commits that are only tagging, regardless of distance.
I should probably add more tests...
Thanks for considering/commenting.

E.g.:

changeset:   174:8b9eeff53bde
user:        Petre Mierlutiu pmierlutiu@...
date:        Fri Oct 14 15:57:13 2016 +0300
summary:     Added tag 1.3.0.2 for changeset 9b5b4375432e

changeset:   173:a9df4646d79c
user:        Petre Mierlutiu pmierlutiu@...
date:        Fri Oct 14 14:25:14 2016 +0300
summary:     Added tag 1.3.0.1 for changeset 9b5b4375432e

changeset:   172:4436d287bfaa
user:        Petre Mierlutiu pmierlutiu@...
date:        Fri Oct 14 14:22:40 2016 +0300
summary:     Added tag 1.3.0-1 for changeset 9b5b4375432e

changeset:   171:9b5b4375432e
tag:         1.3.0-1
tag:         1.3.0.1
tag:         1.3.0.2
parent:      68:afd9055fb46f
parent:      170:ab327fdcdc33
user:        Petre Mierlutiu pmierlutiu@cmedtechnology.com
date:        Thu Oct 13 12:53:17 2016 +0000
summary:     Merged in 1.3.0-dev (pull request #24)

### DIFF
--- a/setuptools_scm/hg.py
+++ b/setuptools_scm/hg.py
@@ -8,13 +8,18 @@ FILES_COMMAND = 'hg locate -I .'
 def _hg_tagdist_normalize_tagcommit(root, tag, dist, node):
     dirty = node.endswith('+')
     node = node.strip('+')
-    st = do('hg st --no-status --change %s' % str(node), root)
-
-    trace('normalize', locals())
-    if int(dist) == 1 and st == '.hgtags' and not dirty:
-        return meta(tag)
+    revset = ("(branch(.) and tag('{0}')::. and file('re:^(?!\.hgtags).*$')"
+              " - tag('{0}'))").format(str(tag))
+    if tag != '0.0':
+        commits = do(['hg', 'log', '-r', revset, '-T', '{node|short}'],
+                     root)
     else:
+        commits = True
+    trace('normalize', locals())
+    if commits or dirty:
         return meta(tag, distance=dist, node=node, dirty=dirty)
+    else:
+        return meta(tag)
 
 
 def parse(root):
@@ -36,7 +41,9 @@ def parse(root):
     out = do(cmd, root)
     try:
         # in merge state we assume parent 1 is fine
-        tag, dist = out.splitlines()[0].split()
+        tags, dist = out.splitlines()[0].split()
+        # pick latest tag from tag list
+        tag = tags.split(':')[-1]
         if tag == 'null':
             tag = '0.0'
             dist = int(dist) + 1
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt:
seems this breaks on some mercurial versions

RonnyPfannschmidt:
personally i'd rather remove the support for ignoring tag commits, it is fragile to begin with
a better alternative would perhaps be to wrap up tags in hg a bit more and make soemthing that works most of the time but can be opted out

the main reason why i added support for tag commit ignorance is, its more easy in manual workflows

correct automation instead could just check out the **actual** tag and needs no guessing as the current and the proposed method do

my personal preference would be to provide more detailed/robust guessing as opt-in
and to phase out guessing as a default

petrem:
I will look at the failures. I don't think I fully follow your comment. Perhaps you could point me to some reading? 
It is the first time i'm attempting to use setuptools_scm so i probably don't get its intentions.
Thanks.

RonnyPfannschmidt:
There is no reading yet

setuptools_scm/hg.py
```diff
@@ -8,13 +8,18 @@
 def _hg_tagdist_normalize_tagcommit(root, tag, dist, node):
     dirty = node.endswith('+')
     node = node.strip('+')
-    st = do('hg st --no-status --change %s' % str(node), root)
-
-    trace('normalize', locals())
-    if int(dist) == 1 and st == '.hgtags' and not dirty:
-        return meta(tag)
+    revset = ("(branch(.) and tag('{0}')::. and file('re:^(?!\.hgtags).*$')"
+              " - tag('{0}'))").format(str(tag))
```
RonnyPfannschmidt: i wonder if instead of using str here, the string itself should use `{tag!r}` and the format call should use `tag=tag`

petrem: I'm not sure repr is guaranteed to use simple quotes. `repr("abc")` is `'abc'`, but if it would ever be `"abc"` that would break the revset. Whereas with `str` (or `{tag!s}) you get the plain string without quotes.

petrem: Turns out it doesn't matter, strings can be represented as single or double quoted in revsets, so never mind.

A slight concern could be that the tag might not be a string and might have an unsuitable repr, but unlikely..

RonnyPfannschmidt:
thanks for your work and the swift followups on the comments

Issue #112, opened 2016-10-18
PyPI trove classifieds do not match what is in setup.py


cclauss:
https://github.com/pypa/setuptools_scm/blob/master/setup.py#L110-L113

Tools that use PyPI data can not automatically determine if setuptools_scm is Python3 compatible.

RonnyPfannschmidt:
So what should I put there?

RonnyPfannschmidt:
Feel free to just open a PR with better values :)

cclauss:
In fact, the problem is not what is in this repo.  Instead, the PyPI record at https://pypi.python.org/pypi/setuptools_scm is missing the trove classifiers that are in setup.py.  Compare the text at the very bottom of that page with the previous version https://pypi.python.org/pypi/setuptools_scm/1.13.0 where the trove classifiers were correct.

RonnyPfannschmidt:
it seems that the release process needs an updated script

RonnyPfannschmidt:
will be testing a new upload procedure for 1.15.0

RonnyPfannschmidt:
fixed in https://pypi.python.org/pypi/setuptools_scm/1.15.0 - but may come up again

cclauss:
Thanks.  The setuptools_scm project is in the top 200 most downloaded PyPI packages and its status is back to green on https://python3wos.mybluemix.net/

PR #113, opened 2016-10-18
Update setup.py

cclauss:
Python 3.4 --> Python 3.5 to match https://github.com/pypa/setuptools_scm/blob/master/.travis.yml#L4-L8

Related to https://github.com/pypa/setuptools_scm/issues/112 but NOT the fix for https://github.com/pypa/setuptools_scm/issues/112 .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -110,7 +110,7 @@ arguments = dict(
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 2.7',
-        'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
         'Topic :: Software Development :: Libraries',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -110,7 +110,7 @@ def make_release_tree(self, base_dir, files):
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 2.7',
-        'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
```
RonnyPfannschmidt, line 113: why replace instead of add?

cclauss: Both are acceptable.  The .travis.yml file talks about 2.6, 2.7, 3.3, 3.4, and 3.5 so you wanted to you could mention all.

PR #125, opened 2016-10-24
suggest write_to config for pkg distribution

majuscule:
I just spent a stupid amount of time trying to figure out why setuptools_scm wasn't magically handling the version handling after building distribution formats via setuptools. Hopefully this README update will clear that up for anyone else as obtuse as me.

Also worth mentioning that I couldn't seem to figure out how to get .rst refs to play nicely with GitHub anchors.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,7 +31,14 @@ To use setuptools_scm just modify your project's setup.py file like this:
            ...,
        )
 
-3. If you will be distributing your package via {s,b}dist, consider using :code:`write_to` (see *Configuration Parameters* below) for visibility after the build process strips vcs data.
+3. Access the version number in your package via :code:`pkg_resources`
+
+   E.g. (`PEP-0396 <https://www.python.org/dev/peps/pep-0396>`_):
+
+   .. code:: python
+
+      import pkg_resources
+      __version__ = pkg_resources.get_distribution(__name__).version
 
 
 Programmatic usage
--- a/README.rst
+++ b/README.rst
@@ -31,6 +31,8 @@ To use setuptools_scm just modify your project's setup.py file like this:
            ...,
        )
 
+3. If you will be distributing your package via {s,b}dist, consider using :code:`write_to` (see *Configuration Parameters* below) for visibility after the build process strips vcs data.
+
 
 Programmatic usage
 ------------------
--- a/README.rst
+++ b/README.rst
@@ -37,8 +37,12 @@ To use setuptools_scm just modify your project's setup.py file like this:
 
    .. code:: python
 
-      import pkg_resources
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
@majuscule i'd like to understand what exactly your problem was,

my personal preference is to never ever have a `version.py` style file, because pkg_resources can provide the metadata

anthrotype:
I'd curious to see if there's any overhead in importing pkg_resources vs using version.py method.

RonnyPfannschmidt:
there is some overhead when using `pkg_resources` (finding a distribution has a certain cost)

but at least anything thats using entrypoints can neglect that cost

majuscule:
It seems my real problem was that I didn't know `pkg_resources` was a thing either. I'm packaging a python project for the first time and despite reading a lot about the process, I've had trouble figuring out some details like this. I was actually searching for the functionality pkg_resources provides earlier today, but with the wrong keywords.

How about instead suggesting

```
import pkg_resources
__version__ = pkg_resources.get_distribution(__name__).version
```

as that seems to be the cleanest solution.

README.rst
```diff
@@ -31,6 +31,15 @@ To use setuptools_scm just modify your project's setup.py file like this:
...
        )
 
+3. Access the version number in your package via :code:`pkg_resources`
+
+   E.g. (`PEP-0396 <https://www.python.org/dev/peps/pep-0396>`_):
+
+   .. code:: python
+
+      import pkg_resources
+      __version__ = pkg_resources.get_distribution(__name__).version
```
RonnyPfannschmidt: good change :+1: 

for ensuring function in all cases, i suggest putting that into a `try/except/`
and issuing a warning in case of being unable to determine the version

majuscule: I'm not sure what exception would be best to catch here, `pkg_resources.DistributionNotFound`?

RonnyPfannschmidt: yes

RonnyPfannschmidt:
Good work, thanks for following through

anntzer:
For an additional data point: I prefer version.py to pkg_resources because the latter quickly goes out of sync when using editable installs, whereas, when using version.py, I can first check whether I am in a git repo and decide accordingly, with a snippet like
```
import setuptools_scm
try:
    __version__ = setuptools_scm.get_version(  # xref setup.py
        root="..", relative_to=__file__)
except LookupError:
    try:
        from ._version import version as __version__
    except ImportError:
        pass
```

This pattern also suggest to me that supporting setting version_scheme and local_scheme from setup.cfg would be nice too (so that I don't have to repeat it both in setup.py and in `__init__.py`).

Issue #126, opened 2016-10-26
AttributeError when last tag is post release tag


jaraco:
Consider this minimal example:

```
$ mkdir issue; cd issue
$ cat > setup.py
import setuptools
setuptools.setup(
  use_scm_version=True,
  setup_requires=['setuptools_scm==1.15.0'],
)
$ hg init .
$ hg add setup.py
$ hg ci -m "Simple project"
$ hg tag 23.24post22+deadbeef
$ python setup.py egg_info
running egg_info
creating UNKNOWN.egg-info
writing UNKNOWN.egg-info/PKG-INFO
writing dependency_links to UNKNOWN.egg-info/dependency_links.txt
writing top-level names to UNKNOWN.egg-info/top_level.txt
writing manifest file 'UNKNOWN.egg-info/SOURCES.txt'
writing manifest file 'UNKNOWN.egg-info/SOURCES.txt'
$ cat UNKNOWN.egg-info/PKG-INFO| grep Version
Metadata-Version: 1.0
Version: 23.24.post22+deadbeef
$ touch foo
$ hg add foo
$ hg ci -m "Adding foo"
$ python setup.py egg_info                   
Traceback (most recent call last):
  File "setup.py", line 4, in <module>
    setup_requires=['setuptools_scm==1.15.0'],
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/distutils/core.py", line 108, in setup
    _setup_distribution = dist = klass(attrs)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/setuptools/dist.py", line 318, in __init__
    _Distribution.__init__(self, attrs)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/distutils/dist.py", line 281, in __init__
    self.finalize_options()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/setuptools/dist.py", line 376, in finalize_options
    ep.load()(self, ep.name, value)
  File "/Users/jaraco/Dropbox/code/public/setuptools_scm/issue/.eggs/setuptools_scm-1.15.0-py3.6.egg/setuptools_scm/integration.py", line 19, in version_keyword
  File "/Users/jaraco/Dropbox/code/public/setuptools_scm/issue/.eggs/setuptools_scm-1.15.0-py3.6.egg/setuptools_scm/__init__.py", line 123, in get_version
[...]

PR #127, opened 2016-10-26
Discard local parts when guessing next version.

jaraco:
This series of commits uses test-driven development to demonstrate the failure and a suitable fix for the described issue.

Side note: in d17972e, I extracted functions for `_bump*` initially thinking that the solution would add a third technique for using a parsed version for detecting the parts, but after further consideration (and other failed tests), I settled on this approach. It would be possible to rebase the subsequent tests on 2743575 and bypass d17972e, but I personally prefer the approach as suggested.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+v1.15.1
+=======
+
+* fix issue #126: the local part of any tags is discarded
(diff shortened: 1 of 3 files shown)


jaraco:
Here I mention #126 to link the PR with the original ticket.

PR #133, opened 2016-11-24
closes #132: adds instructions for passing dict to use_scm_version

### LINKED ISSUE #132: document extended setup.py usage more comptehensive
that #131 was asked indicates my Readme is not good enough and needs to be enhanced

i came up with the hypothesis that for a new user passing a mapping is not at all intuitive and needs examples that ease intuitively understanding that way of using setuptools_scm

reece:
Here're instructions that worked for me. Thanks Ronny!

See the README at the fork https://github.com/reece/setuptools_scm for the rendering.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,7 +18,6 @@ To use setuptools_scm just modify your project's setup.py file like this:
 1. Add :code:`'setuptools_scm'` to the :code:`setup_requires` parameter
 2. Add the :code:`use_scm_version` parameter and set it to ``True``
 
-
    E.g.:
 
    .. code:: python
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt (approved):

RonnyPfannschmidt:
Well done will merge in the pc

jdpatt:
For anyone trying to do this with a `pyproject.toml` file, this worked for me being nested two folders down.

```toml
[tool.setuptools_scm]
root = "../../"
```

Updated with feedback from @RonnyPfannschmidt which also works.

RonnyPfannschmidt:
@jdpatt in future it will be a error to pass relative_to as part of pyproject.toml

Issue #138, opened 2016-12-08
pip install -r requirements.txt fails to "setuptools-scm was unable to detect version for '/tmp'"


tuukkamustonen:
Version: `1.15.0`

I have requirements declared in `setup.py`, so with `requirements.txt` file of:

    .

`pip install -r requirement.txt`:

```
Processing /home/musttu/Code/lib/wires/aws
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-pgsLbN-build/setup.py", line 31, in <module>
        'pastry-common==3.2.1.dev30',
      File "/usr/lib/python2.7/distutils/core.py", line 111, in setup
        _setup_distribution = dist = klass(attrs)
      File "/home/musttu/Code/virtualenvs/wires/local/lib/python2.7/site-packages/setuptools/dist.py", line 318, in __init__
        _Distribution.__init__(self, attrs)
      File "/usr/lib/python2.7/distutils/dist.py", line 287, in __init__
        self.finalize_options()
      File "/home/musttu/Code/virtualenvs/wires/local/lib/python2.7/site-packages/setuptools/dist.py", line 376, in finalize_options
        ep.load()(self, ep.name, value)
      File "/home/musttu/Code/virtualenvs/wires/local/lib/python2.7/site-packages/setuptools_scm/integration.py", line 19, in version_keyword
        dist.metadata.version = get_version(**value)
      File "/home/musttu/Code/virtualenvs/wires/local/lib/python2.7/site-packages/setuptools_scm/__init__.py", line 117, in get_version
        parsed_version = _do_parse(root, parse)
      File "/home/musttu/Code/virtualenvs/wires/local/lib/python2.7/site-packages/setuptools_scm/__init__.py", line 95, in _do_parse
        "use git+https://github.com/user/proj.git#egg=proj" % root)
    LookupError: setuptools-scm was unable to detect version for '/tmp'.
```

I don't believe it's possible to fix this within `setuptools-scm`, however. If pip uses `/tmp` for temporarily building/handling the package, there's simply no version information available and nothing to be done. Correct? Would you have any guesses if it's possible to make it work on pip side?

There is a related ticket in https://github.com/pypa/setuptools_scm/issues/46 but I think that was closed because user tried to install an sdist there.

RonnyPfannschmidt:
[...]

Issue #42, opened 2015-08-17
Failing tests in 1.7.0


vext01:
On OpenBSD, I notice that the tests all fail unless the package has been installed (for some reason test suite looks for setuptools-scm instead of setuptools_scm):

```
============================= test session starts ==============================
platform openbsd5 -- Python 2.7.10 -- py-1.4.23 -- pytest-2.6.1
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/usr/local/lib/python2.7/site-packages/_pytest/main.py", line 82, in wrap_session
INTERNALERROR>     config.hook.pytest_sessionstart(session=session)
INTERNALERROR>   File "/usr/local/lib/python2.7/site-packages/_pytest/core.py", line 413, in __call__
INTERNALERROR>     return self._docall(methods, kwargs)
INTERNALERROR>   File "/usr/local/lib/python2.7/site-packages/_pytest/core.py", line 424, in _docall
INTERNALERROR>     res = mc.execute()
INTERNALERROR>   File "/usr/local/lib/python2.7/site-packages/_pytest/core.py", line 315, in execute
INTERNALERROR>     res = method(**kwargs)
INTERNALERROR>   File "/usr/local/lib/python2.7/site-packages/_pytest/terminal.py", line 285, in pytest_sessionstart
INTERNALERROR>     config=self.config, startdir=self.startdir)
INTERNALERROR>   File "/usr/local/lib/python2.7/site-packages/_pytest/core.py", line 413, in __call__
INTERNALERROR>     return self._docall(methods, kwargs)
INTERNALERROR>   File "/usr/local/lib/python2.7/site-packages/_pytest/core.py", line 424, in _docall
INTERNALERROR>     res = mc.execute()
INTERNALERROR>   File "/usr/local/lib/python2.7/site-packages/_pytest/core.py", line 315, in execute
INTERNALERROR>     res = method(**kwargs)
INTERNALERROR>   File "/usr/ports/pobj/py-setuptools_scm-1.7.0/setuptools_scm-1.7.0/testing/conftest.py", line 10, in pytest_report_header
INTERNALERROR>     version = pkg_resources.get_distribution(pkg).version
INTERNALERROR>   File "/usr/local/lib/python2.7/site-packages/pkg_resources.py", line 344, in get_distribution
INTERNALERROR>     if isinstance(dist,Requirement): dist = get_provider(dist)
[...]

Issue #144, opened 2017-01-22
Support SemVer pre-release versions (like 2.0.0-beta)


ralokt:
Currently, any attempt to use such a version will cause infinite sadness.

This is because of the current implementation of `tag_to_version`:

```
def tag_to_version(tag):
    trace('tag', tag)
    # lstrip the v because of py2/py3 differences in setuptools
    # also required for old versions of setuptools
    version = tag.rsplit('-', 1)[-1].lstrip('v')
    if parse_version is None:
        return version
    version = parse_version(version)
    trace('version', repr(version))
    if isinstance(version, SetuptoolsVersion):
        return version
```

ie, `"2.0.0-beta"` is truncated to `"beta"`.

I understand that this is there in order to support tags like `internal_spam-v1.2.3`, and that what I'm asking for is directly at odds with that.

However, SemVer does mandate using Hyphens for pre-release versions:
> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version.

(from [http://semver.org/#spec-item-9](http://semver.org/#spec-item-9) )

Perhaps this could be adressed in the course of the planned switch to SemVer?

RonnyPfannschmidt:
@ralokt pep44 differs a bit from semver with those regards, and currently i'm inclined to prefer mapping the rules of smever to pep440 as opposed to following a foreign version format for python tools

python versions already capture alpha/beta and a series of other cases

that being said, i wouldn't mind a proposed mechanism that does this correctly, but i do fear "correct" wil lbe far more tricky than anticipated right now

ralokt:
I gave it a try! Let me know what you think, I'm open to suggestions (and implementing them). Possibly we'd even want to be a bit less strict? I'm not considering pep440 in my solution at the moment (but that should currently be covered anyway).

RonnyPfannschmidt:
im closing this one, feel free to reopen when you find time to follow up

PR #145, opened 2017-01-22
add support for SemVer prerelease syntax in a mostly backwards-compat…

ralokt:
…ible way

This pull request contains my proposed solution for #144  It is backwards compatible except in some extreme edge cases - SemVer versions are now extracted using a very specific regex, and if no SemVer is found, the old implementation is used.

The only way this can fail is if the end of the internal part of a tag is the start of a valid SemVer, which sounds like a Very Bad Idea anyway.

### DIFF
--- a/setuptools_scm/version.py
+++ b/setuptools_scm/version.py
@@ -7,6 +7,48 @@ from pkg_resources import iter_entry_points
 
 from distutils import log
 
+NUMERIC_NO_LZ = r"(0|[1-9][0-9]*)"
+PRE_COMPONENT = r"([a-zA-Z1-9\-][a-zA-Z0-9\-]*)"
+META_COMPONENT = r"([a-zA-Z0-9\-]+)"
+
+VERSION_RE_SRC = r"""
+    ^
+    v?                         # optional v prefix (as in v1.2.3)
+    (?P<version>               # begin named group matching the actual version
+        {NUMERIC_NO_LZ}        # major
+        \.
+        {NUMERIC_NO_LZ}        # minor
+        (
+            \.
+            {NUMERIC_NO_LZ}    # optional patch
+        )?
+        (                      # begin optional prerelease part
+            \-
+            {PRE_COMPONENT}    # first component (fE "beta")
+            (
+                \.
+                {PRE_COMPONENT}
+            )*                 # and any number more
+        )?                     # end prerelease
+        (                      # begin optional build metadata part
+            \+
+            {META_COMPONENT}   # first component
+            (
+                \.
+                {META_COMPONENT}
+            )*                 # and any number more
+        )?                     # end metadata
+    )                          # end version
+    $
+""".format(
+    NUMERIC_NO_LZ=NUMERIC_NO_LZ,
+    PRE_COMPONENT=PRE_COMPONENT,
+    META_COMPONENT=META_COMPONENT,
+)
+
+VERSION_RE = re.compile(VERSION_RE_SRC, re.VERBOSE)
+
+
 try:
     from pkg_resources import parse_version, SetuptoolsVersion
 except ImportError as e:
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt:
looks good at first glance, will allocate a time slot for deeper review

thanks for putting it together so quickly

RonnyPfannschmidt:
btw, the appevoyer failures seem to be caused by it being bad again, so this one is definitively tested ok

ralokt:
Yay, awesome, and no problem!

RonnyPfannschmidt:
@ralokt this cant possibly work

add a check for `assert tag_to_version(expected) is not None` and you will see

ralokt:
Ah yes, I see the problem - `parse_version` doesn't like all valid SemVer versions, apparently (which makes sense as it implements PEP 440).

Should we guarantee that `tag_to_version_string` always returns something that `parse_version` can read, though? Currently, if we redefined it to implement the old behaviour, it wouldn't (as evidenced by exactly the kind of inputs I'm trying to get it to recognize).

As far as I see it, the question here is whether you want to support PEP 440, SemVer, or both.

PEP 440 does allow the pre-release separator to be a hyphen, though: [https://www.python.org/dev/peps/pep-0440/#pre-release-separators](https://www.python.org/dev/peps/pep-0440/#pre-release-separators)

So, even if supporting SemVer is not a concern, version numbers like `1.1-a1` are currently still not handled correctly despite being PEP 440-compliant.

If you don't want to support SemVer, I'd like to propose using `parse_version` instead of my SemVer regular expression to find the longest hyphen-delimited valid version at the end of the tag (if it is available, that is).

If you do, then `parse_version` is insufficient as it doesn't aim to support SemVer, and I'd propose using something else to parse versions.

Either way, I'd be glad to help and/or discuss this further!

RonnyPfannschmidt:
setuptools correctly normalizes 

```
>>> v = pkg_resources.parse_version("1.1-a1")
>>> v
<Version('1.1a1')>
```
i'm inclined to switch around behavior and strip all non-digit characters at the start instead of introducing detailed handling that's not supportable

it would need a very strong argument to drop pep 440, after all this is for python and python has standardized on pep440

RonnyPfannschmidt:
aware of an implementation of semver parsing, that can translate to pep440 ? an preparsing mechanism might be a reasonable tool

ralokt:
Yes, but setuptools_scm doesn't right now, as it will feed the string `"a1"` to `parse_version` after splitting off the rest.

`parse_version` is very lenient anyway, maybe we should redefine what we want to do as simply "feed the right thing to `parse_version`"?

Right now, we're kinda munging "extract the version from the tag" and "parse the version" together. While this might make sense implementation-wise, it's different in concept - the first "only" has to correctly extract what was meant to be a valid version by the user, especially if validation follows.

Stripping all non-numeric characters is a nice approach to that, but would break on inputs like `"fix issue #42-v0.7.4"` (which is ugly but probably not entirely unrealistic, and currently supported).

I'm not aware of a direct translation tool, but packages like [semantic_version](https://pypi.python.org/pypi/semantic_version/) or [semver](https://pypi.python.org/pypi/semver) implement parsing of semantic versions that could be used to normalize to PEP 440.

RonnyPfannschmidt:
i think instead of changing how tags are interpreted, it may be better to create the capability to strip project "noise" thats part of the tag away and leave only the version to setuptools_scm as a part of the configuration

ralokt:
And have the old behavior (split off at last hyphen) as the default? I like that a lot, would you like me to implement it?

RonnyPfannschmidt:
it would be great to see a proposal of that one

also it'd keep me free to fix #138 (the breakage was introduced by me 2 years back and needs an api change to fix)

RonnyPfannschmidt:
@ralokt can i close this one for now? im currently triaging issues

ralokt:
Yes, of course! Sorry for being unresponsive, too many things happening on my end. I'll just contact you again once I have the time to put some more work in here!

RonnyPfannschmidt:
@ralokt thanks for the update, i hope things get better for you soon

PR #146, opened 2017-01-26
[wip] Setup/fix version

RonnyPfannschmidt:


testing/runtests_travis.py
```diff
@@ -0,0 +1,14 @@
...
+
+import os
+
+if os.environ['TOXENV']:
+    import tox
+    tox.cmdline()
+elif os.environ['SELFINSTALL']:
+    call('python setup.py sdist', shell=True)
+    call('easy_install dist/*', shell=True)
+    import pkg_ressources
```
jaraco: This looks misspelled.

.travis.yml
```diff
@@ -9,6 +9,7 @@ python:
 - '3.6-dev'
 env:
 - TOXENV=py-test
+- SELFINSTALL=1
```
jaraco: Maybe consider instead just having two TOXENVs, one which is py-test and the other self-install... and perform the self-install checks in that toxenv.

testing/runtests_travis.py
RonnyPfannschmidt: thansk for reviwing, i got sidetracked, i hope to finish until the weekend

.travis.yml
RonnyPfannschmidt: the correctly created sdist of tox is a problem, thus i choose to make different travis level calls

PR #151, opened 2017-02-22
Ensure that LICENSE file is included in wheels

pombredanne:
Without this it is never included.
 
Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>


RonnyPfannschmidt:
thanks :)

PR #153, opened 2017-02-22
Add note about git describe

### LINKED ISSUE #152: revision hash adding extra "g"
Hello! I've just discovered this project, and I'm seeing something that doesn't make sense (I'm most likely just doing something wrong, but don't know where else to ask the question). 

1. I have a git repo in a package called `foo` that I tagged it with `git tag -a 1.0.0`
2. I have added setuptools_scm and enabled it in `setup.py`. 
3. When I run `python setup.py sdist`, I get a tarball of version: `foo-1.0.1.dev39+ng14ecfe5`

There are 2 confusing bits:

**autoincrement**
the `1.0.1` is auto-incremented. I think this is expected behavior, but wanted to call it out anyway. I would have expected the distance and revision hash to be appended to the _last known_ version, rather than assuming that I'm working on the next patch version. (e.g. what if I'm really working on the next major or minor release?)

**`+ng<revision>`**
The docs say that the revision hash will be appended with `+n<revision>`, but I'm seeing that extra "g" character. 

Thanks for this project, I am very excited to stop trying to match tags to published version numbers!

mehulkar:
closes #152

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -123,6 +123,9 @@ and uses roughly the following logic to render the version:
 The next version is calculated by adding ``1`` to the last numeric component
 of the tag.
 
+For git projects, the version is version is relies on `git describe <https://git-scm.com/docs/git-describe>`_,
+so you will see an additional ``g`` prepended to the ``{revision hash}``.
+
 Semantic Versioning (SemVer)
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
well done, thanks 👍

Issue #134, opened 2016-11-25
Regression: get_version works for 1.10.1 and 1.11.1, but not 1.13.1


reece:
# Summary

setuptools_scm 1.13.1 fails for several of my projects. 1.10.1 and 1.11.1 worked fine. The ultimate error is (full traceback below):

```
  File "/home/reece/projects/biocommons/bioutils/venv/lib/python3.5/site-packages/setuptools_scm/version.py", line 86, in meta
    assert tag is not None, 'cant parse version %s' % tag
AssertionError: cant parse version None

```

Python 3.5.2, Ubuntu 16.04, various versions of setuptools_scm (see below). 

This error appears to occur only for head revisions on several repos. Affected repos use hg-git, although I have no evidence that that's related.

# Reproduction

```
snafu$ hg clone git@github.com:biocommons/bioutils.git
destination directory: bioutils
importing git objects into hg
updating to branch default
41 files updated, 0 files merged, 0 files removed, 0 files unresolved

snafu$ cd bioutils/

snafu$ pyvenv venv

snafu$ source venv/bin/activate

(venv) snafu$ pip install --upgrade pip setuptools
Collecting pip
  Using cached pip-9.0.1-py2.py3-none-any.whl
Collecting setuptools
  Using cached setuptools-28.8.0-py2.py3-none-any.whl
Installing collected packages: pip, setuptools
  Found existing installation: pip 8.1.1
    Uninstalling pip-8.1.1:
      Successfully uninstalled pip-8.1.1
  Found existing installation: setuptools 20.7.0
    Uninstalling setuptools-20.7.0:
      Successfully uninstalled setuptools-20.7.0
Successfully installed pip-9.0.1 setuptools-28.8.0

# recent commits, with any tags
(venv) snafu$ hg log -l 7 -G -T '{rev} {node|short} {tags}'
@  59 4aac87427260 default/default/master default/master tip
|
o  58 f9ca4a4c9c1c 0.2.0a2
|
o  57 f7d97e9493ae
|
o  56 57709bd798a4
|
o  55 019f38efb5c2
|
o  54 12c36903cb55 0.2.0a1
|
o  53 6bf205dc9ec0

# generate a table of get_version results for the above commits and setuptools_scm 1.10.1, 1.11.1, 1.13.1
[...]

PR #154, opened 2017-02-24
WIP: Optionally omit the "g" for "git describe"

sschuberth:
Following the discussion started [here](https://github.com/pypa/setuptools_scm/issues/152#issuecomment-282271402) I came up with some basic changes that could lead into the direction of optionally omittin the "g" for "git describe".

The first commit is just a more or less unrelated optimization to do fewer Git calls.

The second commit removes the "g" prefix in the lower level functions by default. This will currently break things, so my idea is to optionally add the "g" back in one of the higher level functions. I'd rather do it that way than the other way around, because in the long run the "g" prefix should be dropped altogether.

Even if this is WIP and not ready yet, I'm posting this here to get some early feedback.

### DIFF
--- a/setuptools_scm/git.py
+++ b/setuptools_scm/git.py
@@ -108,6 +108,11 @@ def parse(root, describe_command=DEFAULT_DESCRIBE, pre_parse=warn_on_shallow):
       dirty = False
 
     tag, number, node = out.rsplit('-', 2)
+
+    # Omit the leading 'g' which is not part of the revision to be consistent with
+    # other code paths.
+    node = [1:]
+
     number = int(number)
     if number:
         return meta(tag, distance=number, node=node, dirty=dirty)
--- a/setuptools_scm/git.py
+++ b/setuptools_scm/git.py
@@ -4,7 +4,7 @@ from os.path import abspath, normcase, realpath, isfile, join
 import warnings
 
 FILES_COMMAND = 'git ls-files'
-DEFAULT_DESCRIBE = 'git describe --tags --long --match *.*'
+DEFAULT_DESCRIBE = 'git describe --tags --long --dirty --match *.*'
 
 
 def _normalized(path):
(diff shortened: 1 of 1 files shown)


setuptools_scm/git.py
```diff
@@ -4,7 +4,7 @@
 import warnings
 
 FILES_COMMAND = 'git ls-files'
-DEFAULT_DESCRIBE = 'git describe --tags --long --match *.*'
+DEFAULT_DESCRIBE = 'git describe --tags --long --dirty --match *.*'
```
RonnyPfannschmidt, line 7: does this work consistently, i very vaugely recall trouble with some older git versions that still may be in some distros

sschuberth: At least I've never had any issues with it.

RonnyPfannschmidt:
note that i proposed the other way around specifically because i perceived it as more clean

because now it removes the g at a lower level scm specific function and would have to conditionally add it back while no longer knowing the scm without looking again

from my pov thats layering of band-aids and not desirable as it introduces more points of fragility

that being said, i did think a bit and i am perfectly fine with releasing a 2.0 for this change as is without bringing back g

on that note i am also open for a short discussion on ensuring each scm gets its own node prefix and how that should look

setuptools_scm/git.py
RonnyPfannschmidt: thats ok, but "works for me" is not a metric end-users appreciate ^^

for examples see #134

jaraco: I seem to recall the dirty flag was to distinguish a working directory with uncommitted changes from a clean working copy. I'm pretty sure you want the flag.

Agreed that this change doesn't belong in this request. If you want to make this change, propose it separately and trace its origin to prove that it's no longer needed (or in which conditions that's true).

sschuberth: Can you elaborate what exactly you mean by "trace its origin"?

jaraco: I mean find when that change was made and why, to try to confirm its purpose. Perhaps using a blame tool or bisecting the changelog or searching the project ticket history.

jaraco: I mean find when that change was made and why, to try to confirm its purpose. Perhaps using a blame tool or bisecting the changelog or searching the project ticket history.

sschuberth: I wasn't understanding @RonnyPfannschmidt in the way that `--dirty` already used to be there, but did cause problems, and thus was removed. I thought that was just a general remark. But I'll investigate.

jaraco: I think I misread the diff. Yeah, Ronnys comment is probably more meaningful.

sschuberth: Turns out `--dirty` indeed already was there: https://github.com/pypa/setuptools_scm/commit/929a5d7c916b37751c12750fe5a2d1372b457b61#diff-7a9c0c3ec91c440b46ba9f3506777bf0L7. Also see my PR #156.

sschuberth:
> while no longer knowing the scm without looking again

Good point. I'll change it, but probably in a separate PR.

> i am also open for a short discussion on ensuring each scm gets its own node prefix

How about simply taking the first letter of the respective command line tool name (`g` for `git`, `h` for `hg` i.e. Mercurial, etc.), or if that would be ambiguous, even the full name of the respective command line tool?

RonnyPfannschmidt:
using g/h sounds good, both are outside hex

sschuberth:
Good. So I'll eventually file a PR that adds the "h" prefix for hg, and removes the global "n" prefix. Closing this PR as it's deprecated now.

RonnyPfannschmidt:
:+1:

PR #156, opened 2017-02-24
Prefer to only use 'git describe' for all required info

sschuberth:
Previously, three git commands were used to gather all required info.
Now all info is gathered from a single 'git describe' call and only if
that fails fallbacks are used.

There already was the --dirty option before commit 929a5d7, but no
reason was given why it was removed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,9 @@ v1.15.1
 
 * fix issue #126: the local part of any tags is discarded
   when guessing new versions
+* minor performance optimization by doing fewer git calls
+  in the usual cases
+
 
 v1.15.0
 =======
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt:
the reason was #86 - as git describe is failing with no tags present or no commits present

sschuberth:
Ok, but in that case the fallback should kick in, and all is good, or?

RonnyPfannschmidt:
@sschuberth yes, i just finished crosschecking, and your adaption is more nice as its faster in all usually relevant cases

RonnyPfannschmidt:
@sschuberth please add a change-log entry, then i can merge, good work 👍

sschuberth:
For which version? 1.15.1 still?

RonnyPfannschmidt:
yup

PR #158, opened 2017-03-03
use sys.executable instead of 'python' for running regression tests

matejcik:
`test_regressions.py` is running a python subprocess, but it is using the "python" it finds on `PATH`, instead of the same executable that is running the test.

This could mean that it is testing the wrong thing (if the generic executable is using a different home path) or wrong version of Python (which also means that it creates stray `pyc`s for the other version).

### DIFF
--- a/testing/test_regressions.py
+++ b/testing/test_regressions.py
@@ -18,16 +18,16 @@ def test_pkginfo_noscmroot(tmpdir, monkeypatch):
         'from setuptools import setup;'
         'setup(use_scm_version={"root": ".."})')
 
-    _, stderr, ret = do_ex('python setup.py --version', p)
+    _, stderr, ret = do_ex(sys.executable + ' setup.py --version', p)
     assert 'setuptools-scm was unable to detect version for' in stderr
     assert ret == 1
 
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
good find, thanks for the fix

RonnyPfannschmidt:
oh fml

@matejcik it seems to have broken the windows tests due to path issues

RonnyPfannschmidt:
@matejcik can you take a look at fixing this one, it was my oversight to merge before the windows tests passed, and now i have an obligation to undo the fix because it doesnt work on windows

matejcik:
sorry, don't have a windows env to test it. are there buildbots i can look at or anything?

RonnyPfannschmidt:
@matejcik we do have a windows ci for pull requests, https://ci.appveyor.com/project/RonnyPfannschmidt/setuptools-scm-r14ad/build/1.0.126

it will build on pull requests, i am simply guilty of merging too early because your pr looked so good and i completely forgot about windows

matejcik:
oh, it's there hidden under the little X

PR #159, opened 2017-03-03
fix sys.executable failure on Windows

matejcik:
fix the windows failure of PR #158 by splitting the arguments manually instead of relying on shlex (which fails horribly on windows paths)


RonnyPfannschmidt:
well done, thanks :)

PR #160, opened 2017-03-04
Use per-SCM node prefix letters instead of the general "n"

sschuberth:
"Node" is an uncommon name for what SCMs usually call "revision", so the
"n" prefix is not very telling. Replace it with a prefix specific to the
SCM, so we actually also get to know with SCM is used.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -114,11 +114,11 @@ and uses roughly the following logic to render the version:
 :code:`no distance and clean`:
     :code:`{tag}`
 :code:`distance and clean`:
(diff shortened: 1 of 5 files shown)


sschuberth:
This is a follow-up to the discussion over [here](https://github.com/pypa/setuptools_scm/pull/154#issuecomment-282342374).

Since this is a breaking change I'm not sure `master` actually is the right target branch. @RonnyPfannschmidt please advise.

RonnyPfannschmidt:
@sschuberth im considering this for a 2.0 release

PR #174, opened 2017-06-19
fix issue with git workdir containing Windows short path names

anthrotype:
When doing pip install from a local git repository, pip moves the working tree to a temporary folder.

The `%TEMP%` environment variable may contain short path names, e.g. "C:\Users\COSIMO~1.LUP\AppData\Local\Temp".

However, when calling `GitWorkdir.from_potential_worktree`, the command `git rev-parse --show-toplevel` always return a path with long names.

Because of this, the root and 'real' working directory returned by `git rev-parse` compare different from each other, even if they are actually the same, and the method will return None.

This adds a new function `get_windows_long_path_name` in utils.py module to get the long path name using the Windows API (via ctypes).

### DIFF
--- a/setuptools_scm/git.py
+++ b/setuptools_scm/git.py
@@ -1,18 +1,10 @@
-from .utils import do_ex, trace, has_command
+from .utils import do_ex, trace, has_command, _normalized
 from .version import meta
-from os.path import abspath, normcase, realpath, isfile, join
+from os.path import isfile, join
 import warnings
-import platform
 
 FILES_COMMAND = 'git ls-files'
 DEFAULT_DESCRIBE = 'git describe --dirty --tags --long --match *.*'
-
-
-def _normalized(path):
-    if platform.system() == 'Windows':
-        from .utils import get_windows_long_path_name
-        path = get_windows_long_path_name(path)
-    return normcase(abspath(realpath(path)))
 
 
 class GitWorkdir(object):
--- a/setuptools_scm/git.py
+++ b/setuptools_scm/git.py
@@ -2,12 +2,16 @@ from .utils import do_ex, trace, has_command
 from .version import meta
 from os.path import abspath, normcase, realpath, isfile, join
 import warnings
+import platform
 
 FILES_COMMAND = 'git ls-files'
 DEFAULT_DESCRIBE = 'git describe --dirty --tags --long --match *.*'
 
 
 def _normalized(path):
+    if platform.system() == 'Windows':
+        from .utils import get_windows_long_path_name
+        path = get_windows_long_path_name(path)
     return normcase(abspath(realpath(path)))
 
 
--- a/setuptools_scm/utils.py
+++ b/setuptools_scm/utils.py
@@ -7,6 +7,7 @@ import sys
 import shlex
 import subprocess
 import os
+from os.path import abspath, normcase, realpath
 import io
 import platform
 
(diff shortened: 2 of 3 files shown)


setuptools_scm/utils.py
```diff
@@ -32,9 +34,7 @@ def _always_strings(env_dict):
     On Windows and Python 2, environment dictionaries must be strings
     and not unicode.
     """
-    is_windows = platform.system == 'Windows'
```
anthrotype, line 35: By the way, `platform.system()` is a function, not a static string attribute; thus the comparison `platform.system == 'Windows'` always returned False there.

https://docs.python.org/3/library/platform.html#platform.system

RonnyPfannschmidt: good find

setuptools_scm/git.py
```diff
@@ -2,12 +2,16 @@
 from .version import meta
 from os.path import abspath, normcase, realpath, isfile, join
 import warnings
+import platform
 
 FILES_COMMAND = 'git ls-files'
 DEFAULT_DESCRIBE = 'git describe --dirty --tags --long --match *.*'
 
 
 def _normalized(path):
```
RonnyPfannschmidt: i propose moving this declaration to utils

setuptools_scm/utils.py
```diff
@@ -107,3 +107,53 @@ def has_command(name):
...
+            pathlen = 1
+
+        for _ in range(2):
+            buf = create_unicode_buffer(pathlen)
+            retval = GetLongPathNameW(path, buf, pathlen)
+
+            if retval == 0:
+                # if the function fails for any reason (e.g. file does not
+                # exist), the return value is zero
+                raise WinError()
```
RonnyPfannschmidt, line 151: please add an error description to the exception

anthrotype: Actually you don't need to add a message for WinError, as it will relay the last error message that was raised by the operating system (if any). When the function return 0, it means some error occurred so it is enough to raise WinError like that. Try to pass an unexistent path to it, you'll see an error message displayed. Actually, I could add a test for that too.

setuptools_scm/git.py
anthrotype: Do you mean, moving the `_normalized` function to utils? Ok, I can do that.

setuptools_scm/utils.py
anthrotype: Search for the word "WinError" in https://docs.python.org/3/library/ctypes.html and look at the example there.

RonnyPfannschmidt: i'd appreciate a quick demo as i don't have a windows box

anthrotype: ok, I added a quick test for WinError in https://github.com/pypa/setuptools_scm/pull/174/commits/d49bdfea7e83ad8771464275e255614d8a4888cc

RonnyPfannschmidt (approved):
looks all good, thanks - i#ll wire up a release after im recovered from my infect

jaraco:
I can't help but wonder if there's a better solution for this issue. It sure seems messy to have to dig into the ctypes library and bypass the Python API for such a common operation.

Perhaps it would be possible to use `samefile` for the comparison:

```
>>> os.path.samefile('C:\\users\\longusername', 'c:\\users\longus~1')
True
```

woodyza:
Hah I was just about to spin up a VM to test if samefile would work. Nice, leveraging the path module abstraction to let ntpath handle the windows weirdness is a much better approach I think.

PR #177, opened 2017-07-12
Set example indentation to multiple of four

willirath:
In `README.rst`, there was an example with wrong indentation.


RonnyPfannschmidt:
good catch, thank you 👍

Issue #87, opened 2016-05-19
support the parsing of custom tag styles


wcooley:
We tag our releases with "**tag/x.y.z**" (don't ask), which causes an _AssertionError_ in `setuptools_scm/version.py/meta()` because `pkg_resources.parse_version` returns _None_ in `tag_to_version`.

All I need is to filter my tag to trim those 4 characters from the beginning: `lambda tag: tag[4:]`,
which makes re-implementing `parse` a bit overkill.

As an experiment, I've monkey-decorated `meta` in my `setup.py`, but it's too ugly for me to be happy schelpping about to all my projects:

``` python
import setuptools_scm.version
meta = setuptools_scm.version.meta
setuptools_scm.version.meta = lambda tag, *args, **kwargs: \
    meta(tag[4:], *args, **kwargs)
```

It would be much nicer to have a `cleanup_tag` callable for `use_scm_version`:

``` python
setup( ...
    use_scm_version={'cleanup_tag': lambda tag: tag[4:]}
    ...
)
```

I can imagine that there are plenty of other potential users in a similar situation, given the variety of release tag naming that I've seen in various repos.

RonnyPfannschmidt:
please show the exception

wcooley:
Are you suggesting that it should be able to handle a tag like "tag/1.0.0"?

My description was a _bit_ wrong; `pkg_resources.parse_version` returns an instance of `SetuptoolsLegacyVersion` instead of `SetuptoolsVersion`:

``` python
>>> import pkg_resources
>>> type(pkg_resources.parse_version('1.0.0'))
<class 'pkg_resources.SetuptoolsVersion'>
>>> type(pkg_resources.parse_version('tag/1.0.0'))
<class 'pkg_resources.SetuptoolsLegacyVersion'>
```

So the `if isinstance(version, SetuptoolsVersion):` in `tag_to_version` fails and falls off the end and `tag_to_version` returns `None`. Here's the backtrace:

```
$ python setup.py --version
Traceback (most recent call last):
  File "setup.py", line 4, in <module>
    use_scm_version=True,
  File "/usr/lib64/python3.4/distutils/core.py", line 108, in setup
    _setup_distribution = dist = klass(attrs)
  File "myvenv/lib/python3.4/site-packages/setuptools/dist.py", line 272, in __init__
    _Distribution.__init__(self,attrs)
  File "/usr/lib64/python3.4/distutils/dist.py", line 280, in __init__
    self.finalize_options()
[...]

PR #178, opened 2017-07-14
Feature: issue #87 clean_tag

wcooley:
An initial implementation of #87, adding a `clean_tag` parameter to `get_version`, which allows fixing up a non-conforming version-tagging scheme.

This isn't a _good_ implementation, but I am submitting it as a proof-of-concept.

Monkey-patching is a pretty horrible way to do this kind of thing but it demonstrates the difficulty of overriding this functionality.
(It also does not allow for overriding via an entrypoint.)
The most obvious alternative is to add parameters down all the way into  `meta` or `tag_to_version`, but that seems pretty ugly too.

I have some ideas for bigger architectural changes that would make it possible to implement more cleanly, but I'd like to know if anyone sees a better way to do this that I am unaware of and if the maintainers would be open to a bigger change before I spend more time on it.

Finally, I have kept the name "clean" instead of "parse" despite @RonnyPfannschmidt's objection for 2 reasons:
There's already enough ambiguity in `version.py` with `_parse_tag` and `meta` and `tag_to_version` and `parse_version` and `.git.parse` and `_do_parse`, etc.
Furthermore, in terms of encouraging conformity, "to clean" carries more negative connotations than "to parse", so the need to use this feature implies non-standard, less-than-desirable naming. (But as naming conventions get entwined in organizational processes, not having this is a major barrier to adoption for much pre-existing software.)

### DIFF
--- a/setuptools_scm/__init__.py
+++ b/setuptools_scm/__init__.py
@@ -104,6 +104,7 @@ def get_version(root='.',
                 write_to_template=None,
                 relative_to=None,
                 parse=None,
+                clean_tag=None,
                 ):
     """
     If supplied, relative_to should be a file from which root may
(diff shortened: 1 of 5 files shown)


RonnyPfannschmidt:
i#ll take a while before i can properly review, im currently caring for my newborn

setuptools_scm/__init__.py
```diff
@@ -116,8 +117,17 @@ def get_version(root='.',
     root = os.path.abspath(root)
     trace('root', repr(root))
 
+    if clean_tag:
```
RonnyPfannschmidt, line 120: please put the hack into a monkeypatching context-manager that is ensured to run and can be tested separately

RonnyPfannschmidt:
closing as it seems there no longer is interest, in case there is, im happy to reopen

denniszollo:
@wcooley 

As a datapoint, I just ran into the same problem, banged my head against the wall for a day, then determined that this package is just not architected to allow this kind of extension in a clean way.

RonnyPfannschmidt:
@denniszollo #276 is just about adding something like that in a clean way

PR #180, opened 2017-09-03
Fix unicode handling for version_scheme

### LINKED ISSUE #179: Options specified as `unicode` strings break
Under 2.7 but using `unicode_literals` in `setup.py`, I get:

```
Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/Users/amw/Projects/stethoscope/stethoscope/setup.py", line 221, in <module>
        setuptools.setup(**setup_params)
      File "/Users/amw/.pyenv/versions/2.7.13/lib/python2.7/distutils/core.py", line 111, in setup
      File "/Users/amw/.pyenv/versions/2.7.13/envs/scope-oss/lib/python2.7/site-packages/setuptools/dist.py", line 328, in __init__
        _Distribution.__init__(self, attrs)
      File "/Users/amw/.pyenv/versions/2.7.13/lib/python2.7/distutils/dist.py", line 287, in __init__
        self.finalize_options()
      File "/Users/amw/.pyenv/versions/2.7.13/envs/scope-oss/lib/python2.7/site-packages/setuptools/dist.py", line 461, in finalize_options
      File "/Users/amw/.pyenv/versions/2.7.13/envs/scope-oss/lib/python2.7/site-packages/setuptools_scm/integration.py", line 22, in version_keyword
      File "/Users/amw/.pyenv/versions/2.7.13/envs/scope-oss/lib/python2.7/site-packages/setuptools_scm/__init__.py", line 125, in get_version
        local_scheme=local_scheme)
      File "/Users/amw/.pyenv/versions/2.7.13/envs/scope-oss/lib/python2.7/site-packages/setuptools_scm/version.py", line 166, in format_version
        main_version = version_scheme(version)
    TypeError: 'unicode' object is not callable
```

I believe this check is responsible since it's only looking for `str` and not `unicode`, as well:
https://github.com/pypa/setuptools_scm/blob/75897ba511010efaabf628217d34b589d9299c8f/setuptools_scm/version.py#L25

jaraco:
Fixes #179. Includes a unit test (git only). I welcome tweaks to the test so it would not be specific to git.

### DIFF
--- a/testing/test_git.py
+++ b/testing/test_git.py
@@ -39,6 +39,12 @@ def test_version_from_git(wd):
     assert wd.version.startswith('0.2')
 
 
+@pytest.mark.issue(179)
+def test_unicode_version_scheme(wd):
+    scheme = b'guess-next-dev'.decode('ascii')
+    assert wd.get_version(version_scheme=scheme)
+
+
 @pytest.mark.issue(108)
 @pytest.mark.issue(109)
 def test_git_worktree(wd):
--- a/setuptools_scm/version.py
+++ b/setuptools_scm/version.py
@@ -22,12 +22,13 @@ def _warn_if_setuptools_outdated():
 
 def callable_or_entrypoint(group, callable_or_name):
     trace('ep', (group, callable_or_name))
-    if isinstance(callable_or_name, str):
-        for ep in iter_entry_points(group, callable_or_name):
-            trace("ep found:", ep.name)
-            return ep.load()
-    else:
+
+    if callable(callable_or_name):
         return callable_or_name
+
+    for ep in iter_entry_points(group, callable_or_name):
+        trace("ep found:", ep.name)
+        return ep.load()
 
 
 def tag_to_version(tag):


RonnyPfannschmidt:
the appveyor failure seems unrelated, thanks for picking it up, its a beautiful solution

testing/test_git.py
```diff
@@ -39,6 +39,12 @@ def test_version_from_git(wd):
     assert wd.version.startswith('0.2')
 
 
+@pytest.mark.issue(179)
+def test_unicode_version_scheme(wd):
```
RonnyPfannschmidt, line 43: how about turning that into a functional test passing it a few well known objects so only the `callable_or_entrypoint` function is under test?

jaraco: I thought it was important to capture the expectation that the user was supplying unicode to `version_scheme`, which in my mind is the functional requirement. I'd be fine with just a unit test on `callable_or_entrypoint` as well. Feel free to adapt it however you think is best.

PR #186, opened 2017-11-14
Respect git's export-ignore for building the sdist.

### LINKED ISSUE #185: support git's export-ignore
`git archive` reads the `export-ignore` gitattribute (https://git-scm.com/docs/gitattributes#_creating_an_archive) which allows marking some files as NOT packaged into the archive.  It would be nice if setuptools_scm did the same for the sdist.

Typical use cases would be to remove CI scripts.

Unfortunately AFAICT there is no way to ask git for the files it would archive short of actually running git archive and examining the resulting tarball.  OTOH it appears that such a step is much faster than the sdist build itself, so perhaps that remains an acceptable approach?

anntzer:
Implemented by replacing the call to 'git ls-files' to '$sys.executable
-msetuptools_scm.git' with the module parsing the output of 'git
archive' as a tar file.

Fixes #185.  Actually that was easier to do than expected...

I haven't edited the README because I don't think it even explicitly mentions right now that it'll sdist exactly all files tracked by the VCS (it's only implicit in the doc for `setuptools_scm.files_command`), so I don't have a good place to introduce the new piece of info...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@
 ## Directory-based project format:
 .idea/
 
+### Other editors
+.*.swp
+
 
 ### Python template
 # Byte-compiled / optimized
--- a/setuptools_scm/git.py
+++ b/setuptools_scm/git.py
@@ -120,8 +120,8 @@ def parse(root, describe_command=DEFAULT_DESCRIBE, pre_parse=warn_on_shallow):
         return meta(tag, node=node, dirty=dirty)
 
 
-def _main():
-    """List the files that 'git archive' wants.
+def _list_files_in_archive():
+    """List the files that 'git archive' generates.
     """
     # TarFile wants a seekable stream.
     stream = BytesIO(subprocess.check_output(['git', 'archive', 'HEAD']))
@@ -130,4 +130,4 @@ def _main():
 
 
 if __name__ == "__main__":
-    _main()
+    _list_files_in_archive()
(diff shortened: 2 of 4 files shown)


RonnyPfannschmidt:
thats a neat trick, lovely 👍

RonnyPfannschmidt:
i jsut crosschecked, the linting failures are unrelated and came in with a update of flake8

the python2.6 failure is unfortunate - im pondering wheter we should finally drop python2.6 here given that older releases support it fine

@jaraco i'd appreciate your input on the lower python version support

RonnyPfannschmidt (approved):

anntzer:
I guess it's not too hard to switch to Popen but I would appreciate if you could pick up that work (Py2.6 is veeeeery far away for me...).  Feel free to push to my branch.

RonnyPfannschmidt:
@jaraco lovely chain, thanks for just stepping in and implementing it 👍

RonnyPfannschmidt:
oh drats, i just noted the errors on windows, its not clear how to fix those over there
i#ll try to take a closer look in the afternoon

jaraco:
> i'd appreciate your input on the lower python version support

I'm happy to drop support for Python 2.6, though I still would prefer this implementation that happens to run on Python 2.6.

RonnyPfannschmidt:
tests still fail on windows :/

anntzer:
Fixed now.

RonnyPfannschmidt:
well done, thanks

PR #201, opened 2017-12-21
Added `get_local_node_and_timestamp()` for timestamp versions

### LINKED ISSUE #200: Allow custom timestamp for `get_local_node_and_date()`
## Expected Behavior
- Ability to set custom timestamps (e.g., `%Y%m%d%H%M%S` instead of `%Y%m%d`)
- Multiple dirty builds are sometimes generated in quick succession

## Actual Behavior
- Hardcoded `%Y%m%d` timestamp: https://github.com/pypa/setuptools_scm/blob/46df188a54e3eb2b0dfb93dbae59fb23249fbcce/setuptools_scm/version.py#L140-L144

## Specs
- `Python 3.6.1`
- `setuptools-scm 1.15.6`
```

### PR DESCRIPTION
**Pull request by engnadeau**
## Affected Issues
- Fixes #200 

## Proposed Changes
- Added function `get_local_node_and_timestamp(version, fmt='%Y%m%d%H%M%S')`
- This function is callable as is (e.g., `get_version(local_scheme=get_local_node_and_timestamp)`) or as an entry point (e.g., `get_version(local_scheme='node-and-timestamp')`)
- Wrappers can be built around this function to provide custom timestamps, see examples below:

### Using Entry Point
```python
v = setuptools_scm.get_version(root='..', relative_to=__file__,
                               local_scheme='node-and-timestamp')
print(v)
```

### Using Callable
```python
v = setuptools_scm.get_version(root='..', relative_to=__file__,
                               local_scheme=get_local_node_and_timestamp)
print(v)
```

### Using a Format Wrapper
```python
def min_timestamp(version):
    return get_local_node_and_timestamp(version, fmt='%Y%m%d%H%M')


v = setuptools_scm.get_version(root='..', relative_to=__file__,
                               local_scheme=min_timestamp)
print(v)
```

```python
def micro_timestamp(version):
    return get_local_node_and_timestamp(version, fmt='%Y%m%d%H%M%S%f')


v = setuptools_scm.get_version(root='..', relative_to=__file__,
                               local_scheme=micro_timestamp)
print(v)
```

### DIFF
--- a/setuptools_scm/version.py
+++ b/setuptools_scm/version.py
@@ -144,6 +144,20 @@ def get_local_node_and_date(version):
         return version.format_choice("+{node}", "+{node}.d{time:%Y%m%d}")
 
 
+def get_local_node_and_timestamp(version, fmt='%Y%m%d%H%M%S'):
+    if version.exact or version.node is None:
+        return version.format_choice("",
+                                     "+d{time:"
+                                     + "{fmt}".format(fmt=fmt)
+                                     + "}")
+    else:
+        return version.format_choice("+{node}",
+                                     "+{node}"
+                                     + ".d{time:"
+                                     + "{fmt}".format(fmt=fmt)
+                                     + "}")
+
+
 def get_local_dirty_tag(version):
     return version.format_choice('', '+dirty')
 
--- a/setup.py
+++ b/setup.py
@@ -90,7 +90,8 @@ arguments = dict(
 
         [setuptools_scm.local_scheme]
         node-and-date = setuptools_scm.version:get_local_node_and_date
-        node-and-timestamp = setuptools_scm.version:get_local_node_and_timestamp
+        node-and-timestamp = \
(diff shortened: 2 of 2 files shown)


setuptools_scm/version.py
```diff
@@ -144,6 +144,20 @@ def get_local_node_and_date(version):
         return version.format_choice("+{node}", "+{node}.d{time:%Y%m%d}")
 
 
+def get_local_node_and_timestamp(version, fmt='%Y%m%d%H%M%S'):
+    if version.exact or version.node is None:
+        return version.format_choice("",
+                                     "+d{time:"
+                                     + "{fmt}".format(fmt=fmt)
```
RonnyPfannschmidt, line 151: this should use nested replacements as per https://www.python.org/dev/peps/pep-3101/#format-specifiers

this needs an api addition however

RonnyPfannschmidt: a follow-up issue was created

PR #204, opened 2017-12-21
fix build by pinning pytest

RonnyPfannschmidt:
pytest 3.3.x has bad pypi uploads

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,7 +8,7 @@ max-complexity = 10
 [testenv]
 usedevelop=True
 deps=
-    pytest
+    pytest<3.3.0 # pinned due to bad pypi uploads
 commands=
     test: py.test []
 


tox.ini
```diff
@@ -8,7 +8,7 @@ max-complexity = 10
 [testenv]
 usedevelop=True
 deps=
-    pytest
+    pytest<3.3.0 # pinned due to bad pypi uploads
```
kryskool, line 11: Comment broke test
```
line 96, in __init__
    requirement_string[e.loc:e.loc + 8]))
InvalidRequirement: Invalid requirement, parse error at "'# pinned'"
```

RonnyPfannschmidt: thanks for the note, but as far as i can tell pypi metadata was fixed so this pr is no longer needed to begin with

PR #207, opened 2017-12-25
Use samefile to compare if two paths refer to the same path.

jaraco:
After reviewing #174, I believe this change is a simpler and superior approach.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+v1.15.8
+======
+
+* #207: Re-use samefile backport as developed in jaraco.windows,
+  and only use the backport where samefile is not available.
+
 v1.15.7
 ======
 
--- a/setuptools_scm/git.py
+++ b/setuptools_scm/git.py
@@ -1,6 +1,6 @@
-from .utils import do_ex, trace, has_command, _normalized
+from .utils import do_ex, trace, has_command
 from .version import meta
-from os.path import isfile, join
+from os.path import isfile, join, samefile
 import warnings
 
 FILES_COMMAND = 'git ls-files'
@@ -21,7 +21,7 @@ class GitWorkdir(object):
         if ret:
             return
         trace('real root', real_wd)
-        if _normalized(real_wd) != _normalized(wd):
+        if not samefile(real_wd, wd):
             return
 
         return cls(real_wd)
--- a/setuptools_scm/git.py
+++ b/setuptools_scm/git.py
@@ -1,7 +1,13 @@
 from .utils import do_ex, trace, has_command
 from .version import meta
-from os.path import isfile, join, samefile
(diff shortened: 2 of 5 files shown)


jaraco:
I've uploaded a build of this PR to devpi.net/jaraco/dev. You should be able to test it out easily by setting an environment variable `PIP_INDEX_URL=https://devpi.net/jaraco/dev` and maybe `PIP_PRE=1`. As long as whatever package you're trying to install isn't pinned to a specific setuptools_scm version, I'd think it would pull in `1.15.7.dev14+gcf14597` as uploaded.

@woodyza and @timlehr and @anthrotype: Would you please give this build a run in your environment and report back your results?

anthrotype:
That would be great, but according to the python docs os.path.samefile is only available for Unix platform, not Windows. At least, for python 2.7...
https://docs.python.org/2/library/os.path.html

anthrotype:
It looks like Windows support was added since python 3.2
(note I haven’t tested this on a real Windows machine, i’ll report back once I find out)

anthrotype:
oh well, Appveyor CI beat me to it

https://ci.appveyor.com/project/RonnyPfannschmidt/setuptools-scm-r14ad/build/1.0.187/job/hl69mf7406p16jht#L78

> ImportError: cannot import name samefile

on Windows python 2.7 :cry:

jaraco:
Oh, good work. Yes. Python 2.7 should probably be supported (to my dismay).

In that case, I recommend setuptools_scm provide a backport of samefile (which may have the ctypes implementation removed herein, but limited to systems without os.path.samefile).

jaraco:
I did some research into how Python 3.2 implements samefile for Windows. It does it by re-using the same posixpath.samefile and by generating compatible `stat` results that compare equal when the same file is referenced.

And while it may be possible to create a compatibility library that generates compatible stat objects, there may be a less complicated way to implement samefile - by comparing [final paths](https://github.com/jaraco/jaraco.windows/blob/89d818506101f639c26634263038eab7d5e4c6b2/jaraco/windows/filesystem/__init__.py#L161-L201).

Either technique involves substantial ctypes support and multiple API calls, so it probably makes sense for this functionality to go in a separate module. I'll want to first prove (with tests) the functionality in a project like jaraco.windows and then port it to a module here.

timlehr:
@jaraco Sounds great! Once there is Python 2.7 support I’ll try it in my application and report back the results.

jaraco:
That change fixed the failing test on Python 2.7 and I've uploaded setuptools_scm-1.15.7.dev15+gc161ca2.zip to devpi. Please give that a try.

timlehr:
@jaraco Awesome, this solves the issue on W10 with Python 2.7. Would be great if this could be rolled out as an update asap. :)

PR #220, opened 2018-03-08
Fix #219

### LINKED ISSUE #219: Regression: hg repos with setuptools_scm>=1.15 are missing dev version bumps
`setuptools_scm` 1.15.* is reporting incorrect version numbers for my hg repositories.

Specifically, a merge commit that follows a tag commit does not seem to trigger the version bump like I would expect. In the screenshot below, for instance, I would expect the tip to be version `1.1.dev[N]+[sha]`. However, versions 1.15.* report the tip's version as `1.0`.

To reproduce:
```bash
mkdir hgmergetest && cd hgmergetest && hg init

# create initial commit and tag it
touch a && hg add a && hg commit -m "initial commit" 
hg tag 1.0

# create a branch
hg branch branch1
touch b && hg add b && hg commit -m "create branch1"

# merge the branch into default
hg update default && hg merge branch1 && hg commit -m "merge branch1 into default"
```

Expected behavior (setuptools_scm 1.13 and 1.14):
```bash
> pip install "setuptools_scm<1.15"
> python -m setuptools_scm
Guessed Version 1.1.dev3+n8dce1535e70a
```

Behavior with `1.15.*`:
```bash
> pip install setuptools_scm==1.15.7
> python -m setuptools_scm
Guessed Version 1.0
```

image

avirshup:
Summary:
 - issue is isolated and tested in 3edb8d7
 - added additional tests to guard against side effects from this fix (commits 966898c through c1cf286)
 - actually fixed the issue in c3f35e6

### DIFF
--- a/setuptools_scm/hg.py
+++ b/setuptools_scm/hg.py
@@ -8,8 +8,14 @@ FILES_COMMAND = 'hg locate -I .'
 def _hg_tagdist_normalize_tagcommit(root, tag, dist, node):
     dirty = node.endswith('+')
     node = 'h' + node.strip('+')
-    revset = ("(branch(.) and tag({tag!r})::. and file('re:^(?!\.hgtags).*$')"
-              " - tag({tag!r}))").format(tag=tag)
+
+    # Detect changes since the specified tag
+    revset = ("(branch(.)"  # look for revisions in this branch only
+              " and tag({tag!r})::."  # after the last tag
+              # ignore commits that only modify .hgtags and nothing else:
+              " and (merge() or file('re:^(?!\.hgtags).*$'))"
+              " and not tag({tag!r}))"  # ignore the tagged commit itself
+              ).format(tag=tag)
     if tag != '0.0':
         commits = do(['hg', 'log', '-r', revset, '--template', '{node|short}'],
                      root)
--- a/testing/test_mercurial.py
+++ b/testing/test_mercurial.py
@@ -134,6 +134,13 @@ def test_version_bump_before_merge_commit(wd):
     assert wd.version.startswith('1.1.dev1+')
 
 
+@pytest.mark.issue(219)
+@pytest.mark.usefixtures("pre_merge_commit_after_tag")
+def test_version_bump_from_merge_commit(wd):
+    wd.commit()
+    assert wd.version.startswith('1.1.dev3+')  # issue 219
+
+
 @pytest.mark.usefixtures("version_1_0")
 def test_version_bump_from_commit_including_hgtag_mods(wd):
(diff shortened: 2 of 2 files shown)


avirshup:
Here's my stab at a fix for #219. Please let me know if this seems reasonable and/or if there are any changes that should be made.

setuptools_scm/hg.py
```diff
@@ -8,8 +8,14 @@
 def _hg_tagdist_normalize_tagcommit(root, tag, dist, node):
     dirty = node.endswith('+')
     node = 'h' + node.strip('+')
-    revset = ("(branch(.) and tag({tag!r})::. and file('re:^(?!\.hgtags).*$')"
-              " - tag({tag!r}))").format(tag=tag)
+
+    # Detect changes since the specified tag
+    revset = ("(branch(.)"  # look for revisions in this branch only
+              " and tag({tag!r})::."  # after the last tag
```
RonnyPfannschmidt, line 14: nice enhancement :+1:

testing/test_mercurial.py
```diff
@@ -108,3 +108,48 @@ def test_version_in_merge(wd):
...
+    wd('hg branch testbranch')
+    wd.write('branchfile', 'branchtext')
+    wd(wd.add_command)
+    wd.commit()
+    wd('hg update default')
+    wd('hg merge testbranch')
+    return wd
+
+
+def test_version_bump_before_merge_commit(pre_merge_commit_after_tag):
```
RonnyPfannschmidt: the fixture naming interaction seems a bit unnatural to me since those fixtures are actions that build up the content of a different fixture

i would propose to always add the wd fixture to the argument set, and to use the use-fixtures marker in tests

RonnyPfannschmidt (commented):
this is a nice iterative enhancement,

i wonder if the history should be squashed before merge

testing/test_mercurial.py
avirshup: Oh, I hadn't seen the usefixtures marker before - will change it, that's much clearer 👍

avirshup:
Thanks. Cleaned up the test setup; rebase incoming.

avirshup:
Hi @RonnyPfannschmidt. I rebased into 2 commits:
 - 5b59b07 tests some additional edge cases, just to make sure the bug fix doesn't break something else (they should be passing _before and after_ the bugfix)
 - 4263f57 actually fixes the bug

But I can also just squash the whole thing if you think that's cleaner.

RonnyPfannschmidt:
thank ofr the cleanup, looks good to me, thanks

PR #226, opened 2018-03-19
Fix support for old version of setuptools

pikacic:
After 47934061 I'm getting failures when using old versions of setuptools.

When `VERSION_CLASS` is `None` (old setuptools), line 54 fails with
```
  File ".../setuptools_scm/version.py", line 54, in tag_to_version
TypeError: isinstance() arg 2 must be a class, type, or tuple of classes and types
```

This change should fix the issue.


RonnyPfannschmidt:
good find, this was an oversight on my side

PR #227, opened 2018-03-19
git: ignore directory of the git archive output

sileht:
Since 1.16.0, if the git project have subdirectory, bdist fails.

For example:

running bdist_wheel
running build
running build_py
running egg_info
writing requirements to gnocchi.egg-info/requires.txt
writing gnocchi.egg-info/PKG-INFO
writing top-level names to gnocchi.egg-info/top_level.txt
writing dependency_links to gnocchi.egg-info/dependency_links.txt
writing entry points to gnocchi.egg-info/entry_points.txt
reading manifest template 'MANIFEST.in'
writing manifest file 'gnocchi.egg-info/SOURCES.txt'
error: can't copy 'gnocchi/cli': doesn't exist or not a regular file

gnocchi/cli is a directory here.

This change ignores directory of "git archive" output to fix the issue.

### DIFF
--- a/setuptools_scm/git.py
+++ b/setuptools_scm/git.py
@@ -129,8 +129,9 @@ def _list_files_in_archive():
     cmd = ['git', 'archive', 'HEAD']
     proc = subprocess.Popen(cmd, stdout=subprocess.PIPE)
     tf = tarfile.open(fileobj=proc.stdout, mode='r|*')
-    for name in tf.getnames():
-        print(name)
+    for member in tf.getmembers():
+        if member.type != tarfile.DIRTYPE:
+            print(member.name)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


devkral:
there is a better solution, wait a moment

devkral:
sry, I was wrong. Your solution is better (export-ignore feature).

setuptools_scm/git.py
```diff
@@ -130,7 +130,8 @@ def _list_files_in_archive():
     proc = subprocess.Popen(cmd, stdout=subprocess.PIPE)
     tf = tarfile.open(fileobj=proc.stdout, mode='r|*')
     for name in tf.getnames():
-        print(name)
+        if not isdir(name):
```
RonnyPfannschmidt: instead of checking for `isdir` here, i believe loop should be `for member in tf.getmembers()`

`member.type` should not equal `tarfile.DIRTYPE`

RonnyPfannschmidt (changes requested):
good work at identifying the cause and making a pr :+1:

setuptools_scm/git.py
sileht: Indeed, that looks better, I have done that.

RonnyPfannschmidt (approved):

RonnyPfannschmidt:
will release in a few minute - thanks again :+1:

PR #233, opened 2018-03-20
git: Use a method instead of a command for listing files

### LINKED ISSUE #232: since 1.16.0 setuptools_scm can't work if only installed via setup_requires
Hi,

If setuptools_scm is in setup_requires, and the virtualenv or the operating system doesn't have setuptools_scm installed, the sdist/bdist is generated correctly.

Also we can sse this warning:

/home/tester/src/.tox/py35-postgresql-file-upgrade-from-4.2/bin/python3.5: Error while finding spec for 'setuptools_scm.git' (ImportError: No module named 'setuptools_scm')

This warning should be fatal, since the generated sdist/bdist will not work.

I'm guessing this is due to the installation done by easy_install. The console_script that run "git archive" will be not available.

sileht:
We can't use "python -m" because if setuptool_scm is install with
easy_install (e.g.: via setup_requires), the module will not be found.

This change use a method instead of a command.

Fixes #232

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,7 +82,7 @@ arguments = dict(
 
         [setuptools_scm.files_command]
         .hg = setuptools_scm.hg:FILES_COMMAND
-        .git = setuptools_scm.git:FILES_COMMAND
+        .git = setuptools_scm.git:list_files_in_archive
 
         [setuptools_scm.version_scheme]
         guess-next-dev = setuptools_scm.version:guess_next_dev_version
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt:
good job, also OMG, i missed that the surrounding tooling for making it a function was already in place, i could have completely avoided that regression with a little more care earlier

RonnyPfannschmidt (approved):

PR #234, opened 2018-03-20
stop if find_files fail

sileht:
We must reraise any expection during find_files instead
of generating an almost empty sdist/bdist.

Related #232


RonnyPfannschmidt:
good work :+1: 

this one will require a v1.17.0 and a prominent change-log note since it may break certain builds that "worked"

RonnyPfannschmidt (approved):

Issue #229, opened 2018-03-19
Git and mercurial disagree on the latest tag


avirshup:
Currently, [`setuptools_scm.git` only pays attention to git tags that contain a "`.`" by using](https://github.com/pypa/setuptools_scm/blob/b7182bad424848e86abf1d747fad8327d7933920/setuptools_scm/git.py#L1): 
```
git describe --dirty --tags --long --match *.*
```

OTOH, [`setuptools_scm.hg` considers _all_ tags by running](https://github.com/pypa/setuptools_scm/blob/b7182bad424848e86abf1d747fad8327d7933920/setuptools_scm/hg.py#L50):
```
hg parents --template "{latesttag} {latesttagdistance}
```

So, for instance, the tag `some-random-tag` is ignored in a git repository, but is used (and causes an error upon parsing) in a mercurial repo.

### To replicate (in an existing repository):
(I'm using `setuptools_scm 1.16.1`)

With git:
```
> git tag 1.0.0
> touch tempfile && git add tempfile && git commit tempfile -m "commit tempfile"
> git tag some-random-tag
>
> python -m setuptools_scm
Guessed Version 1.0.1.dev1+ngffe9b95
```

With hg:
```
> hg tag 1.0.0
> touch tempfile && hg add tempfile && hg commit tempfile -m "commit tempfile"
> hg tag some-random-tag
>
> python -m setuptools_scm
Traceback (most recent call last):
[...]
  File "[...]/setuptools_scm/version.py", line 86, in meta
    assert tag is not None, 'cant parse version %s' % tag
AssertionError: cant parse version None
```

(Fixes proposed below)

avirshup:
I can submit a PR to fix this either by:

1) replicating the git behavior in `setuptools_scm.hg` (which, given common naming conventions, works for many but definitely not all use cases), OR
2) adding a function that finds correct right tag using a more generalized test (say, by trying to normalize the tag into a PEP440 version using `pkg_resources`) for both `git` _and_ `hg`

Would either of these make sense?

RonnyPfannschmidt:
both make sense, i'm not sure which makes more sense, but i believe we should start with 1. for consistency and note 2. for a follow-up at a later point

avirshup:
Yeah, that seems sensible - I'll submit a PR for 1) and open a new issue for 2). Thanks!

RonnyPfannschmidt:
closing as the pr and the followup are issued

Issue #235, opened 2018-03-20
git and hg: ignore tags that can't be normalized to version numbers


avirshup:
As discussed in #229, when determining the latest tag, it would be ideal to use a general, unambiguous test for whether or not to ignore a given SCM tag.

For instance, a commit marked with tag `v2.1.0a4` should, of course, be considered version 2.1.0a4. A commit tagged only with `my.awesome.tag` should be ignored (and definitely NOT parsed as `major='my', minor='awesome', patch='tag'`).

After dealing with #229, I'll add a more complete spec here for how this test should work. My preference is to delegate to an existing tool that normalizes version strings according to PEP440.

reece:
I've recently started getting errors like 
`AssertionError: cant parse version default/1.2.6.dev1`
in several repos. 

(I believe this tag was generated by hg-git (unconfirmed). In any case, I did make the `1.2.6.dev1` tag, but not with the `default/` prefix.)

Regardless of the cause, it seems that setuptools_scm does not deal gracefully with tags that look like version numbers but aren't. Currently, it appears that all tags with a `.` are deemed to be version tags (in hg.py:get_latest_normalizable_tag()), which is clearly too permissive.

RonnyPfannschmidt:
After carefully reconsidering im closing this for now 

There's some tooling in the pipeline to remove ambiguous situations but in general  the tools are beholden to the describe/identify commands a scm supplies

PR #236, opened 2018-03-20
hg: use same logic as git for finding latest tagged commit

avirshup:
This is a fix for #229

### DIFF
--- a/setuptools_scm/hg.py
+++ b/setuptools_scm/hg.py
@@ -46,20 +46,32 @@ def parse(root):
         trace('initial node', root)
         return meta('0.0', dirty=dirty)
 
-    # the newline is needed for merge stae, see issue 72
-    cmd = 'hg parents --template "{latesttag} {latesttagdistance}\n"'
-    out = do(cmd, root)
     try:
-        # in merge state we assume parent 1 is fine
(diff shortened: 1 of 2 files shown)


avirshup:
This PR ignores hg tags that don't contain a '.' in them (similarly to the way git.py uses `git describe`).

I've created a couple new helper functions in `hg.py` to do this -
 - `get_graph_distance` for computing the distance between two revisions in mercurial, and
 - `get_latest_normalizable_tag`, for returning the latest "version-like" hg tag. Right now, it just inelegantly filters for the presence of '.' in the tag.  But it will likely be rewritten by a more general test in the fix for #235.

RonnyPfannschmidt:
sorry for the delay, i have been on travel and didnt get to sorting it before, i plan to sort the backlog on setuptools_scm until the weekend, thanks for the patience

RonnyPfannschmidt:
thanks :+1:

avirshup:
Thanks - and no worries whatsoever about the timing, open source maintainers deserve vacations too :)

RonnyPfannschmidt:
i plan to do a release with support for branches added and support for ancient broken setuptools dropped, please take a look

PR #239, opened 2018-04-05
Improve development version in semver

ggtools:
When dealing with semantic versioned code, the original behaviour was to
increment the latest digit. For instance: v1.2.0 gives v1.2.1.dev1. This
commit implements something that might look more logical in semver which
is to increase the minor rather than the patch: v1.2.0 -> v1.3.0.dev1

### DIFF
--- a/setuptools_scm/version.py
+++ b/setuptools_scm/version.py
@@ -1,13 +1,14 @@
 from __future__ import print_function
+
 import datetime
-import warnings
 import re
-from .utils import trace
+import warnings
+from distutils import log
 
 from pkg_resources import iter_entry_points
-
-from distutils import log
 from pkg_resources import parse_version
+
+from .utils import trace
 
 
 def _get_version_class():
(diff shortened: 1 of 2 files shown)


ggtools:
To give more details about the motivation behind this PR. I'm working on projects using sementic versioning. We have a *main* branch which is alway deliverable. Minor/major releases are done by creating a release branch from the *main* branch as late as possible. Patches are done on the release branches, most of the time by cherry picking commits from the *main* branch onto the release branch.

With the current algorithm which increase the rightmost number leads to development versions on the *main* branch to be *lower* than the patch releases from the release branch.

RonnyPfannschmidt:
at first glance it would only be acceptable to do this as a opt-in in some manner, since this change would get version wrong for at least a dozen of projects im aware of/working on

before i was working on a way to include the branch name in the available meta-data, which would allow to pick a valid increment based on branch names - i believe that approach would also be helpful in your case

ggtools:
Yep although I try to limit it to semver version but I didn't find a natural way to add a configuration switch to enable it. 

Branch in the metadata would definitely be an improvement.

RonnyPfannschmidt:
@ggtools please take a look at #240

RonnyPfannschmidt:
@ggtools im closing this one in favour of #240 - based on what you wrote before i believe the simplified semver support i add in that pr also fits your use-case, but please correct me if i miss-understood

PR #240, opened 2018-04-06
add support for branch name based testing, add simplified semver support

RonnyPfannschmidt:


testing/test_mercurial.py
```diff
@@ -114,7 +114,7 @@ def test_parse_no_worktree(tmpdir):
 def version_1_0(wd):
     wd('hg branch default')
     wd.commit_testfile()
-    wd('hg tag 1.0 -u test -d "0 0"')
+    wd('hg tag 1.0.0 -u test -d "0 0"')
```
jaraco, line 117: What's happening here? Do mercurial tags now require full semver and drop support for N.N numbers?

testing/test_version.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+from setuptools_scm.version import meta, simplified_semver_version
+
+
+@pytest.mark.parametrize('version, expected_next', [
+    pytest.param(meta("1.0.0"), "1.0.0", id='exact'),
+
+    pytest.param(meta("1.0"), "1.0.0", id='short_tag'),
```
jaraco, line 8: Does test imply that a tag of "1.0" generates a version of "1.0.0"? I'm unsure what the implications are of that. Is this an opt-in feature?

setuptools_scm/version.py
```diff
@@ -145,6 +148,31 @@ def guess_next_dev_version(version):
...
+    if distance:
+        parts.append("dev" + str(distance))
+    return '.'.join(parts)
+
+
+def simplified_semver_version(version):
+    if version.exact:
+        return guess_next_simple_semver(version.tag, 0, 3, increment=False)
+    else:
+        if version.branch is not None and 'feature' in version.branch:
```
jaraco: Given that these two code branches are 90% the same, I'd write it as:
```
retain = 2 + (version.branch is not None and 'feature' in version.branch)
return guess_next_simple_semver(version.tag, version.distance, retain=retain)
```

setuptools_scm/version.py
```diff
@@ -145,6 +148,31 @@ def guess_next_dev_version(version):
...
+    del parts[3:]
+    if distance:
+        parts.append("dev" + str(distance))
+    return '.'.join(parts)
+
+
+def simplified_semver_version(version):
+    if version.exact:
+        return guess_next_simple_semver(version.tag, 0, 3, increment=False)
+    else:
```
jaraco: This else is unneeded.

setuptools_scm/version.py
```diff
@@ -145,6 +148,31 @@ def guess_next_dev_version(version):
         return guess_next_version(version.tag, version.distance)
 
 
+def guess_next_simple_semver(version, distance, retain, increment=True):
+    parts = str(version).split('.', retain)
```
jaraco: These two lines use retain twice for the same meaning. Consider replacing with
```
parts = str(version).split('.', retain)[:-1]
```

setuptools_scm/version.py
```diff
@@ -145,6 +148,31 @@ def guess_next_dev_version(version):
...
+    del parts[retain:]
+    while len(parts) < retain:
```
jaraco: These next two lines are padding the list of version numbers with zeros. I'd either document that with a comment and whitespace or extract a function whose docstring described the intention.

setuptools_scm/version.py
```diff
@@ -145,6 +148,31 @@ def guess_next_dev_version(version):
...
+        parts.append("0")
+    if increment:
+        parts[-1] = str(int(parts[-1]) + 1)
```
jaraco: It feels a little awkward here to be working with strings as numbers. It seems it'd be preferable to me to parse everything as numbers, and then convert back to a string at the end. But that may not be possible if some of the parts are meant to be non-numbers.

jaraco (approved):
The hardest part for me in reviewing this change is there's no documentation, so I'm not really sure what was intended. I only have one point of reference and that's "what changed". The topic of the PR suggests that two changes are happening here, "support for branch name based testing" and "add simplified semver support". Perhaps that suggests this should be two changes.

My biggest concern is about the apparent rewriting of N.N numbers to N.N.0, which may cause unexpected behavior and consternation for projects that intentionally release two-part version numbers. I'd like to know what is the story for such projects?

Otherwise, these changes seem pretty sane. I made a few comments about how I'd improve it, but overall, after spending 20-30 minutes on it, I think I understand what's going on and it seems fine.

testing/test_version.py
RonnyPfannschmidt: this is opt-in, one has to specifiy simplified semver

setuptools_scm/version.py
RonnyPfannschmidt: @jaraco this needs a different expression perhaps, bur the `[:-1]` is wrong, as t fails for versions where the parts contain less than retain elements

setuptools_scm/version.py
RonnyPfannschmidt: unfortunately its permissible for `parts[0]` to be `'v123'`

gerritholl:
How do I use this?

RonnyPfannschmidt:
@gerritholl by setting the version scheme to python-simplified-semver

I can elaborate once I get back to the computer later today

wookayin:
pyproject.toml
```diff
 [tool.setuptools_scm]
 write_to = ...
+version_scheme = "python-simplified-semver"
```

The most important line is
```
if version.branch is not None and "feature" in version.branch:
```

so I think this version scheme bumps minor version when the branch name contains `feature`. (Not possible when the branch name is `main`, `master`, or `dev` ??)

Issue #251, opened 2018-04-28
file finder fails when not run from the top level of the SCM clone


sbidoul:
When setup.py is not at the root of the SCM directory (ie where .git or .hg reside), the file finders are disabled.

This is due to the discovery mechanism https://github.com/pypa/setuptools_scm/blob/a99467848bec1782b8dd2ea32259da21b3a94a39/setuptools_scm/discover.py#L6-L13

RonnyPfannschmidt:
i recall roughly that the reason for this is partly that file finders dont get a distirubtion object,
so at file finding time we have no idea what the surrounding context and/or an acceptable root folder is

sbidoul:
In #252 I assume `git rev-parse --show-toplevel` or `hg root` when run in `path` provides an acceptable root folder. This sounds a reasonable extension compared to looking for the presence of `.git` or `.hg`  in `path`.

RonnyPfannschmidt:
:+1:

Issue #248, opened 2018-04-27
file finder is not 100% spec compliant


sbidoul:
It replaces an empty path with '.', while the [setuptools file finder spec](http://setuptools.readthedocs.io/en/latest/setuptools.html#adding-support-for-revision-control-systems) explicitly warns against it.

https://github.com/pypa/setuptools_scm/blob/e4f80b9f248aea2cfea92a9fc2e2ace5d4584b9c/setuptools_scm/integration.py#L25-L27

This is a minor issue that does not seem to have any negative impact in practice.

RonnyPfannschmidt:
good find, thanks

PR #252, opened 2018-04-28
File finders spec compliance, work in subdirectories too

sbidoul:


RonnyPfannschmidt:
thanks for adapting the tests, since there is no fix yet it would hep to keep those xfail for now

sbidoul:
So while I was in this rabbit hole, I went to the bottom of it :)

So this is now
- a fix for #251 
- a fix for #248 
- a refactoring of the file finders so hg and git use the same algorithm (and the same test suite)

RonnyPfannschmidt:
fabulous work as is, i'm happy to merge :+1:
im currently on a vacation but i'll try to get this released soon as it is quite an enhancement

sbidoul:
@RonnyPfannschmidt Thanks for the review and merge, and the kind words.

Perhaps releasing a beta 2.1 first would be best? Although I've been very careful and tried to think about many tests, there is still a slight chance that the new algorithm would exclude some files that were included by the previous version. Not sure how `setup_requires` works with beta releases though.

Enjoy the holidays.

RonnyPfannschmidt:
beta releases wont work for setup_requires in a discoverable way , in my experience its better to pull a release if a major issue is found rather than having a beta that nobody uses

sbidoul:
ok, go for a release then

We have pretty advanced use cases in the Odoo community so if anything goes wrong, there is a fair chance it will show up there first (and I did test these quite a lot already).

RonnyPfannschmidt:
:+1:

PR #253, opened 2018-05-10
file finder: symlink edge cases

sbidoul:
It turns out we had a edge case to handle, as the algorithm
included scm controlled files reachable through symlinks while the
symlink itself was not scm controlled.

Additionally, better handle situations where we include a directory
below 'path' as well as a symlink to it. Before it would randomly
return the original directory, or include it under the symlink.

### DIFF
--- a/setuptools_scm/file_finder.py
+++ b/setuptools_scm/file_finder.py
@@ -8,6 +8,7 @@ def scm_find_files(path, scm_files, scm_dirs):
     - scm_files: set of scm controlled files and symlinks
       (including symlinks to directories)
     - scm_dirs: set of scm controlled directories
+      (including directories containing no scm controlled files)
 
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt:
thanks for providing this, is this just a inconvenience, or is this a potential security issue that we should report?

sbidoul:
IMO it's just an inconvenience. At worst it would have included too many files, but still from the same github repo. And 2.1 has not been released anyway, so no one should have been impacted.

RonnyPfannschmidt:
thanks for the summary - great job :+1:

sbidoul:
Thanks. Do you have plans to cut a release soon?

RonnyPfannschmidt:
i'll try to today

PR #255, opened 2018-05-13
inflict black on the codebase :)

RonnyPfannschmidt:
- [x] initial black run
- [x] add post commit hook
- [-] add travis hook to check/ensure compliance


RonnyPfannschmidt:
@jaraco btw, about when can we remove that py31_win workaround?

jaraco:
> about when can we remove that py31_win workaround?

When setuptools_scm drops support for Python 2.7. I'm pretty sure the appveyor tests would capture this expectation.

jaraco:
Blarg. Black converts all quotes to double-quotes (`"`)? My understanding from other linters was that single quotes were preferred (cherrypy uses a pre-commit's double-quote-string-fixer). I prefer single quotes because they're unchorded in my keyboard (one keypress instead of two coordinated). I guess these projects will have to fight over what's the right way to write a string literal.

RonnyPfannschmidt:
@jaraco wrt quoting - you still can write them however you like, just run black before commit ^^

on my keyboard any quote needs 2 stokes and i did also prefer single quotes - but i much pefer having a tool sort out all code formatting over having slight inconsistencies from developers or different preferences over time

PR #256, opened 2018-05-13
Globally exclude the vendored `win_py31_compat.py` file

asottile:


RonnyPfannschmidt:
@asottile neat - thanks for that trick - can you point me where its in the docs i missed it (i did search for exclude but only found per hook)

asottile:
Yep!  It's right here at the top: https://pre-commit.com/#pre-commit-configyaml---top-level

It's a ~relatively new feature so it might not've been there when you last looked.

RonnyPfannschmidt:
nah, i really completely missed it and i'm really glad you sorted it out directly :+1: 

thanks

PR #270, opened 2018-06-11
Add `node-and-timestamp` to README.rst

sTywin:
It was not obvious that `node-and-timestamp` was an available implementation for `local_scheme`. This PR adds it to the list of implementations in `README.rst`.


RonnyPfannschmidt:
thanks for addressing that oversight !!

Issue #273, opened 2018-06-30
document a get_version pattern for editable installs staying as correct as sensible


blueyed:
Is the following reasonable, e.g. for output of `--version`?

I am using this to get the current version with editable installs, useful
during development:

```python
def get_version():
    try:
        from setuptools_scm import get_version
        return get_version(root='..', relative_to=__file__)
    except (ImportError, LookupError):
        from pkg_resources import get_distribution
        return get_distribution(__package__).version
```

setup.py uses the basic:

```python
    use_scm_version=True,
    setup_requires=['setuptools_scm'],
```

btw: what is a use case for `write_to`?  The file itself says that the file
should not be tracked in the SCM itself.

Ref: https://github.com/Vimjas/covimerage/pull/43

RonnyPfannschmidt:
@blueyed ´write_to` is for when people dont want to pay the cost of `pkg_resources.get_distribution`

personally i think its perfectly valid to work with stale metadata on development, just doing semi-regular `pip install -e ...`

blueyed:
Thanks for clarification!

blueyed:
Closing.
README update in #274.

majuscule:
Just want to voice my support for availability of access to updated version metadata from within an editable install. I work from within an editable install constantly, and having the version data up to date would be very helpful to avoid confusion when interacting with others who are using experimental releases of my packages. I don't feel that forcing developers to remember to repeatedly `pip install ...` is a valid solution.

Please reopen this issue.

taldcroft:
For the `get_version()` function shown in the description, will the initial attempt to run `setuptools_scm.get_version(..)` be lightweight / fast?  In particular that it doesn't actually end up running a `git` executable if there is no git repo in the installed package directory.  I looked into the code a bit but confess to finding the code that detects the repo to be a bit inscrutable.

Running in debug mode I see for a successful detection it says:
```
found ep .git = setuptools_scm.git:parse  # found a .git directory??
cmd 'git rev-parse --show-toplevel'  # runs a git command now
...
```
While for unsuccessful:
```
root '/Users/aldcroft/git/sparkles'
looking for ep setuptools_scm.parse_scm /Users/aldcroft/git/sparkles
looking for ep setuptools_scm.parse_scm_fallback /Users/aldcroft/git/sparkles
Traceback (most recent call last):
  File "setup.py", line 21, in <module>
    cmdclass=cmdclass,
```
So it all looks good, but again just don't know what's going on under the hood.

[...]

PR #274, opened 2018-06-30
README: add use case for write_to

blueyed:
Ref: https://github.com/pypa/setuptools_scm/issues/273#issuecomment-401555022

[ci skip]


RonnyPfannschmidt:
fabulous, thanks :+1:

Issue #275, opened 2018-06-30
tags with dashes still broken


punkadiddle:
In #266 you tried to make setuptools_scm more compatible to accept [Semver](https://semver.org/) version tags like "v1.2.3-rc.1". The regex used to strip the prefix keeps it broken since it greedyly consumes everything including a version until the dash.

setuptools_scm should accept everything pkg_resources takes. I don't want to make this a world vs python issue but in my environment we use several languages and tagging them all differently is not acceptable.

The following works flawlessly:
```
In [2]: pkg_resources.parse_version('v1.0.1-rc.1')
Out[2]: <Version('1.0.1rc1')>
```

Line to strip prefix:
```
TAG_PREFIX = re.compile(r"^\w+-(.*)")
```

Maybe you could allow to configure a constant prefix / regexp to strip? That would make it easy to adapt to the environment.

RonnyPfannschmidt:
I fixed a bug with actual python version numbers, semver is not supported in detail

RonnyPfannschmidt:
I would accept a PR implementing it

RonnyPfannschmidt:
But to be clear, this has to be a opt in, as its non python versions

punkadiddle:
see #276

PR #276, opened 2018-06-30
configurable version number extraction from tag strings

punkadiddle:
As suggested in #275 a modification to use regex to extract the version information from tags. Regex can be changed from the configuration to adapt to tagging schemes more flexible.

Extracted the whole configuration to a singleton class to have it available without passing all params through the stack and reduce the number of changes. Don't think there will be situations where different configurations would be used at the same time, do you?

Tested on Ubuntu using 2.7 and 3.6.

What do you think?


punkadiddle:
removed editor configuration

src/setuptools_scm/utils.py
```diff
@@ -103,3 +105,18 @@ def has_command(name):
     if not res:
         warnings.warn("%r was not found" % name)
     return res
+
+
+def singleton(cls):
```
RonnyPfannschmidt: this kind of singleton is NOT acceptable

testing/test_version.py
```diff
@@ -37,3 +38,17 @@
...
+@pytest.mark.parametrize(
+    "tag, expected",
+    [
+        pytest.param("v1.0.0", "1.0.0"),
+        pytest.param("v1.0.0-rc.1", "1.0.0rc1"),
+        pytest.param("v1.0.0-rc.1+-25259o4382757gjurh54", "1.0.0rc1"),
+    ],
+)
+def test_tag_regex1(tag, expected):
+    Configuration().tag_regex = r'^(?P<prefix>v)?(?P<version>[^\+]+)(?P<suffix>.*)?$'
```
RonnyPfannschmidt, line 52: the reason why the singleton is not acceptable demonstrates in this test - if it was executed as first test, - all tests would use the different configuration

src/setuptools_scm/version.py
```diff
@@ -61,23 +77,23 @@ def tag_to_version(tag):
...
-    # lstrip the v because of py2/py3 differences in setuptools
-    # also required for old versions of setuptools
-    prefix_match = TAG_PREFIX.match(tag)
-    if prefix_match is not None:
-        version = prefix_match.group(1)
-    else:
-        version = tag
+
+    tagdict = _parse_version_tag(tag)
+    if tagdict is None or len(tagdict['version']) < 1:
```
RonnyPfannschmidt: this should use a boolean test even for the string

src/setuptools_scm/config.py
```diff
@@ -0,0 +1,49 @@
...
+import warnings
+
+from .utils import singleton
+
+DEFAULT_TAG_REGEX = r'^(?P<prefix>\w+-)?(?P<version>v?\d+(?:\.\d+){0,2}[^\+]+)(?P<suffix>\+.*)?$'
+TAG_REGEX_EXPECTED_GROUPS = set(['prefix', 'version', 'suffix'])
+
+
+@singleton
+class Configuration(object):
```
RonnyPfannschmidt: i'm really unhappy about the way this configuration class works

src/setuptools_scm/config.py
```diff
@@ -0,0 +1,49 @@
...
+    """ Global Configuration single-instance class """
+
+    root = ''
```
RonnyPfannschmidt: all of those should go into instance variables at least

src/setuptools_scm/config.py
```diff
@@ -0,0 +1,49 @@
+""" configuration """
+from __future__ import print_function, unicode_literals
+import re
+import warnings
+
+from .utils import singleton
+
+DEFAULT_TAG_REGEX = r'^(?P<prefix>\w+-)?(?P<version>v?\d+(?:\.\d+){0,2}[^\+]+)(?P<suffix>\+.*)?$'
```
RonnyPfannschmidt: the strict requirement for prefix/suffix as groups makes the parsing needlessly complex for other users,

src/setuptools_scm/config.py
```diff
@@ -0,0 +1,49 @@
...
+        self.local_scheme = "node-and-date"
+        self.tag_regex = DEFAULT_TAG_REGEX
+
+    @property
+    def tag_regex(self):
+        return self._tag_regex
+
+    @tag_regex.setter
+    def tag_regex(self, value):
+        if value is None:
```
RonnyPfannschmidt: None shouldn't be a magic constant

src/setuptools_scm/config.py
```diff
@@ -0,0 +1,49 @@
...
+            value = DEFAULT_TAG_REGEX
+        regex = re.compile(value)
+
+        group_names = set(regex.groupindex.keys())
```
RonnyPfannschmidt: the compilation/validation/warnings should be a utility function that's unit-tested and invoked here

RonnyPfannschmidt (changes requested):
overall this is a good start in making the regex configurable, but the implementation as singleton as is isn't acceptable as it litters the codebase with time-bombs

src/setuptools_scm/version.py
```diff
@@ -148,10 +164,15 @@ def _parse_tag(tag, preformatted):
 
 
 def meta(tag, distance=None, dirty=False, node=None, preformatted=False, **kw):
-    parsed_version = _parse_tag(tag, preformatted)
-    trace("version", tag, "->", parsed_version)
-    assert parsed_version is not None, "cant parse version %s" % tag
-    return ScmVersion(parsed_version, distance, node, dirty, preformatted, **kw)
+    if tag is not None:
```
RonnyPfannschmidt: this is a api-breaking change that pushes a breakage down into unsuspecting code

testing/test_version.py
punkadiddle: You are right. Would it be acceptable to keep the class instance and push it through all the way from get_version to meta instead?

src/setuptools_scm/config.py
punkadiddle: Easy to drop. I did that to have those parts available for reconstruction and because you have an explicit warning for dropping the suffix part which would need to go otherwise.
Shall I remove prefix and suffix then?

testing/test_version.py
RonnyPfannschmidt: that seems to be the only way (and one of the reasons why i avoided implementing myself

also its not clear if that config object should be everywhere

src/setuptools_scm/config.py
RonnyPfannschmidt: the explicit warnigns are still needed, its just not clear what to do with the groups and regex requirements off hand

RonnyPfannschmidt:
ps extra note, im currently on a vacation and not tracking everything in the same diligence as usual -

testing/test_version.py
punkadiddle: Had a look into it. Not that difficult, but configuration needs to be passed to any scanner's `parse` entry points on to the version parser `meta`. Parse doesn't have kwargs. Even if optional it would be a change to an API your exposing. Config would be made available to both scanners and parser.

Separating parsers/scanners by passing back scanner results and then parse them from the main routine would probably the better approach from an architectural view. That would require even more refactoring but would allow keep the entry points stable until you would want to configure those further.

I was thinking about doing a trimmed down fork to only read version information borrowing your scanners as an alternative. Versioning being handled by the build-systems seems to be the preferred approach today.

src/setuptools_scm/version.py
punkadiddle: Sorry, unrelated and wasn't meant to go into this pull request.

src/setuptools_scm/config.py
punkadiddle: Just found that it's (of course) possible to look before and after a group match to get prefix and suffix. Ergo those groups can be dropped entirely without loosing their information.

testing/test_version.py
RonnyPfannschmidt: im under the impression that some more design work is needed

note that setuptools_scm is intended as a entry in the build-systems requires table ^^

src/setuptools_scm/config.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _check_absolute_root(root, relative_to):
+    if relative_to:
+        if os.path.isabs(root) and not root.startswith(relative_to):
+            warnings.warn("absolute root path '%s' overrides relative_to '%s'" % (root, relative_to))
+        root = os.path.join(os.path.dirname(relative_to), root)
+    return os.path.abspath(root)
+
+
+class Configuration(object):
```
RonnyPfannschmidt, line 34: part of me wonders if it shouldn't be ok to just make Configuration immutable
in order to avoid having a set of properties that need to be smart,

this thought is for a followup however and mainly food forthought

src/setuptools_scm/git.py
```diff
@@ -82,20 +83,23 @@ def fail_on_shallow(wd):
         )
 
 
-def parse(root, describe_command=DEFAULT_DESCRIBE, pre_parse=warn_on_shallow):
+def parse(root, config=None, describe_command=DEFAULT_DESCRIBE, pre_parse=warn_on_shallow):
```
RonnyPfannschmidt, line 86: due to legacy consideration for now we have to add config as a argument at the end
while i dont expect to break any code, its still a useful precaution for users calling without keyword args

src/setuptools_scm/git.py
```diff
@@ -82,20 +83,23 @@ def fail_on_shallow(wd):
...
     """
     :param pre_parse: experimental pre_parse action, may change at any time
     """
+    if not config:
```
RonnyPfannschmidt, line 90: i believe have an idea for a followup pr to get the positional arguments and the keyword arguments in a desirable shape for backward compat support, at that point i'll also introduce DeprecationWarnings for the old apis so it can be dropped in 1-2 years

src/setuptools_scm/git.py
```diff
@@ -82,20 +83,23 @@ def fail_on_shallow(wd):
...
 
-    wd = GitWorkdir.from_potential_worktree(root)
+    wd = GitWorkdir.from_potential_worktree(config.absolute_root)
     if wd is None:
         return
     if pre_parse:
         pre_parse(wd)
 
-    out, err, ret = wd.do_ex(describe_command)
+    out, unused_err, ret = wd.do_ex(describe_command)
```
RonnyPfannschmidt, line 102: this one is a good find, i believe an extra issue should go open to output those errors, thanks :+1:

src/setuptools_scm/hg.py
```diff
@@ -28,10 +29,13 @@ def _hg_tagdist_normalize_tagcommit(root, tag, dist, node, branch):
         return meta(tag)
 
 
-def parse(root):
+def parse(root, config=None):
```
RonnyPfannschmidt, line 32: as i seeing this pattern i believe we should have a common pattern for this 

```python
def func(root, ...):
   ...
```
should replace with

```
def func(config_or_root=None, ..., root=None):
   config = Configuration.maybe_from_legacy_args(root_or_config, root)
```
(its possible i missed some calls where its not always possible that way)

this helper could trigger deprecationwarnings

src/setuptools_scm/version.py
```diff
@@ -13,14 +14,34 @@
...
-TAG_PREFIX = re.compile(r"^\w+-(.*)")
 
 
 def _pad(iterable, size, padding=None):
     padded = chain(iterable, repeat(padding))
     return list(islice(padded, size))
 
 
+def _parse_version_tag(tag, config):
+    tagstring = tag if not isinstance(tag, string_types) else '%s' % tag
```
RonnyPfannschmidt: please use `str(tag)` instead of `%` formatting

src/setuptools_scm/version.py
```diff
@@ -56,33 +77,46 @@ def callable_or_entrypoint(group, callable_or_name):
...
+        warnings.warn("tag %r will be stripped of its suffix '%s'" % (tag, tagdict['suffix']))
+
+    if VERSION_CLASS is not None:
+        version = pkg_parse_version(version)
+        trace("version", repr(version))
+    
+    return version
 
-def tags_to_versions(tags):
-    versions = map(tag_to_version, tags)
```
RonnyPfannschmidt, line 84: i believe there is a neat trick to use here - 

```
from itertools import repeat
map(tag_to_version, tags, repeat(config, len(tags))
```
but im not sure that truly aids readability as much ^^

src/setuptools_scm/version.py
```diff
@@ -139,16 +173,18 @@ def format_next_version(self, guess_next, fmt="{guessed}.dev{distance}", **kw):
...
         return tag
     if VERSION_CLASS is None or not isinstance(tag, VERSION_CLASS):
-        tag = tag_to_version(tag)
+        tag = tag_to_version(tag, config)
     return tag
 
 
-def meta(tag, distance=None, dirty=False, node=None, preformatted=False, **kw):
-    parsed_version = _parse_tag(tag, preformatted)
+def meta(tag, config=None, distance=None, dirty=False, node=None, preformatted=False, **kw):
```
RonnyPfannschmidt: unfortunately this is a breaking api change, please move config to the end right after preformatted - this one really makes me want to start to use keyword only args from python3 sonner

testing/test_functions.py
```diff
@@ -34,11 +34,11 @@ def test_next_tag(tag, expected):
 
 
 VERSIONS = {
-    "exact": meta("1.1", None, False),
-    "zerodistance": meta("1.1", 0, False),
-    "dirty": meta("1.1", None, True),
-    "distance": meta("1.1", 3, False),
-    "distancedirty": meta("1.1", 3, True),
+    "exact": meta("1.1", distance=None, dirty=False),
```
RonnyPfannschmidt, line 37: this change here really makes things more beautiful,
i'd like to keep it even tho it was made necessary by the accidental api break before

RonnyPfannschmidt (approved):
awesome work, this is shaping up really nicely
i'm mostly seeing very minor details around api and nice to haves for followups :+1: 

please revisitt the argument ordering, i believe afterwards this one is ready to merge

RonnyPfannschmidt (changes requested):
whops, a missclick

src/setuptools_scm/git.py
punkadiddle: Will do. Didn't see this as an external API.

src/setuptools_scm/git.py
punkadiddle: was probably a flake8 warning in vscode

src/setuptools_scm/version.py
punkadiddle: ok

src/setuptools_scm/version.py
punkadiddle: done, not sure why I did it in the first place

src/setuptools_scm/hg.py
punkadiddle: Hmm, the basic idea was to keep existing `parse` functions as they were, because someone could have implemented a different provider relying on the existing `parse` signature. Requiring adding something to parse would be a breaking change for them.

Since the introspection clicks on the named argument, root could be left out entirely since it's redundant. Not having positional args could be checked. Then again you considered git.parse as API to keep stable above.
```python
def func(config=, **kwargs)
```

src/setuptools_scm/version.py
punkadiddle: ;-) needs to be filtered for None also.

src/setuptools_scm/hg.py
RonnyPfannschmidt: i see, i'm deferring handling this for the future then

src/setuptools_scm/config.py
RonnyPfannschmidt: right now i'm not happy with this config class, but its working and adds the feature needed so i'm happy to merge, its up to me or someone else to make me more happy with it afterwards

RonnyPfannschmidt (approved):
looks good to merge now to me,

unless something comes up i'll merge this tomorrow morning and wire up a release

thanks for pulling trough and wiring the config class all trough the codebase, its a tedious process

punkadiddle:
Thank you, was nice to work with you.

RonnyPfannschmidt:
sorry for the delay

PR #277, opened 2018-07-01
Do not parse tag when pretending version

### LINKED ISSUE #271: PRETEND behaviour
Hi,

I am not sure I understand the `SETUPTOOLS_SCM_PRETEND_VERSION` usage.
I use it in a container context where I don't wan't the scm repository to be included in the container image.
Thus I use something like:
```shell
SETUPTOOLS_SCM_PRETEND_VERSION=$(python ${PWD}/setup.py --version) 
```
prior to create the container image so that `setuptools_scm` can retrieve the version without the scm repository. 
However, I don't get the same result as with the scm repository.

With a version like `1.0.1.dev1+ge871260.d20180625` and the scm repository, the generated artifact (here a wheel) will be `my_python_project-1.0.1.dev1+ge871260.d20180625-py3-none-any.whl`, while without the scm repository and with `SETUPTOOLS_SCM_PRETEND_VERSION` set to `1.0.1.dev1+ge871260.d20180625` the generated artifact is `my_python_project-1.0.1.dev1-py3-none-any.whl`.

Looking at the `_do_parse` function code:
```python
if pretended:
        # we use meta here since the pretended version
        # must adhere to the pep to begin with
        return meta(pretended)
```
I understand this is done on purpose, but I don't get why.
I would expect the result to be the same, unless I don't get the `SETUPTOOLS_SCM_PRETEND_VERSION` usage properly.

jpiron:
Fixes #271.

From what I understand of the pretend behavior I am not sure it should go through the parsing process.

### DIFF
--- a/src/setuptools_scm/__init__.py
+++ b/src/setuptools_scm/__init__.py
@@ -60,7 +60,7 @@ def _do_parse(root, parse):
     if pretended:
         # we use meta here since the pretended version
         # must adhere to the pep to begin with
-        return meta(pretended)
+        return meta(tag=pretended, preformatted=True)
 
     if parse:
         parse_result = parse(root)
--- a/testing/test_basic_api.py
+++ b/testing/test_basic_api.py
@@ -51,10 +51,12 @@ def test_root_parameter_pass_by(monkeypatch, tmpdir):
     setuptools_scm.get_version(root=tmpdir.strpath)
 
 
-def test_pretended(monkeypatch):
-    pretense = "2345"
-    monkeypatch.setenv(setuptools_scm.PRETEND_KEY, pretense)
-    assert setuptools_scm.get_version() == pretense
+@pytest.mark.parametrize(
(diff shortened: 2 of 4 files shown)


jpiron:
If the fix is ok for you I will add some tests for it

src/setuptools_scm/version.py
```diff
@@ -147,8 +147,12 @@ def _parse_tag(tag, preformatted):
     return tag
 
 
-def meta(tag, distance=None, dirty=False, node=None, preformatted=False, **kw):
-    parsed_version = _parse_tag(tag, preformatted)
+def meta(tag, distance=None, dirty=False, node=None, preformatted=False,
```
RonnyPfannschmidt: actually, its enough to pass preformatted, which does the same in parse_tag, jsut see above

man i really should have added that flag back then

jpiron: Indeed, much simpler this way !

RonnyPfannschmidt:
@jpiron can you perhaps add a test as well, this blunder wouldn't have happened if that code path was tested as intended

RonnyPfannschmidt (approved):

PR #314, opened 2018-10-08
Fix invalid escape sequences.

anntzer:
All's in the title.


RonnyPfannschmidt:
thanks, for contributing , i had to do the same change while i fixed issues with black usage

PR #316, opened 2018-10-27
Fix to #303

### LINKED ISSUE #283: support configuration of the git describe command
https://github.com/apache/arrow/blob/355ff0879d5fc7bd5c95aa05fd0e9daeafbaccd3/python/setup.py#L489-L500

there should be a better way to configure that

### LINKED ISSUE #303: non-matching tags are not ignored
If I give a `tag_regex`, I would expect non-matching tags to be ignored, instead they seem to fail:
```
$ python setup.py sdist
/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'long_description_content_type'
  warnings.warn(msg)
/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/site-packages/setuptools_scm/version.py:92: UserWarning: tag 'apache-arrow-js-0.9.9' no version found
  warnings.warn("tag %r no version found" % (tag,))
  File "setup.py", line 557, in <module>
    url="https://arrow.apache.org/"
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/site-packages/setuptools/__init__.py", line 129, in setup
    return distutils.core.setup(**attrs)
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/distutils/core.py", line 108, in setup
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/site-packages/setuptools/dist.py", line 333, in __init__
    _Distribution.__init__(self, attrs)
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/distutils/dist.py", line 281, in __init__
    self.finalize_options()
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/site-packages/setuptools/dist.py", line 476, in finalize_options
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/site-packages/setuptools_scm/integration.py", line 23, in version_keyword
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/site-packages/setuptools_scm/__init__.py", line 135, in get_version
    parsed_version = _do_parse(config)
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/site-packages/setuptools_scm/__init__.py", line 88, in _do_parse
    config, "setuptools_scm.parse_scm"
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/site-packages/setuptools_scm/__init__.py", line 45, in _version_from_entrypoint
    version = _call_entrypoint_fn(config, ep.load())
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/site-packages/setuptools_scm/__init__.py", line 33, in _call_entrypoint_fn
    return fn(config.absolute_root, config=config)
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/site-packages/setuptools_scm/git.py", line 134, in parse
    return meta(tag, config=config, node=node, dirty=dirty, branch=branch)
  File "/home/antoine/miniconda3/envs/pyarrow/lib/python3.6/site-packages/setuptools_scm/version.py", line 196, in meta
    assert parsed_version is not None, "cant parse version %s" % tag
AssertionError: cant parse version apache-arrow-js-0.9.9
```

In this case my options are as follows:
```
    use_scm_version={"root": "..",
                     "relative_to": __file__,
                     "tag_regex": '^apache-arrow-([\.0-9]+)$'},
```

1ucian0:
As part of the NYC PyPA sprint, I'm trying to fix #303. So far, the test triggers the original submitter error.

fixes #303 
fixes #283

### DIFF
--- a/src/setuptools_scm/__init__.py
+++ b/src/setuptools_scm/__init__.py
@@ -121,6 +121,7 @@ def get_version(
     relative_to=None,
     tag_regex=None,
     parse=None,
+    git_describe_command=None,
 ):
     """
     If supplied, relative_to should be a file from which root may
@@ -138,6 +139,7 @@ def get_version(
     config.relative_to = relative_to
     config.tag_regex = tag_regex
     config.parse = parse
+    config.git_describe_command = git_describe_command
 
     parsed_version = _do_parse(config)
 
--- a/src/setuptools_scm/config.py
+++ b/src/setuptools_scm/config.py
@@ -61,6 +61,7 @@ class Configuration(object):
         self.write_to_template = None
         self.parse = None
         self.tag_regex = DEFAULT_TAG_REGEX
+        self.git_describe_command = None
 
     @property
     def absolute_root(self):
--- a/src/setuptools_scm/git.py
+++ b/src/setuptools_scm/git.py
@@ -101,9 +101,12 @@ def parse(
     if pre_parse:
         pre_parse(wd)
 
+    if config.git_describe_command:
+        describe_command = config.git_describe_command
+
     out, unused_err, ret = wd.do_ex(describe_command)
     if ret:
-        # If 'git describe' failed, try to get the information otherwise.
+        # If 'git git_describe_command' failed, try to get the information otherwise.
         rev_node = wd.node()
         dirty = wd.is_dirty()
 
--- a/testing/test_git.py
+++ b/testing/test_git.py
@@ -191,3 +191,18 @@ def test_git_feature_branch_increments_major(wd):
     wd("git checkout -b feature/fun")
     wd.commit_testfile()
     assert wd.get_version(version_scheme="python-simplified-semver").startswith("1.1.0")
+
+
+@pytest.mark.issue("https://github.com/pypa/setuptools_scm/issues/303")
+def test_not_matching_tags(wd):
+    wd.commit_testfile()
+    wd("git tag apache-arrow-0.11.1")
+    wd.commit_testfile()
+    wd("git tag apache-arrow-js-0.9.9")
+    wd.commit_testfile()
+    assert wd.get_version(
+        tag_regex=r"^apache-arrow-([\.0-9]+)$",
+        git_describe_command="git describe --dirty --tags --long --exclude *js* ",
+    ).startswith(
+        "0.11.2"
+    )
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,8 @@ v3.1.0
 * fix #297 - correct the invocation in version_from_scm and deprecate it as its exposed by accident
 * fix #298 - handle git file listing on empty repositories
 * fix #268 - deprecate ScmVersion.extra
+* fix #303 and #283 by adding the option `git_describe_command` to allow the user to control the
+way that `git describe` is called.
 
 v3.0.6
 ======
--- a/README.rst
+++ b/README.rst
@@ -210,6 +210,13 @@ The currently supported configuration keys are:
   defaults to the value of ``setuptools_scm.config.DEFAULT_TAG_REGEX``
   (see `config.py <src/setuptools_scm/config.py>`_).
 
+:git_describe_command:
+  This command will be used instead the default `git describe` command.
+  Use with caution, this is a function for advanced use, and you should be
+  familiar with the setuptools_scm internals to use it.
+
+  The default value is set by ``setuptools_scm.git.DEFAULT_DESCRIBE``
+  (see `git.py <src/setuptools_scm/git.py>`_).
 
 To use setuptools_scm in other Python code you can use the
 ``get_version`` function:


1ucian0:
This is my attempt to fix #283, by adding to configuration elements `match` and `exclude`. The value set in these elements is used for the `git describe` command:
```
git describe --dirty --tags --long --match <match_value> --exclude <exclude_value>
```
In this way, issue #303 is fixed by adding `exclude="*js*"` into the configuration.

src/setuptools_scm/git.py
```diff
@@ -12,7 +12,7 @@
     from .win_py31_compat import samefile
 
 
-DEFAULT_DESCRIBE = "git describe --dirty --tags --long --match *.*"
+DEFAULT_DESCRIBE = "git describe --dirty --tags --long"
```
RonnyPfannschmidt: thats a breaking change for outisde users

src/setuptools_scm/config.py
```diff
@@ -8,7 +8,7 @@
 
 DEFAULT_TAG_REGEX = r"^(?:[\w-]+-)?(?P<version>[vV]?\d+(?:\.\d+){0,2}[^\+]+)(?:\+.*)?$"
 DEFAULT_VERSION_SCHEME = "version_scheme"
-
+DEFAULT_MATCH = '*.*'
```
RonnyPfannschmidt: unless there is a scm-portable version of this, we should use a single declaration for the complete git describe command

RonnyPfannschmidt (changes requested):

jaraco:
> In this way, issue #303 is fixed by adding exclude="*js*" into the configuration.

Although you may be right that the specific example presented in #303 might be addressed, I don't think it addresses the issue in general. I don't think the failed expectation is that `js` is included, but that one can't select exactly the format of tags needed. In other words, I think the original reporter (or similar users) also want to exclude `apache-arrow-node-N.N.N` and `foo-bar-N.N.N`. In other words _only_ packages that match `apache-arrow-N.N.N` should match... and other tags in the repo shouldn't crash. I don't think the recommendation to add an 'exclude' for every tag that might not match is a good recommendation in general.

testing/test_git.py
```diff
@@ -191,3 +191,31 @@ def test_git_feature_branch_increments_major(wd):
...
+
+@pytest.mark.issue("https://github.com/pypa/setuptools_scm/issues/303")
+def test_match(wd):
+    wd.commit_testfile()
+    wd("git tag apache-arrow_0.11.1")
+    wd.commit_testfile()
+    wd("git tag apache-arrow-js_0.9.9")
+    wd.commit_testfile()
+    assert wd.get_version(
+        tag_regex=r"^apache-arrow_([\.0-9]+)$", match="apache-arrow_*"
```
jaraco: This feels redundant to me - I think the project should try hard not to require the user to specify both a tag_regex and a match just to avoid a bug in tag_regex.

jaraco:
In Gitter, Ronny also suggested that the solution should ideally apply to other SCMs (such as Mercurial, which is also supported by this library). I agree with that too. Something like `tag_regex` should apply to tags from any SCM.

testing/test_git.py
```diff
@@ -191,3 +191,31 @@ def test_git_feature_branch_increments_major(wd):
...
+def test_exclude(wd):
+    wd.commit_testfile()
+    wd("git tag apache-arrow-0.11.1")
+    wd.commit_testfile()
+    wd("git tag apache-arrow-js-0.9.9")
+    wd.commit_testfile()
+    assert wd.get_version(
+        tag_regex=r"^apache-arrow-([\.0-9]+)$", exclude="*js*"
+    ).startswith(
+        "0.11.2"
```
jaraco: How does this pass if 0.11.1 is the tag added above?

src/setuptools_scm/__init__.py
```diff
@@ -138,6 +139,7 @@ def get_version(
     config.relative_to = relative_to
     config.tag_regex = tag_regex
     config.parse = parse
+    config.describe = describe
```
RonnyPfannschmidt: this should be called `git_describe_command?

jaraco:
In [this branch](https://github.com/pypa/setuptools_scm/tree/bugfix/303-tag_regex), I added a simple test, based on the work in this ticket, capturing what I think is the essential bug that needs to be addressed. But this issue is really tricky. I'll try to describe my renewed understanding more in #303.

src/setuptools_scm/__init__.py
1ucian0: makes sense. Fixed in ecbc46f

RonnyPfannschmidt (approved):
well done, please think about squahsing and apply it before removing the wip state

once the wip is gone i'll merge

:+1:

Issue #315, opened 2018-10-09
is setup_requires method deprecated?


anthrotype:
Following up from https://github.com/pypa/pip/issues/5465#issuecomment-395045552

If easy_install can no longer reliably install setuptools_scm via `setup_requires`, at least of some platforms or versions of pip with old SSL, shouldn't users be officially recommended to use pyproject.toml and specify setuptools_scm in `[build-system] requires`?

The current README only mention the `setup_requires` installation method.

E.g. I noticed that tox recently removed `setuptools_scm` from its setup.py `setup_requires` list:
https://github.com/tox-dev/tox/commit/da219ee40783c0385c784fd358f6ec5f7486d784#diff-2eeaed663bd0d25b7e608891384b7298

RonnyPfannschmidt:
@anthrotype thanks for the note - while the method is not yet deprecated, its indeed advisable to have people switch to the mechanisms outlined

tox already used pep-517 before, and just had to keep the list for a broken tool

would you like to contribute the documentation change, personally im not sure when i can get to it

RonnyPfannschmidt:
Closed by #323

soerenwolfers:
The problems with setup_requires disappeared with setuptools 42.0.0, which switched to using pip to resolve setup_requires (https://setuptools.readthedocs.io/en/latest/history.html#v42-0-0)

PR #323, opened 2018-12-12
Add documentation for 'setup.cfg' configuration

stephenfin:
An issue with `[egg_info]` was seen with https://github.com/stephenfin/rst2txt. Take the opportunity to highlight this issue along with general `setup.cfg` support.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -110,7 +110,7 @@ version of ``setuptools``:
 
     # pyproject.toml
     [build-system]
-    requires = ["setuptools>=30.3.0", "wheel"]
+    requires = ["setuptools>=30.3.0", "wheel", "setuptools_scm"]
 
 For more information, refer to the `setuptools issue #1002
 <https://github.com/pypa/setuptools/issues/1002>`_.
--- a/README.rst
+++ b/README.rst
@@ -97,8 +97,23 @@ However, ``use_scm_version`` must still be placed in ``setup.py``. For example:
       setuptools_scm
     ...
 
-If using this, you must neither the ``[metadata] version`` option nor the
-``[egg_info]`` section are not defined.
+.. important::
+
+    Ensure neither the ``[metadata] version`` option nor the ``[egg_info]``
+    section are not defined, as these will interfere with ``setuptools_scm``.
+
+You may also need to define a ``pyproject.toml`` file (`PEP-0518
+<https://www.python.org/dev/peps/pep-0518>`_) to ensure you have the required
+version of ``setuptools``:
+
+.. code:: ini
+
+    # pyproject.toml
+    [build-system]
+    requires = ["setuptools>=30.3.0", "wheel"]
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt (changes requested):
great job, the markup enhancements are neat and the added informations realy drive home a more metadata oriented approach

for it to be fully embracing of the setuptools features we also need to show people the patterns needed for minimal setuptools versions, else they are in for bad surprises on enterprise distro style setups

stephenfin:
> for it to be fully embracing of the setuptools features we also need to show people the patterns needed for minimal setuptools versions, else they are in for bad surprises on enterprise distro style setups

Good call. Done.

anthrotype:
if you add setuptools_scm to the build requirements in pyproject.toml, then you could do away with setup_requires and eggs altogether, no?

anthrotype:
maybe for the time being would be better to recommend to add setuptools_scm to both pyproject.toml and the setup_requires, so that newer pips will use the former (and skip the latter as already resolved by the time setup_requires are evaluated), and older pips that don't know about pyproject.toml can keep using setup_requires (provided they don't fail with some obscure SSL error).

stephenfin:
> maybe for the time being would be better to recommend to add setuptools_scm to both pyproject.toml and the setup_requires, so that newer pips will use the former (and skip the latter as already resolved by the time setup_requires are evaluated), and older pips that don't know about pyproject.toml can keep using setup_requires (provided they don't fail with some obscure SSL error).

Done

RonnyPfannschmidt (approved):
great work, i'm going to merge later this evening, please give me a note if you wish to streamline history a bit

stephenfin:
Yup, happy for you to squash the last three commits. I just kept them separate because I've had people complain about force pushing on GitHub.

RonnyPfannschmidt:
@stephenfin my personal stance is

* i keep commits separate while the pr is being developed
* i will squash my own commits after the review process is done and i never squash commits of others
* i give contributors the opportunity to squash themselves before i merge

stephenfin:
Oh, that all makes sense. In this particular case though, I'm away from a computer and won't be near one again until next week. As such, if you're happy to squash them, go for it.

README.rst
```diff
@@ -1,61 +1,119 @@
...
+
+    [options]
+    setup_requires =
+      setuptools_scm
+    ...
+
+.. important::
+
+    Ensure neither the ``[metadata] version`` option nor the ``[egg_info]``
+    section are not defined, as these will interfere with ``setuptools_scm``.
```
marcelm, line 103: The extra “not” in this line should probably be removed

marcelm:
This PR would close #315

RonnyPfannschmidt:
good reminder, it got lost over the holiday season, i'll put it in my week todo

PR #308, opened 2018-09-05
Allow setting a fallback version in case everything else fails.

anntzer:
See https://github.com/pypa/setuptools_scm/issues/198#issuecomment-418666112.  Needs tests.
Feel free to adopt the PR.

Issue #198, opened 2017-12-18
setuptools_scm explicit silent fallback


gaborbernat:
Now  if in ``setup.py``  we turn on setuptools_scm and there's no SCM repository an exception is raised that no SCM repository, is it possible to fallback to a value instead of raising this error and still build the package? e.g. ``0.0.0-dev`` ?

Kentzo:
Take a look at https://github.com/GreatFruitOmsk/setuptools_scm_about

Best Regards
Ilya Kulakov

RonnyPfannschmidt:
@gaborbernat the env var `SETUPTOOLS_SCM_PRETEND_VERSION` is the canonical way to make it use a given value instead of trying to find a scm

gaborbernat:
I see, however that overwrites the look for SCM. I wan't fallback to this value if no SCM. The about is what I want.

RonnyPfannschmidt:
@gaborbernat in what cases does this happen to be required, i'd like to know more about the use-case before blindly enabling something

gaborbernat:
the use case is the following:
- let there be a plugin-able library that uses the mechanism at https://packaging.python.org/guides/creating-and-discovering-plugins/
- now you want to write  a plugin for it,  you create a virtualenv, a ``setup.py`` for the plugin and to test it/develop it you use ``pip install -e .``
- in order to mock something together locally (maybe you just want to try something out) with static version in ``setup.py`` no git repository is needed; with ``setuptools_scm`` current version before ``pip install -e .`` you're forced the init an SCM repository (otherwise you're install will fail with an error) - at this point ``setuptools_scm`` forces you a step which you may forget, blowing up with an error. If you determine the plugin viable you'll create the git repository later on, at which point comes handy to have ``setuptools_scm`` manage versioning. The reason the plugins ``setup.py`` contains the use setuptools flag, is because it's generated by a scaffold generator (e.g. cookiecutter).

RonnyPfannschmidt:
hmm, i'd like to point out that a cookiecutter plugin that makes something that requires scm but doesn't ensure/init the scm repo as well is broken and should be fixed instead - in particular if the init step is as cheap as a git init / reminidng the user to do a git init

in addition you can just comment/mask the use_scm_version line and it just stops caring
[...]

PR #327, opened 2019-02-19
Allow setting a fallback version in case everything else fails.

anntzer:
Revival of #308, but with tests now :)
Initially discussed at https://github.com/pypa/setuptools_scm/issues/198#issuecomment-418666112.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,8 @@
+next
+====
+
+* fix #198 by adding the ``fallback_version`` option, which sets the version to be used when everything else fails.
+
 v3.2.0
 ======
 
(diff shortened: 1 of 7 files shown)


RonnyPfannschmidt (changes requested):
looks good :+1: thank for revising and completing it

 please add it to the change-log as well

anntzer:
should that go under a `next` heading in the changelog?

edit: done

anntzer:
good to go?

RonnyPfannschmidt (approved):

RonnyPfannschmidt:
thanks for the ping, nicely done, for personal reasons i might not make it to a release myself until next weekend

anntzer:
No hurries, thanks for taking care of this.

PR #334, opened 2019-05-10
Change fallbacks to use a static root

### LINKED ISSUE #333: Relative `__file__` in `relative_to` error
Following the example in the documentation, I have a `relative_to` that contains `__file__`. In CI, recently and for reasons still unknown, this started failing. Inspecting the failure, I noticed that `__file__` was relative (`setup.py`) and that providing an absolute path fixed it.

I am not sure what to do about this other than to raise it for investigation. Perhaps something changed in setuptools that caused this to become relative now, or something changed in setuptools_scm, but I worry that others could trip over this.

ionelmc:
Should fix #333.

### DIFF
--- a/src/setuptools_scm/__init__.py
+++ b/src/setuptools_scm/__init__.py
@@ -34,9 +34,9 @@ def version_from_scm(root):
     return _version_from_entrypoint(config, "setuptools_scm.parse_scm")
 
 
-def _call_entrypoint_fn(config, fn):
+def _call_entrypoint_fn(root, config, fn):
     if function_has_arg(fn, "config"):
-        return fn(config.absolute_root, config=config)
+        return fn(root, config=config)
     else:
         warnings.warn(
             "parse functions are required to provide a named argument"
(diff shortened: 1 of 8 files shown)


src/setuptools_scm/__init__.py
```diff
@@ -34,22 +34,27 @@ def version_from_scm(root):
...
             stacklevel=2,
         )
-        return fn(config.absolute_root)
+        return fn(root)
 
 
-def _version_from_entrypoint(config, entrypoint):
-    for ep in iter_matching_entrypoints(config.absolute_root, entrypoint):
-        version = _call_entrypoint_fn(config, ep.load())
+def _version_from_entrypoint(config, entrypoint, fallback=False):
```
RonnyPfannschmidt: please infer the entry-point from the Boolean, so the entry-point name is no longer passed to this function
if you get a good idea for the new function name from that, please apply it as well

RonnyPfannschmidt (changes requested):
great job at following trough the code and finding a working solution,
i do believe that this sorts the issue out much better

ionelmc:
So the failures are mostly incorrect tests. But that reveals a problem with the current API, there's no way to say "this is the root for fallbacks". Perhaps add new fallback_root/fallback_relative_to settings?

src/setuptools_scm/__init__.py
RonnyPfannschmidt: https://github.com/pypa/setuptools_scm/blob/master/testing/conftest.py#L66 - this needs to pass fallback_root as well

src/setuptools_scm/config.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, relative_to=None, root="."):
         self.write_to = ""
         self.write_to_template = None
         self.fallback_version = None
+        self.fallback_root = _check_absolute_root(".", None)
```
RonnyPfannschmidt: i believe fallback_root can be a parameter - and for the testing it needs to be

src/setuptools_scm/__init__.py
ionelmc: Changed to _version_from_entrypoint**s**

RonnyPfannschmidt:
if you sneak in the changelog for 3.3.1 as  well i can cut a release right after merge using a tag

RonnyPfannschmidt (approved):

PR #338, opened 2019-05-14
#235: get_latest_normalizable_tag: return most recent tag that matches config.tag_regex

reece:
This PR attempts to fix issues caused by tags that contain periods and are assumed to be version numbers, but fail when used as version numbers elsewhere in setuptools_scm.  See an example in #235.

Please confirm that the return value should be `"null"` (i.e., a string) rather than None, which would be more conventional for Python.

Note that the argument for `get_latest_normalizable_tag` is now `config`, not `root`. This is based on an assumption that the function is not used by external packages.

If this function is considered public (i.e., used outside the package), the signature could be made backward compatible as `get_latest_normalizable_tag(root, tag_regex=None)`, where `tag_regex` can default to `config.DEFAULT_TAG_REGEX`.

### DIFF
--- a/src/setuptools_scm/hg.py
+++ b/src/setuptools_scm/hg.py
@@ -1,4 +1,6 @@
 import os
+import re
+
 from .config import Configuration
 from .utils import do, trace, data_from_mime, has_command
 from .version import meta, tags_to_versions
@@ -58,7 +60,7 @@ def parse(root, config=None):
         return meta("0.0", config=config, dirty=dirty, branch=branch)
 
     try:
-        tag = get_latest_normalizable_tag(config.absolute_root)
+        tag = get_latest_normalizable_tag(config)
         dist = get_graph_distance(config.absolute_root, tag)
         if tag == "null":
             tag = "0.0"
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
mercurial uses null as signaling value, so it is correct in that specific context

src/setuptools_scm/hg.py
```diff
@@ -68,15 +70,23 @@ def parse(root, config=None):
...
     cmd = [
         "hg", "log", "-r", "ancestors(.) and tag('re:\\.')", "--template", "{tags}\n"
     ]
     outlines = do(cmd, root).split()
     if not outlines:
         return "null"
-    tag = outlines[-1].split()[-1]
+    tag_regex = re.compile(config.tag_regex)
+    try:
+        tag = next(v for v in reversed(outlines) if tag_regex.match(v))
```
RonnyPfannschmidt, line 87: please use a list comp before the `if not outlines` line to filter using the regex instead

RonnyPfannschmidt:
please also add a unit/integration test

reece:
Will do, but it'll take me a few weeks to come back to this.

RonnyPfannschmidt:
thanks for providing the etas, im in a similar boat

reece:
I'm not going to come back to this PR. I don't use hg + hg-git any longer, so the events that trigger this bug no longer occur. (Sadly... I still much prefer hg over git.)

If someone else would like to pick up the PR, help yourself. 

Alternatively, @RonnyPfannschmidt, you can close without merging if you like.

RonnyPfannschmidt:
personally i also prefer hg, the actions of atlassian however are kind of killing the mind-share and the public utility of mercurial - its an utter pain and the main reason why im on git since a while (git is still completely horrible and a mess of mangled concepts)

src/setuptools_scm/hg.py
```diff
@@ -68,15 +70,23 @@ def parse(root, config=None):
         pass  # unpacking failed, old hg
 
 
-def get_latest_normalizable_tag(root):
-    # Gets all tags containing a '.' (see #229) from oldest to newest
+def get_latest_normalizable_tag(config):
```
webknjaz, line 73: I don't think that changing the signature of a public function is a good idea.

RonnyPfannschmidt: good catch

RonnyPfannschmidt:
i'm closing this one as victim of the times
i wont have time to pick it off, i hope a mercurial user has
i'm happy to review if anyone picks it up

PR #345, opened 2019-07-16
Fix pip wheel

RonnyPfannschmidt:
this pr adds acceptance tests for pip wheel and pip install usages where its possible to be copied to a tempdir


RonnyPfannschmidt:
closing for now as pip evolved in the meantime

PR #351, opened 2019-07-31
Document file_finders in more detail

vanschelven:
* The `sdist` command is affected by its existance
* Correction: MANIFEST.in can still be used (in principle) to include files too.


RonnyPfannschmidt:
looks great, thanks

PR #356, opened 2019-09-05
get branch If default 'git git_describe_command' failed

hbasria:
➜  mkdir demo
➜  cd demo
➜  git init
Initialized empty Git repository in demo/.git/
➜  git describe --dirty --tags --long --match *.* --first-parent
no matches found: *.*
➜  demo git symbolic-ref --short HEAD
master

### DIFF
--- a/src/setuptools_scm/git.py
+++ b/src/setuptools_scm/git.py
@@ -107,11 +107,16 @@ def parse(
     out, unused_err, ret = wd.do_ex(describe_command)
     if ret:
         # If 'git git_describe_command' failed, try to get the information otherwise.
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt (changes requested):
Nice addition, please extend the git tests to consider this case as well

RonnyPfannschmidt:
gentle ping, i'l aim for a release this weekend, unfortunately i cant pick up the test writing for this one until then, so if you manage to add the tests i can include it, else i will try to pick things up afterward

RonnyPfannschmidt:
turns out adding a test was trivial, thanks again

PR #364, opened 2019-10-19
Implement declarative config for compatibility with setuptools declarative config

### LINKED ISSUE #181: setuptools_scm does not work with setup.cfg
I was debuging  the behaviour of `setuptools_scm` because I received exception from function `version_keyword`
``` 
File "/home/oglop/.virtualenvs/b/lib/python3.6/site-packages/setuptools_scm/integration.py", line 22, in version_keyword
```

in `setuptools` file `config.py`, all options go through parser except additional options like `use_scm_version`, I guess this should be handled by `setuptools_scm`
*file* : `setuptools/config.py`
```
class ConfigOptionsHandler(ConfigHandler):

    section_prefix = 'options'

    @property
    def parsers(self):
        """Metadata item name to parser function mapping."""
        parse_list = self._parse_list
        parse_list_semicolon = partial(self._parse_list, separator=';')
        parse_bool = self._parse_bool
        parse_dict = self._parse_dict

        return {
            'zip_safe': parse_bool,
            'use_2to3': parse_bool,
            'include_package_data': parse_bool,
            'package_dir': parse_dict,
            'use_2to3_fixers': parse_list,
            'use_2to3_exclude_fixers': parse_list,
            'convert_2to3_doctests': parse_list,
            'scripts': parse_list,
            'eager_resources': parse_list,
            'dependency_links': parse_list,
            'namespace_packages': parse_list,
            'install_requires': parse_list_semicolon,
            'setup_requires': parse_list_semicolon,
            'tests_require': parse_list_semicolon,
            'packages': self._parse_packages,
            'entry_points': self._parse_file,
            'py_modules': parse_list,
        }
```

in `setuptools_scm`
 something goes wrong here:
```
[distutils.setup_keywords]
        use_scm_version = setuptools_scm.integration:version_keyword
```
because function `version_keyword` receives parameter `value` as `str` but that function expects `bool` - because the value did not go through the *parsing process*

what I do not know is how that `version_keyword` is called

jaraco:
Fixes #181 

In irc.freenode.net#pylib, Ronny asked about supporting setuptools_scm on setuptools in a declarative config world. This change illustrates how I'd propose for that to happen. Setuptools would expose a new hook, something like

https://gist.github.com/b45998781be03918a21f17da996ee687

And then setuptools_scm would provide a hook into that such as the one (building on all of the existing work already done).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,8 +13,62 @@ It also handles file finders for the supported SCMs.
 .. image:: https://tidelift.com/badges/github/pypa/setuptools_scm
    :target: https://tidelift.com/subscription/pkg/pypi-setuptools_scm?utm_source=pypi-setuptools_scm&utm_medium=readme
 
+``pyproject.toml`` usage
+------------------------
+
+The preferred way to configure ``setuptools_scm`` is to author
+settings in a ``tool.setuptools_scm`` section of ``pyproject.toml``.
+
+This feature requires Setuptools 42 or later, released in Nov, 2019.
+If your project needs to support build from sdist on older versions
+of Setuptools, you will need to also implement the ``setup.py usage``
+for those legacy environments.
+
+First, ensure that ``setuptools_scm`` is present during the project's
+built step by specifying it as one of the build requirements.
+
+.. code:: ini
+
+    # pyproject.toml
+    [build-system]
+    requires = ["setuptools>=42", "wheel", "setuptools_scm[toml]>=3.4"]
+
+Note that the ``toml`` extra must be supplied.
+
+That will be sufficient to require ``setuptools_scm`` for projects
+that support PEP 518 (`pip <https://pypi.org/project/pip>`_ and
+`pep517 <https://pypi.org/project/pep517/>`_). Many tools,
+especially those that invoke ``setup.py`` for any reason, may
+continue to rely on ``setup_requires``. For maximum compatibility
+with those uses, consider also including a ``setup_requires`` directive
+(described below in ``setup.py usage`` and ``setup.cfg``).
+
+To enable version inference, add this section to your pyproject.toml:
+
+.. code:: ini
+
+    # pyproject.toml
+    [tools.setuptools_scm]
+
+Including this section is comparable to supplying
+``use_scm_version=True`` in ``setup.py``. Additionally,
+include arbitrary keyword arguments in that section
+to be supplied to ``get_version()``. For example::
+
+.. code:: ini
+
+    # pyproject.toml
+    [tools.setuptools_scm]
+    write_to = pkg/version.py
+
+
 ``setup.py`` usage
 ------------------
+
+The following settings are considered legacy behavior and
+superseded by the ``pyproject.toml`` usage, but for maximal
+compatibility, projects may also supply the configuration in
+this older form.
 
 To use ``setuptools_scm`` just modify your project's ``setup.py`` file
 like this:
--- a/setup.py
+++ b/setup.py
@@ -113,8 +113,8 @@ arguments = dict(
         "Topic :: System :: Software Distribution",
         "Topic :: Utilities",
     ],
-    install_requires=["toml"],
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
+    extras_require=dict(toml=["toml"]),
 )
 
 if __name__ == "__main__":
--- a/setup.py
+++ b/setup.py
@@ -113,6 +113,7 @@ arguments = dict(
         "Topic :: System :: Software Distribution",
         "Topic :: Utilities",
     ],
+    install_requires=["toml"],
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
 )
 
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,9 @@ arguments = dict(
         [setuptools.file_finders]
         setuptools_scm = setuptools_scm.integration:find_files
 
+        [setuptools.finalize_distribution_options]
+        setuptools_scm = setuptools_scm.integration:infer_version
+
         [setuptools_scm.parse_scm]
         .hg = setuptools_scm.hg:parse
         .git = setuptools_scm.git:parse
--- a/src/setuptools_scm/config.py
+++ b/src/setuptools_scm/config.py
@@ -114,5 +114,5 @@ class Configuration(object):
         with open(name) as strm:
             defn = __import__('toml').load(strm)
         config = cls()
-        vars(config).update(defn.get('setuptools_scm', {}))
+        vars(config).update(defn.get("tool",{}).get('setuptools_scm', {}))
         return config
--- a/src/setuptools_scm/config.py
+++ b/src/setuptools_scm/config.py
@@ -107,12 +107,12 @@ class Configuration(object):
         self._tag_regex = _check_tag_regex(value)
 
     @classmethod
-    def from_file(cls, name='pyproject.toml'):
+    def from_file(cls, name="pyproject.toml"):
         """
         Read Configuration from pyproject.toml (or similar)
         """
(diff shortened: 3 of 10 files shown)


jaraco:
Still to do:

- [x] Settle on the right hook/API and release changes in setuptools.
- [x] Decide if setuptools_scm needs to support both `setup()` invocation and this declarative mode.
- [x] If so, implement compatibility protocol.
- [x] Add tests for this new functionality.
- [x] Implement the stubbed config file reader.
- [x] Update docs to describe how to use this preferred functionality.

RonnyPfannschmidt (commented):
I like this,
the setup keyword can actually stay relatively easy with this,

Could setuptools provide a helper to get the settings section from pyproject.toml

Thanks!

jaraco:
Unfortunately, setuptools itself doesn't yet have toml (or pyproject.toml) support... but it might be worthwhile for setuptools to own that functionality to avoid plugins needing to own the parsing.

How much harm is it for setuptools_scm to have a dependency (toml)?

RonnyPfannschmidt:
its a pain for anyone using the legacy support as it will pull in toml via easy_install

so having a toml dependency outside of what setuptools does would be a big blow for backward compat

its possible to make it a extra dependency 
but its not clear to me how to make the entrypoint work properly then in all situation

RonnyPfannschmidt:
@jaraco got thinking again, its easy to have a extra and a optional import, after all - i can sort out most of the pain i believe (bit it would still be much nicer if the toml data was readily available in some way)

jaraco:
> its a pain for anyone using the legacy support as it will pull in toml via easy_install

Are there scenarios where requiring a pure-python source distribution wouldn't work as a simple dependency (as [toml installs under easy_install](https://gist.github.com/ed775dc0859eedb5c955b8ebf09079f4))? I'm not opposed to finding another solution, but I'd like to know what scenario we're seeking to solve.

> its easy to have a extra and a optional import

Oh, so `setuptools_scm[toml]>=version_with_toml` (or similar) would be the requirement specified by `setup_requires` or PEP 518 `build_requires`? That sounds feasible.

testing/test_config.py
```diff
@@ -18,3 +18,10 @@ def test_tag_regex(tag, expected_version):
     match = config.tag_regex.match(tag)
     version = match.group("version")
     assert version == expected_version
+
+
+def test_config_from_pyproject(tmpdir):
+    fn = tmpdir / 'pyproject.toml'
+    fn.write_text('[setuptools_scm]\nenabled = true\n', encoding='utf-8')
```
RonnyPfannschmidt: Needs to use the `tool.setuptools_scm` section

setup.py
```diff
@@ -110,6 +113,7 @@ def parse(root):
         "Topic :: System :: Software Distribution",
         "Topic :: Utilities",
     ],
+    install_requires=["toml"],
```
RonnyPfannschmidt: This should use an extra for now

RonnyPfannschmidt (commented):
This is shaping up nicely, I like it

setup.py
RonnyPfannschmidt: I just noted the last comment which means just that

testing/test_config.py
RonnyPfannschmidt:
```suggestion
    fn.write_text('[tool.setuptools_scm]\nenabled = true\n', encoding='utf-8')
```

src/setuptools_scm/config.py
```diff
@@ -105,3 +105,14 @@ def tag_regex(self):
...
+
+    @classmethod
+    def from_file(cls, name='pyproject.toml'):
+        """
+        Read Configuration from pyproject.toml (or similar)
+        """
+        with open(name) as strm:
+            defn = __import__('toml').load(strm)
+        config = cls()
+        vars(config).update(defn.get('setuptools_scm', {}))
```
RonnyPfannschmidt:
```suggestion
        vars(config).update(defn.get("tool",{}).get('setuptools_scm', {}))
```

RonnyPfannschmidt: this fun hack is definitively a little bit evil ^^

setup.py
jaraco: Done in 6fb77d9

testing/test_config.py
```diff
@@ -18,3 +20,10 @@ def test_tag_regex(tag, expected_version):
     match = config.tag_regex.match(tag)
     version = match.group("version")
     assert version == expected_version
+
+
+def test_config_from_pyproject(tmpdir):
+    fn = tmpdir / "pyproject.toml"
+    fn.write_text("[tool.setuptools_scm]\nenabled = true\n", encoding="utf-8")
```
RonnyPfannschmidt: i would like to suggest write_bytes + a bytes literal at this point

jaraco: In my opinion, unicode literals are more friendly and the intended use here (a6ae830).

src/setuptools_scm/config.py
```diff
@@ -49,6 +49,7 @@ class Configuration(object):
     parse = None
     _tag_regex = None
     _absolute_root = None
+    enabled = False
```
RonnyPfannschmidt: this needs to be handled in the legacy integration (setup_keyword)

since 
```
In [2]: toml.loads("[foo.bar]")                                                                                                                                                                                                                              
Out[2]: {'foo': {'bar': {}}}
```
i believe we should use the existence of the section as indicator for trigger rather than introducing a extra boolean

jaraco: > i believe we should use the existence of the section as indicator for trigger rather than introducing a extra boolean

Yes, I think I agree with that. I do wish to avoid booleans in general.

My reasoning behind having the boolean was that I was imagining some users might have `setuptools_scm` present and some configuration present for a number of reasons:

- they have a draft of the config but aren't yet ready to enable it
- they want to control some other aspect of setuptools_scm not related to versioning (file finding or maybe some future functionality)

Basically, I worry that if the section is `tool.setuptools_scm`, that's to general to imply "infer a version from scm metadata". If the section were `tool.setuptools_scm.infer_version`, I would agree whole-heartedly.

RonnyPfannschmidt: as the key feature of the package is version finding from scm metadata,
if people want to write up a draft and "hide it", they can use `tool._setuptools_scm`

in this case i believe that trying to account for non-existing or currently non-configurable features would constitute a premature optimization

should a new need arise, it needs to get a new name in the tool section

jaraco: Done in bf15ea9.

RonnyPfannschmidt (approved):
great work, what about the wip status

jaraco:
I've got a few more commits to address the docs.

RonnyPfannschmidt:
thanks for implementing and documenting this

with this fabulous work many are finally free of `setup.py` 

i plan to release mid week after including the ci/pre-commit fixups

README.rst
```diff
@@ -13,9 +13,63 @@ It also handles file finders for the supported SCMs.
...
 
+``pyproject.toml`` usage
+------------------------
+
+The preferred way to configure ``setuptools_scm`` is to author
+settings in a ``tool.setuptools_scm`` section of ``pyproject.toml``.
+
+This feature requires Setuptools 42 or later, released in Nov, 2019.
+If your project needs to support build from sdist on older versions
+of Setuptools, you will need to also implement the ``setup.py usage``
```
webknjaz, line 24: Looks like you placed the closing backticks too far...

madig:
Just chiming in to say FUCK YEAH. Thanks! One less file lying around 👍

README.rst
jaraco: I wasn't quite sure how to handle that, given that the backticks are in the title. I'm happy to accept edits on it.

webknjaz: Oh. I'm not sure either. Could probably try using a link syntax.

PR #371, opened 2019-11-25
Upgrade virtualenv to ensure late pip and old pytest

### LINKED ISSUE #370: Appveyor tests failing on Python 3.4
Looking at the [CI tests](https://ci.appveyor.com/project/RonnyPfannschmidt/setuptools-scm-13mi5/builds/29088646/job/8898m9jn5aedyhan), they seem to be failing due to pytest 5 being installed on Python 3.4. Probably that's because tox is installed with some old virtualenv satistfying it, getting an old pip in the virtualenvs where pytest is installed.

jaraco:
Fixes #370

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,6 +20,7 @@ init:
   - python -m pip install -U pip
   - python -m pip install -U wheel
   - python -m pip install -U tox
+  - python -m pip install -U virtualenv
 
 
 install:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,6 +17,7 @@ init:
 
   - ECHO "Updating Environment"
   - python -m pip install -U pip
+  - python -m pip install -U wheel
   - python -m pip install -U --upgrade-strategy=eager tox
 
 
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -16,10 +16,8 @@ init:
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
 
   - ECHO "Updating Environment"
-  - python -m pip install -U setuptools
   - python -m pip install -U pip
-  - python -m pip install -U wheel
```
RonnyPfannschmidt, line 21: it seems both are needed for a after test action,, i dont recall why that one was necessary however

jaraco: Hmm. That's odd. I looked for steps in the appveyor.yml that relied on it, and I didn't see any... and the tests seem to fail within the test run, not after. I'll investigate and see if I can work out a solution for those failing tests so that the test suite isn't dependent on factors outside of tox.

jaraco: Nevermind, I [see it now](https://github.com/pypa/setuptools_scm/blob/8a69f1ce3ab532e1fd1bfb081c9932c1063c96d3/appveyor.yml#L39).

RonnyPfannschmidt (approved):
thanks, please  merge when you think its ready

PR #373, opened 2019-11-25
Fix the Tidelift badge in README

webknjaz:
Ref: https://forum.tidelift.com/t/new-urls-for-project-badges/288


webknjaz:
@RonnyPfannschmidt I imagine you'd want to merge this before making a release so that it'd hit PyPI updated.

PR #379, opened 2019-12-09
fix TOML examples

mrg29:
One of the `pyproject.toml` examples in the readme is busted:
```toml
write_to = pkg/version.py
```
it needs to be quoted:
```toml
write_to = "pkg/version.py"
```


RonnyPfannschmidt:
good catch, will investigate the ci failure when im no longer ill

RonnyPfannschmidt:
merging now, the ci fixup will happen later

PR #383, opened 2019-12-13
switch to github action

RonnyPfannschmidt:


nicoddemus:
I narrowed it down the windows problem to the `scm_find_files` function:

Depending on the system's temp location and system settings, the path might be using the "short DOS format", here in my system I get the same failure and the variables are:

```python
realdirpath = 'c:\\users\\brunoe~1.co\\appdata\\local\\temp\\pytest-of-bruno\\pytest-3\\test_git_archive_run_from_subd0\\wd\\foobar'
scm_dirs = set(['c:\\users\\bruno.esss.co\\appdata\\local\\temp\\pytest-of-bruno\\pytest-3\\test_git_archive_run_from_subd0\\wd', 'c:\\users\\bruno.esss.co\\appdata\\local\\temp\\pytest-of-bruno\\pytest-3\\test_git_archive_run_from_subd0\\wd\\foobar'])
```

So it is trying to find `c:\\users\\brunoe~1.co\\...` (short version) in the set, which only contains the full version (`c:\\users\\bruno.esss.co\...`), that's why it is failing to find any scm files.

All tests pass if I use `--basetemp tmp` in the command-line, because I'm in a directory which doesn't have this problem.

Unfortunately it doesn't seem like any of the `os.path` functions can resolve the path properly, but `pathlib` can:

```python
>>> import os, pathlib
>>> p = 'c:\\users\\brunoe~1.co'
>>> os.path.normpath(p)
'c:\\users\\brunoe~1.co'
>>> os.path.normcase(p)
'c:\\users\\brunoe~1.co'
>>> os.path.realpath(p)
'c:\\users\\brunoe~1.co'
>>> os.path.abspath(p)
'c:\\users\\brunoe~1.co'
>>> pathlib.Path(p).resolve()
WindowsPath('C:/Users/bruno.ESSS.CO')
```

Also using `pathlib` here would make the overall logic much simpler.

RonnyPfannschmidt:
i just noted that i have this branch on the computer that wont be accessible until the end of the year,

PR #384, opened 2019-12-14
fix typo README.rst

aodag:
`tools` is incorrect name of section.

- https://www.python.org/dev/peps/pep-0518/#json-schema
- https://github.com/pypa/setuptools_scm/blob/master/src/setuptools_scm/config.py#L123

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ To enable version inference, add this section to your pyproject.toml:
 .. code:: ini
 
     # pyproject.toml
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt (approved):

kryskool (approved):

RonnyPfannschmidt:
thanks for the fix, i'm just getting around my opensource backlog as the Christmas/new-year holiday had me occupied otherwise

PR #387, opened 2020-01-07
Mention importlib.metadata as alternative to pkg_resources

layday:
Split the `pkg_resources` example into a separate
section because this is not specific to setup.py.


RonnyPfannschmidt (approved):
looks good, thanks

i plan to catch up on my backlog within this week or on the weekend

PR #346, opened 2019-07-18
git: DEFAULT_DESCRIBE: add --first-parent

blueyed:
This will ignore tags from merged release branches.

Available since Git 1.8.4: https://github.com/git/git/commit/e00dd1e9485c50f202cc97dfae19d510e108b565

Issue #395, opened 2020-01-18
tag_regex not working with setuptools_scm 3.4.1


pkolbus:
With setuptools_scm 3.4.1, custom tag_regex expressions are not working correctly.

My setup.py contains:

    ...
    setup_args["use_scm_version"] = {
        "tag_regex": r"^(?P<prefix>.*/)?(?P<version>\d+(?:\.\d+){1,2})$"
        }
    setup_args["setup_requires"] = ['setuptools_scm']

    setuptools.setup(**setup_args)

setuptools_scm 3.4.0 works:

    $ python3 setup.py --version
    0.4.1.dev7+gc2397c5'

With 3.4.1:

    $ python3 setup.py --version
    /home/kolbus/.local/lib/python3.6/site-packages/setuptools_scm/version.py:92: UserWarning: tag 'development/0.4.0' no version found
      warnings.warn("tag {!r} no version found".format(tag))
    Traceback (most recent call last):
    ...
      File "/home/kolbus/.local/lib/python3.6/site-packages/setuptools_scm/git.py", line 135, in parse
    branch=branch,
      File "/home/kolbus/.local/lib/python3.6/site-packages/setuptools_scm/version.py", line 205, in meta
        assert parsed_version is not None, "cant parse version %s" % tag
    AssertionError: cant parse version development/0.4.0

RonnyPfannschmidt:
please provide the full traceback

RonnyPfannschmidt:
also could this be related to https://github.com/pypa/setuptools_scm/pull/346/files ?

blueyed:
@RonnyPfannschmidt the issue is between 3.4.0 and 3.4.1.

RonnyPfannschmidt:
on careful review i noted that i missed a line when sorting out the options

pkolbus:
Verified. Thanks for the quick fix!

RonnyPfannschmidt:
Thanks for the verification, I will release shortly

PR #396, opened 2020-01-19
Revert "git: DEFAULT_DESCRIBE: add --first-parent"

RonnyPfannschmidt:
Reverts pypa/setuptools_scm#346

as  it turns out i missed how badly this breaks workflows of people

CC @blueyed

### DIFF
--- a/src/setuptools_scm/git.py
+++ b/src/setuptools_scm/git.py
@@ -12,7 +12,7 @@ except ImportError:
     from .win_py31_compat import samefile
 
 
-DEFAULT_DESCRIBE = "git describe --dirty --tags --long --match *.* --first-parent"
+DEFAULT_DESCRIBE = "git describe --dirty --tags --long --match *.*"
 
 
 class GitWorkdir(object):


blueyed:
Fair enough.

But JFI: it is rather uncommon to merge tags (as pytest does it) - normally you have the tag on the branch itself.
`--first-parent` then ensures that you do not use tags from release/hotfix branches then (where the release tag is on the release branch).
I assume pytest also wants to use `--first-parent` once a more common release / branch workflow is used.

RonnyPfannschmidt:
@blueyed as far as i  understood, it breaks the well used git flow style (which fits the branch names in #395  )

blueyed:
@RonnyPfannschmidt seems not related (https://github.com/pypa/setuptools_scm/issues/395#issuecomment-575992836).

Re git-flow: tags are on master there (https://danielkummer.github.io/git-flow-cheatsheet/ - in case that's the one you mean)

blueyed:
I agree though in general that it should have at least been mentioned in the changelog.. ;)
And I can see that it affects/breaks uncommon flows, of course.
So go ahead with reverting I'd say - maybe there is some heuristic that could be used however, given that previously it did not work as expected with the "git-flow style".

RonnyPfannschmidt:
thanks for putting in some extra refs/ding the extra research

PR #400, opened 2020-01-21
Fix #399: ensure the git file finder closes file descriptor after reading archive

### LINKED ISSUE #399: Hangs on Windows
When using setuptools_scm on Windows, builds hang and don't proceed. This occurs because of the `proc.wait()` in `def _git_ls_files_and_dirs` in `file_finder_git.py`. After `_git_interpret_archive` returns, the call to `proc.wait()` never returns. This is because the stdout of the process is not empty. It is filled with a bunch of `\x00` bytes.

Solutions that seem to work:

1. Change `proc.wait()` to `proc.terminate()` or `proc.kill()`
2. Before calling `proc.wait()` call `proc.stdout.close()` to close the open fd.
3. Before calling `proc.wait()` call `proc.stdout.read()` to read out the fd.

Tested while building robotpy/robotpy-wpilib. setuptools_scm==3.3.3 works but the latest release does not.

Output of `python setup.py develop` when the hang occurs:
```
running develop
running egg_info
writing wpilib.egg-info\PKG-INFO
writing dependency_links to wpilib.egg-info\dependency_links.txt  
writing entry points to wpilib.egg-info\entry_points.txt
writing requirements to wpilib.egg-info\requires.txt
writing top-level names to wpilib.egg-info\top_level.txt
```

Data in `proc.stdout` after `_git_interpret_archive` returns:
`\x00\x00\x00...` (9728 null bytes)

gesslerpd:


RonnyPfannschmidt (approved):
Thanks for the pr, do you by chance have a good idea for a test that would fail without this patch

In any case I'll merge and release today

gesslerpd:
Wasn't able to get the hanging issue to reproduce in the tests, but was seeing the problem in practice since Friday on windows specifically.

I added a `proc.stdout.write(b'\x00' * 9000)` to emulate the investigation mentioned in the issue and was able to emulate the hanging on windows using that.

gesslerpd:
Have some more info, git only on windows will add a ton of zero bytes to the end of the archive file but it still is a valid tar file regardless.

Git on Linux run on the same repo will create an archive without these zero bytes, which is why this hanging problem only happens on windows.

Also, not every git repo on windows will have extra zero bytes added to the archive, it may depend on the repository state so it's not a 100% reproducible issue on windows.

TheTripleV:
This PR does fix the issue on the repo mentioned in the issue: robotpy/robotpy-wpilib

RonnyPfannschmidt:
sorry for the delay, the toddler was sick

PR #404, opened 2020-01-25
#Fix 403: Close file descriptor before terminating process

TheTripleV:
This fix passes the test `test_file_finder_no_history` in `test_git.py` .


RonnyPfannschmidt:
thanks for sorting that out

PR #405, opened 2020-02-07
no-local-version scheme and improved documentation

con-f-use:
* Adds an example for how to use entry point version schemes in a `setup.py` that was sorely missing
 * Documents the existence of the `python-simplified-semver` local version scheme
 * Adds `no-local-version` scheme, which is especially useful because pypi.org does not support local versions in packages
 * Includes tests for the aforementioned new scheme
 * Adds a changelog entry for these changes

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -100,7 +100,11 @@ Arguments to ``get_version()`` (see below) may be passed as a dictionary to
     from setuptools import setup
     setup(
         ...,
-        use_scm_version = {"root": "..", "relative_to": __file__},
+        use_scm_version = {
+            "root": "..",
+            "relative_to": __file__,
+            "local_scheme": "node-and-timestamp"
+        },
         setup_requires=['setuptools_scm'],
         ...,
     )
@@ -463,6 +467,7 @@ Version number construction
 
     :guess-next-dev: automatically guesses the next development version (default)
     :post-release: generates post release versions (adds :code:`postN`)
+    :python-simplified-semver: basic semantic versioning similar to ``guess-next-dev``
 
 ``setuptools_scm.local_scheme``
     Configures how the local part of a version is rendered given a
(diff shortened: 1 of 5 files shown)


RonnyPfannschmidt (approved):
thanks, this looks good at first glance,
im going to look at merge/release after a good night of sleep

con-f-use:
Thank you. Even if you decide not to merge the code changes, the documentation should be updated. It's just two lines or so, but they would have made a difference of an hour for me, way back when. :smiley:

README.rst
```diff
@@ -100,7 +100,11 @@ Arguments to ``get_version()`` (see below) may be passed as a dictionary to
     from setuptools import setup
     setup(
         ...,
-        use_scm_version = {"root": "..", "relative_to": __file__},
+        use_scm_version = {
+            "root": "..",
+            "relative_to": __file__,
+            "local_scheme": "node-and-timestamp"
```
RonnyPfannschmidt, line 106:
```suggestion
            "local_scheme": "node-and-timestamp"  # the default
```

con-f-use: Hmm, `config.py` says: `DEFAULT_LOCAL_SCHEME = "node-and-date"`...

RonnyPfannschmidt: good catch, i think i clearly need more coffee

pelson:
:+1: This is a really helpful feature for a usecase I have (I found it by reading the README on master a few days ago). Just wondering what the timescale is for releasing this? Is there anything I can do to help?

RonnyPfannschmidt:
Thanks for the ping, im caught up with personal life, but as the pr has a changelog ill just tag a release

pelson:
Thanks @RonnyPfannschmidt.
I didn't want to put any pressure on (hence the offer to help if I can). Much appreciated. :+1:

Issue #391, opened 2020-01-14
ScmVersion time sould be UTC for (dirty) repo with local_scheme 'node-and-date' or 'node-and-timestamp'


con-f-use:
I noticed that  for a repository (maybe only if it's dirty), there is no time zone information passed to the `local_scheme` of choice and the time information is in local time:

```python
from setuptools import setup

def loc_vers(*a, **kw):
    scm = a[0]
    print(scm.time)  # 2020-01-14 14:56:16.031252 (my local time, UTC would be an hour earlier)
    print(scm.time.tzname())  # None
    return ""

setup(
    use_scm_version=dict(local_scheme=loc_vers),
    # ...
)
```

Seeing that (linux epoch) timestamps are supposed to be UTC and git repository timestams are UTC as well, I think this should be UTC or at least have a `tzinfo` set.

con-f-use:
I've been doing a little thinking on this one. For the normal `ScmVersion.time`, we should use datetime.datetime.now(datetime.timezone.utc)` instead of just `datetime.datetime()`. That would include timezone information make the time location independent. Since it's part of the version and version should hopefully be location independent, that makes sense to me.

Also, I'd like to add a new attribute to the `ScmVersion` class that tracks the "commit time", e.g. for git that would come from `git show -s --format=%ct` which is a UNIX timestamp. The thinking here is, that `node-and-date` and `node-and-timestamp` local schemes should use:

 * The system's local time when the commit is dirty
 * The time of commit, when the commit is not dirty

Because currently, we get a new version (i.e. a higher timestamp) even if a package is build a second time from the same unchanged commit.

In any case, all times should have timezone information or be in UTC.

How does that sound conceptually?

@RonnyPfannschmidt  at some point I'd love a talk with if you're still in Innsbruck and the whole virus-thing is over.

RonnyPfannschmidt:
i like the general direction, the exact details will be a bit more tricky,
we may be ok to use `.utcnow()` instead of now

as for having a talk, i strongly propose a virtual one as we don't know what the future may bring

con-f-use:
Agreed, I was planning to discuss this issue further here. I think a nice solution is `datetime.datetime.now(datetime.timezone.utc)` because the date and time are part of the version and version should be (ideally) location independent :wink: 

Sadly `.utcnow()` does not set `tzinfo`:

```
>>> import datetime
>>> datetime.datetime.now(datetime.timezone.utc) 
[...]

PR #408, opened 2020-03-18
 Change ScmVersion.time to UTC

con-f-use:
See #391


src/setuptools_scm/version.py
```diff
@@ -141,7 +141,7 @@ def __init__(
             distance = 0
         self.distance = distance
         self.node = node
-        self.time = datetime.datetime.now()
+        self.time = datetime.datetime.now(utc)
```
RonnyPfannschmidt, line 144: im wondering if a switch to `utcnow` would make sense for the older pythons

con-f-use: My problem with that is, it  creates a `datetime` object with utc time, but does not attach timezone information to the object.

con-f-use: If you really really want me to use `utcnow`, please give me another ping, and I'll change the PR accordingly. But I recommend against it.

RonnyPfannschmidt: It's fine as is, ill coin this as a major release, im still adjusting to daycare shutdown and the house arrest from corona, since I have a toddler i chances for Open-source are rare

RonnyPfannschmidt:
thanks, next release will be a major one

PR #410, opened 2020-03-22
Allow getting the version from the parent directory suffix.

anntzer:
Add a `parentdir_prefix_version` config parameter which allows lets
setuptools_scm get a version string from the name of the parent
directory: if `parentdir_prefix_version = "foo-v"` and the parent
directory is named `foo-v12.34`, then the version is 12.34.  This is
only active if the parameter is set (it defaults to unset, i.e. the
feature is opt-in), if getting the version from the SCM failed, *and*
there's no metadata already present.  However, this has higher priority
than `fallback_version` (which remains the last-resort fallback).

This feature is intended to support the tarballs that GitHub
automatically creates for all releases, which have no metadata but
follow the correct naming scheme for the feature to work.

This feature is directly inspired from versioneer's `parentdir_prefix`
setting.

(I am open to changing the name of the param, of course.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+next
+====
+
+* Add ``parentdir_project_version`` to support installs from GitHub release
+  tarballs.
+
 v3.5.1
 ======
 
--- a/README.rst
+++ b/README.rst
@@ -370,6 +370,18 @@ The currently supported configuration keys are:
     Defaults to the value of ``setuptools_scm.config.DEFAULT_TAG_REGEX``
     (see `config.py <src/setuptools_scm/config.py>`_).
 
+:parentdir_prefix_version:
+    If the normal methods for detecting the version (SCM version,
+    sdist metadata) fail, and the parent directory name starts with
+    ``parentdir_prefix_version``, then this prefix is stripped and the rest of
+    the parent directory name is matched with ``tag_regex`` to get a version
+    string.  If this parameter is unset (the default), then this fallback is
+    not used.
+
+    This is intended to cover GitHub's "release tarballs", which extract into
+    directories named ``projectname-tag/`` (in which case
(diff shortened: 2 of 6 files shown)


RonnyPfannschmidt:
Thanks, i like it as it enhances a number of things,

I can't promise timely review as personal time is currently used mostly for offsetting the daycare shutdown

RonnyPfannschmidt (changes requested):
lovely, i like the general approach,
but i would like to see the version regex be applied to the unpack folder basename for handling interesting tagging schemes/canges better and to make things simpler on users

additionally the setuptools integration could potentially use this as fallback for when the folder name matches the distribution name (but thats out of scope for now, just a follow up to keep in mind as its not clear how safe it is to do)

to be safe in general, it may be necessary to consider relative_to the same way it is done to match setup.py location vs repo root location

im not sure when i will find the time to test out the finder details i just took into consideration

anntzer:
I switched to passing the string to tag_regex (I don't have a strong opinion there).  I think it's already the resolved root that gets passed to the parsers in hacks.py -- or if it isn't, how do I get it?

RonnyPfannschmidt:
thanks for verification, its indeed already handled

RonnyPfannschmidt (approved):
looks good now, thanks

CHANGELOG.rst
```diff
@@ -1,3 +1,9 @@
+next
+====
+
+* Add ``parentdir_project_version`` to support installs from GitHub release
```
jab, line 4: Typo: `parentdir_prefix_version`, not `parentdir_project_version`

jsy1001:
I would like to ask, is this feature intended to work when installing a release tarball directly (i.e. `pip install myproject-v1.2.3.tar.gz`) or only if the tarball is unpacked first? I have tried it with setuptools_scm==7.0.5 and only the latter use case seems to work.

Issue #411, opened 2020-03-25
Default tag regex fails for single-digit versions


Flameeyes:
```
>>> import re
>>> DEFAULT_TAG_REGEX = r"^(?:[\w-]+-)?(?P<version>[vV]?\d+(?:\.\d+){0,2}[^\+]+)(?:\+.*)?$"
>>> re.search(DEFAULT_TAG_REGEX, 'usbmon-tools-1')
>>> re.search(DEFAULT_TAG_REGEX, 'usbmon-tools-1.0')
<re.Match object; span=(0, 16), match='usbmon-tools-1.0'>
>>> re.search(DEFAULT_TAG_REGEX, 'usbmon-tools-10')
<re.Match object; span=(0, 15), match='usbmon-tools-10'>
```

The reason is that `[^\+]+` that is looking for _one or more_ characters that are not `+`.

I'll try to send a pull request to fix this.

RonnyPfannschmidt:
closed by #412

PR #412, opened 2020-03-25
Fix documentation for tag_regex and allow single-digit version numbers.

Flameeyes:
This fixes Issue #411, and reduces confusion by matching documentation to code.

### DIFF
--- a/src/setuptools_scm/config.py
+++ b/src/setuptools_scm/config.py
@@ -6,7 +6,7 @@ import warnings
 
 from .utils import trace
 
-DEFAULT_TAG_REGEX = r"^(?:[\w-]+-)?(?P<version>[vV]?\d+(?:\.\d+){0,2}[^\+]+)(?:\+.*)?$"
+DEFAULT_TAG_REGEX = r"^(?:[\w-]+-)?(?P<version>[vV]?\d+(?:\.\d+){0,2}[^\+]*)(?:\+.*)?$"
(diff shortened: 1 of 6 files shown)


Flameeyes:
Ah I found a second problem: the `git-describe --match *.*` command would ignore my non-dotted version, as well.

src/setuptools_scm/git.py
```diff
@@ -12,7 +12,7 @@
     from .win_py31_compat import samefile
 
 
-DEFAULT_DESCRIBE = "git describe --dirty --tags --long --match *.*"
+DEFAULT_DESCRIBE = "git describe --dirty --tags --long --match *[0-9]*"
```
RonnyPfannschmidt, line 15: that one is a good find

src/setuptools_scm/config.py
```diff
@@ -6,7 +6,7 @@
 
 from .utils import trace
 
-DEFAULT_TAG_REGEX = r"^(?:[\w-]+-)?(?P<version>[vV]?\d+(?:\.\d+){0,2}[^\+]+)(?:\+.*)?$"
+DEFAULT_TAG_REGEX = r"^(?:[\w-]+-)?(?P<version>[vV]?\d+(?:\.\d+){0,2}[^\+]*)(?:\+.*)?$"
```
RonnyPfannschmidt, line 9: i believe this fix is actually correct, and by now im horrified enough of the fine details of this regex, that i want to transform it into a verbose commented regex in a future followup

PR #417, opened 2020-04-06
fallback if configurated

chitran96:
When the git describe command return fail, we should return fallback_version if it is present instead of "0.0".

### DIFF
--- a/src/setuptools_scm/git.py
+++ b/src/setuptools_scm/git.py
@@ -110,11 +110,16 @@ def parse(
         rev_node = wd.node()
         dirty = wd.is_dirty()
 
+        fallback_version = "0.0"
+
+        if config.fallback_version:
+            fallback_version = config.fallback_version
+
         if rev_node is None:
-            return meta("0.0", distance=0, dirty=dirty, config=config)
+            return meta(fallback_version, distance=0, dirty=dirty, config=config)
 
         return meta(
-            "0.0",
+            fallback_version,
             distance=wd.count_all_nodes(),
             node="g" + rev_node,
             dirty=dirty,


RonnyPfannschmidt (changes requested):
It's expected to fail if there are no tags, aka when a repository starts out

For shallow clones it's expect to fail clear unless configured to fetch missing data.

chitran96:
> It's expected to fail if there are no tags, aka when a repository starts out
> 
> For shallow clones it's expect to fail clear unless configured to fetch missing data.

Yep it can be failed but why return "0.0" instead of config.fallback_version? 

A version string that will be used if no other method for detecting the version worked (e.g., when using a tarball with no metadata). If this is unset (the default), setuptools_scm will error if it fails to detect the version.
--

src/setuptools_scm/git.py
```diff
@@ -111,16 +111,29 @@ def parse(
...
-
-        return meta(
-            "0.0",
-            distance=wd.count_all_nodes(),
-            node="g" + rev_node,
-            dirty=dirty,
-            branch=wd.get_branch(),
-            config=config,
-        )
+            if config.fallback_version:
```
nagrubb: you could make this a little bit cleaner with
```
fallback_version = "0.0"

if config.fallback_version:
  fallback_version = config.fallback_version
```
... then use `fallback_version` instead of `"0.0"`.

nagrubb: Also consider adding `expandvars(fallback_version)`

chitran96: > expandvars

why need expandvars?

RonnyPfannschmidt:
Currently fallback version is for when the scm completely fail, not to argument missing Metadata for incomplete checkouts, im not sure what the best way forward

src/setuptools_scm/git.py
nagrubb: to use environment variables in the fallback_version, but that could be a different change

nagrubb (approved):

chitran96:
> Currently fallback version is for when the scm completely fail, not to argument missing Metadata for incomplete checkouts, im not sure what the best way forward

> e.g., when using a tarball with no metadata

This case is supposed to return fallback_version in the document, so an incomplete checkout is kind of it?

RonnyPfannschmidt:
which is why https://github.com/pypa/setuptools_scm/blob/69eb0fa5660645a9b28c078b5b3e8dddc74267fa/src/setuptools_scm/git.py#L65-L88 is provided for example

from my pov a incomplete  checkout is a redeemable lack of metadata while a broken download is not

which reminds me, the fallback should attach a local version scheme tag to indicate its a fallback and not suitable for official package repos

RonnyPfannschmidt:
also please excuse my rather terse answers, i'm currently a bit stressed

i appreciate that you want to ensure there are more cases where people get workable metadata,
i just want to ensure setuptools_scm never provides metadata that is misleading by accident

an error one has to fix before it does damage is better than an error one has to fix after it does damage

RonnyPfannschmidt:
i finally found some time to work on setuptools_scm

unfortunately i have to reject this contribution as of now as the danger of accidentally wrong metadata seems to be a larger issue to me for now

i still would like to make the gt interaction/error experience better and would appreciate input/suggestions on how you would expect things to work

thanks for experimenting and i hope to hear more from you in future

PR #425, opened 2020-05-06
Extend error message if "unable to detect version"

MaxGaukler:
In some cases it's not possible to include the .git subdirectory and the version number doesn't matter. Then, the error message "unable to detect version, make sure your git repository is intact" doesn't help. Add a note on SETUPTOOLS_SCM_PRETEND_VERSION so users spend less time googling for a workaround.

(I do understand that this is not the recommended use-case, but there are situations in which it cannot be avoided without even uglier workarounds.)


RonnyPfannschmidt:
I can understand where you are coming from but I rather thorn on whether to include that as is, ill sleep over it

RonnyPfannschmidt:
the file `.setuptools_scm_fallback_version` should be introduced as additional fallback that can be created

PR #426, opened 2020-05-06
Skip mercurial tests when hg is not available

mgorny:
Mercurial is broken (read: crashing with bus error) on SPARC.
Nevertheless, many packages need setuptools_scm even if we never ever
use Mercurial checkouts.  This patch makes it possible to run the test
suite successfully when Mercurial is not available on the system.

### DIFF
--- a/testing/test_mercurial.py
+++ b/testing/test_mercurial.py
@@ -2,7 +2,15 @@ from setuptools_scm import format_version
 from setuptools_scm.hg import archival_to_version, parse
 from setuptools_scm import integration
 from setuptools_scm.config import Configuration
+from setuptools_scm.utils import has_command
 import pytest
+import warnings
+
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore")
+    if not has_command("hg"):
+        pytestmark = pytest.mark.skip(reason="hg executable not found")
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


testing/test_mercurial.py
```diff
@@ -2,7 +2,15 @@
 from setuptools_scm.hg import archival_to_version, parse
 from setuptools_scm import integration
 from setuptools_scm.config import Configuration
+from setuptools_scm.utils import has_command
 import pytest
+import warnings
+
+
+with warnings.catch_warnings():
+    warnings.filterwarnings('ignore')
```
mgorny: Note that I hate this but I have no clue how to otherwise make pytest not crash about `has_command()` throwing a warning. Nor that I have any clue why a command returning a bool would issue a warning that it returns False.

RonnyPfannschmidt:
Thanks, I will merge as is and refine later

If you like you can take a look at using xfail for this

RonnyPfannschmidt (approved):

mgorny:
> If you like you can take a look at using xfail for this

I don't think xfail is correct here. The way I understand it, xfail means that the package or tests are broken. For unsatisfied dependency (which is a clear cut case of 'we cannot test this properly') I think skip is more correct.

mgorny:
By the way, I can look at doing something similar for git, if you're interested.

RonnyPfannschmidt:
i am, if you like you can also take a look at enhancing the error message we give when the detected scm is not installed

PR #430, opened 2020-05-11
Add release-branch-semver scheme

chrisjbillington:
This is a scheme that assumes that the upcoming release for most branches is a minor release, and that only for release branches is the upcoming version a patch release. This matches the workflow used by many projects who do development in master with feature branches, and fork off release branches that have bugfix-only changes made to them.

The scheme only considers the current branch to be a release branch if its name (stripped of any namespaces) parses as a version that matches the most recent tag up to the minor segment. Thus a tag `1.2.0rc0`for example will be matched by `1.2`, `1.2.x`, `maintenance/1.2`, `release-1.2.x` etc, but not `bugfix/123123`, since although the latter parses as a `LegacyVersion` `123123.0.0`, it doesn't match the tag up to the minor segment. So that's intended to reduce false positives from branch names that happen to parse as versions even though they're not.

This PR also increases the level of detail documenting the differences between the schemes in `README.rst`

### DIFF
--- a/src/setuptools_scm/version.py
+++ b/src/setuptools_scm/version.py
@@ -132,6 +132,7 @@ class ScmVersion(object):
         dirty=False,
         preformatted=False,
         branch=None,
+        config=None,
         **kw
     ):
         if kw:
@@ -146,6 +147,7 @@ class ScmVersion(object):
         self.dirty = dirty
         self.preformatted = preformatted
         self.branch = branch
+        self.config = config
 
     @property
     def extra(self):
@@ -193,7 +195,14 @@ def _parse_tag(tag, preformatted, config):
 
 
 def meta(
-    tag, distance=None, dirty=False, node=None, preformatted=False, config=None, **kw
+    tag,
+    distance=None,
+    dirty=False,
+    node=None,
+    preformatted=False,
+    branch=None,
+    config=None,
+    **kw
 ):
     if not config:
         warnings.warn(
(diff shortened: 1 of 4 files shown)


src/setuptools_scm/version.py
```diff
@@ -132,6 +132,7 @@ def __init__(
         dirty=False,
         preformatted=False,
         branch=None,
+        config=None,
```
RonnyPfannschmidt, line 129: i am on the fence on whether configuration should be a parameter to versions

RonnyPfannschmidt (commented):
thanks for bringing this addition to the table, it makes a few details much more nice (and actually matches something i want/need for pytest anyway

src/setuptools_scm/version.py
chrisjbillington: That's fair enough, but if the branch name is to be parsed it needs config from somewhere - I was having it use the default config, but this way it can honour version regexes set by the user, I guess. Happy to have it just use the default config though!

RonnyPfannschmidt: i understand the necessity - i'm just wondering about how to pass in the config
it must land in the version formatter/generator somehow

and its also going to be useful for a semver-towncrier scheme, that will pick versions number based on 

i'll sleep a night over this

both the pass in as part of the version object as well as adding it as extra parameter to the version formatter have different ups and downs

RonnyPfannschmidt:
bringing this in as is for now as the other way around is potentially problematic as well and it wil lhelp the use-cases for towncrier based versioning as well

chrisjbillington:
Thank you!

PR #441, opened 2020-05-24
Add release-branch-semver to setup.cfg

chrisjbillington:
the release-branch-semver scheme doesn't seem to have survived the transition from `setup.py` to `setup.cfg`.
 
This adds it back in.
Also, I renamed the internal function to `release_branch_semver_version`, for consistency with the other functions for versions schemes.

### DIFF
--- a/src/setuptools_scm/version.py
+++ b/src/setuptools_scm/version.py
@@ -267,7 +267,7 @@ def simplified_semver_version(version):
             )
 
 
-def release_branch_semver(version):
+def release_branch_semver_version(version):
     if version.exact:
         return version.format_with("{tag}")
     if version.branch is not None:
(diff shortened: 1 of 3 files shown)


RonnyPfannschmidt:
good catch, i believe i lost this in a rebase after having fixed it before

src/setuptools_scm/version.py
```diff
@@ -267,7 +267,7 @@ def simplified_semver_version(version):
             )
 
 
-def release_branch_semver(version):
+def release_branch_semver_version(version):
```
RonnyPfannschmidt, line 270: changing the now exposed name unfortunately constitutes a breaking change, and would need a rename shim + a deprecation warning so the old name can be removed in 5.0

RonnyPfannschmidt (changes requested):
please either implement a shim for the rename or defer the renaming as a new isssue

src/setuptools_scm/version.py
chrisjbillington: Ok, no problem I'll just revert it.

chrisjbillington:
Actually I went for the shim instead.

RonnyPfannschmidt (approved):

RonnyPfannschmidt:
thanks for the swift fixup

PR #443, opened 2020-05-25
Fix missing git/hg handling in tests for py27

mgorny:
I'm sorry but apparently I haven't tested my original commits well enough.


RonnyPfannschmidt (approved):
neither i did, thanks for the followup

PR #444, opened 2020-05-25
tox.ini: Fix paths in flake8 target

mgorny:


RonnyPfannschmidt:
god catch, missed that since i don't use tox for myself

Issue #442, opened 2020-05-25
Version 4.0 breaks on Windows + 2.7


henryiii:
The latest version seems to have lost the use of the samefile backport on Windows + 2.7, causing this error now in CI:

```
  Downloading setuptools_scm-4.0.0.tar.gz (46 kB)
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  ERROR: Command errored out with exit status 1:
   command: 'C:\hostedtoolcache\windows\Python\2.7.18\x86\python.exe' 'C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\pip\_vendor\pep517\_in_process.py' get_requires_for_build_wheel 'c:\users\vssadm~1\appdata\local\temp\tmpacqthi'
       cwd: c:\users\vssadm~1\appdata\local\temp\pip-install-yxxmaf\setuptools-scm
  Complete output (22 lines):
  c:\users\vssadm~1\appdata\local\temp\pip-install-yxxmaf\setuptools-scm\src
  <pkg_resources.WorkingSet object at 0x02F22D30>
  Traceback (most recent call last):
    File "C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 280, in <module>
      main()
    File "C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 263, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 114, in get_requires_for_build_wheel
      return hook(config_settings)
    File "c:\users\vssadm~1\appdata\local\temp\pip-build-env-7itkho\overlay\Lib\site-packages\setuptools\build_meta.py", line 146, in get_requires_for_build_wheel
      return self._get_build_requires(config_settings, requirements=['wheel'])
    File "c:\users\vssadm~1\appdata\local\temp\pip-build-env-7itkho\overlay\Lib\site-packages\setuptools\build_meta.py", line 127, in _get_build_requires
      self.run_setup()
    File "c:\users\vssadm~1\appdata\local\temp\pip-build-env-7itkho\overlay\Lib\site-packages\setuptools\build_meta.py", line 142, in run_setup
[...]

PR #445, opened 2020-05-25
Add universal wheels

henryiii:
See #442

It would be best to rebuild the 4.0.0 wheel locally with the `--universal` flag and upload that; it will have a different name than the 3-only wheel that is there now, and that way users will not have to explicitly avoid 4.0.0 when building.


RonnyPfannschmidt:
thanks for the suggestion

RonnyPfannschmidt:
@henryiii the universal wheel was uploaded

Issue #440, opened 2020-05-24
The 'jaraco.windows' distribution was not found and is required by the application


aaronkvanmeerten:
While trying to install the pubnub python library, I ran into this issue: pubnub depends on setuptools_scm, which in turns seems to need this jaraco.windows package.  However, even when I did a `pip install jaraco.windows setuptools_scm pubnub` I got this error.  I eventually had to install jaraco.windows in a separate command to fulfill the dependency before I installed setuptools_scm and pubnub.

The full error message is something like this (sorry for ansible escaping):

`"/usr/local/lib/python2.7/dist-packages/setuptools/dist.py\", line 739, in finalize_options\n        for ep in sorted(eps, key=by_order):\n      File \"/usr/local/lib/python2.7/dist-packages/setuptools/dist.py\", line 738, in <lambda>\n        eps = map(lambda e: e.load(), pkg_resources.iter_entry_points(group))\n      File \"/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py\", line 2442, in load\n        self.require(*args, **kwargs)\n      File \"/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py\", line 2465, in require\n        items = working_set.resolve(reqs, env, installer, extras=self.extras)\n      File \"/usr/local/lib/python2.7/dist-packages/pkg_resources/__init__.py\", line 786, in resolve\n        raise DistributionNotFound(req, requirers)\n    pkg_resources.DistributionNotFound: The 'jaraco.windows' distribution was not found and is required by the application\n    \n    ----------------------------------------\n\n:stderr: Command \"python setup.py egg_info\" failed with error code 1 in /tmp/pip-build-FfPU4A/setuptools-scm/\n"`

I'm opening this issue here in case anyone else runs into this problem.  I wonder if there's some way to explicitly define this as a requirement to avoid needing to do this manually?

r1prakash:
Getting same error at my end:-

`    <pkg_resources.WorkingSet object at 0x7f24aef02b90>
    Traceback (most recent call last):
      File "<string>", line 20, in <module>
      File "/tmp/pip-build-CDBOV8/setuptools-scm/setup.py", line 53, in <module>
        setuptools.setup(**scm_config())
      File "/opt/rh/python27/root/usr/lib64/python2.7/distutils/core.py", line 111, in setup
        _setup_distribution = dist = klass(attrs)
[...]

PR #446, opened 2020-05-25
Python 2.7 fixes

henryiii:
Should solve #442 and #440 for a patch release. Adds windows test and some skips.


RonnyPfannschmidt (approved):
Lovely, thanks

PR #476, opened 2020-09-08
honour SOURCE_DATE_EPOCH environment variable

Villemoes:
In the interest of producing more reproducible builds, honour SOURCE_DATE_EPOCH if set.

### DIFF
--- a/src/setuptools_scm/version.py
+++ b/src/setuptools_scm/version.py
@@ -2,6 +2,8 @@ from __future__ import print_function
 import datetime
 import warnings
 import re
+import time
+import os
 
 from .config import Configuration
 from .utils import trace, string_types, utc
@@ -136,7 +138,9 @@ class ScmVersion(object):
             distance = 0
         self.distance = distance
         self.node = node
-        self.time = datetime.datetime.now(utc)
+        self.time = datetime.datetime.utcfromtimestamp(
+            int(os.environ.get("SOURCE_DATE_EPOCH", time.time()))
+        )
         self._extra = kw
         self.dirty = dirty
         self.preformatted = preformatted
--- a/src/setuptools_scm/version.py
+++ b/src/setuptools_scm/version.py
@@ -6,7 +6,7 @@ import time
 import os
 
 from .config import Configuration
-from .utils import trace, string_types, utc
+from .utils import trace, string_types
 
 from pkg_resources import iter_entry_points
 
--- a/testing/test_functions.py
+++ b/testing/test_functions.py
@@ -15,11 +15,6 @@ from setuptools_scm.utils import has_command
 PY3 = sys.version_info > (2,)
 
 
-class MockTime(object):
-    def __format__(self, *k):
-        return "time"
-
-
 @pytest.mark.parametrize(
     "tag, expected",
     [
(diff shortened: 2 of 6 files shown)


src/setuptools_scm/version.py
```diff
@@ -136,7 +138,9 @@ def __init__(
             distance = 0
         self.distance = distance
         self.node = node
-        self.time = datetime.datetime.now(utc)
+        self.time = datetime.datetime.utcfromtimestamp(
+            int(os.environ.get("SOURCE_DATE_EPOCH", time.time()))
```
RonnyPfannschmidt, line 142: i'd like to see a unittest for that

RonnyPfannschmidt (changes requested):
please excuse the late review,
for personal reason i had to put a lot into my backlog and i'm just now processing

testing/test_functions.py
```diff
@@ -36,6 +31,8 @@ def test_next_tag(tag, expected):
 
 c = Configuration()
 
+# 2009-02-13T23:31:30+00:00
+os.environ["SOURCE_DATE_EPOCH"] = "1234567890"
```
RonnyPfannschmidt: nice idea

please use `monkeypatch.setenv` inside of the test instead of doing a global patching there

else the behavior bleeds to other tests an

Villemoes: yeah, I thought about that. Problem is that, AFAICT, the `meta()` objects in the `VERSIONS` array get created at global scope, so modifying the environment within the test case doesn't affect the `.time` attribute that's already set. That is, if I do
```
 # 2009-02-13T23:31:30+00:00
-os.environ["SOURCE_DATE_EPOCH"] = "1234567890"
+#os.environ["SOURCE_DATE_EPOCH"] = "1234567890"
 VERSIONS = {
     "exact": meta("1.1", distance=None, dirty=False, config=c),
     "zerodistance": meta("1.1", distance=0, dirty=False, config=c),
@@ -60,7 +60,8 @@ VERSIONS = {
         ("distancedirty", "post-release node-and-date", "1.1.post3+d20090213"),
     ],
 )
-def test_format_version(version, scheme, expected):
+def test_format_version(version, monkeypatch, scheme, expected):
+    monkeypatch.setenv("SOURCE_DATE_EPOCH", "1234567890")
     version = VERSIONS[version]
```
on top, the test fails (as I expected) with
```
E       AssertionError: assert '1.1.post3+d20201014' == '1.1.post3+d20090213'
E         - 1.1.post3+d20201014
E         ?               - ^^^
E         + 1.1.post3+d20090213
E         ?              +++  ^
```
I don't really know much about pytest (I can't even get the full test suite to run), so I'm at a loss of how to add a real test.

Villemoes: hmm, conftest.py already sets one env variable globally (which then gets unset in a few tests). Perhaps SOURCE_DATE_EPOCH should be set there as well and used for all the date tests - after all, the `assert today.strftime(".d%Y%m%d") in wd.version in test_git.py has a microscopic chance of a false negative is the test is run very close to (UTC) midnight.

RonnyPfannschmidt: Good idea,

For pedantic we ought to have a single test that checks the no env var case

Villemoes: All good now?

RonnyPfannschmidt (approved):

testing/test_functions.py
RonnyPfannschmidt: yup, nicely done, thanks

Issue #482, opened 2020-10-24
Auto-generated write_to file triggering errors in flake8, pydocstyle, and pylint


jamesbraza:
I use the `write_to` option in my `pyproject.toml` config file to write to a Python `__version__.py` file.  It creates a file that looks like this:

```python
# coding: utf-8
# file generated by setuptools_scm
# don't change, don't track in version control
version = '0.1.dev41+ge5eee1b.d20201024'
```

---

**`flake8`**

I run [`flake8`](https://github.com/PyCQA/flake8) (PEP8 linter) on my package with the [`flake8-black`](https://github.com/peterjc/flake8-black) plugin for [`black`](https://github.com/psf/black).

Unfortunately, `flake8` fails with this message:

```none
__version__.py:4:11: BLK100 Black would make changes.
```

I have since fixed things on my side using the [`exclude`](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-exclude) configuration option for `flake8`.

---

**`pydocstyle`**

The same is true for [`pydocstyle`](https://github.com/PyCQA/pydocstyle/):

```none
__version__.py:1 at module level:
        D100: Missing docstring in public module
```

I think we can ignore `pydocstyle`, it seems superfluous to have auto-generated files be given module-level docstrings.

---

**`pylint`**

And also for [`pylint`](https://github.com/PyCQA/pylint):

```none
__version__.py:4:0: C0103: Constant name "version" doesn't conform to UPPER_CASE naming style (invalid-name)
```

I can work around the `pylint` "invalid-name" error using the [`ignore` configuration option](https://github.com/PyCQA/pylint/blob/master/examples/pylintrc#L13).

---

**The Request**

Can the auto-generated files be made compatible with PEP8 (and by the transitive property, `black`)?

jamesbraza:
Per the `pydocstyle` "D100: Missing docstring in public module" error being raised, I learned from reading [`pydocstyle`'s Exclude given files #186](https://github.com/PyCQA/pydocstyle/issues/186) of using [`flake8-docstrings`](https://pypi.org/project/flake8-docstrings/) to ignore docstring problems on a per-file basis.

My issue still stands on the PEP8 front.

RonnyPfannschmidt:
We can add a better template, but it needs a transition period/plan , as it is a breaking change

jamesbraza:
To review:
- `pydocstyle`: let's ignore that I said this... I don't agree with adding a module-level docstring
[...]

PR #483, opened 2020-10-25
Changing from repr to string format with surrounding double quotes

jamesbraza:
Changed `write_to` file option for `.py` files to be version string to be `version = "0.1.0"` as opposed to `version = '0.1.0'`.

This will bring the generated file into compliance with the [`Black` code formatter](https://github.com/psf/black)'s established standard.

This was done in relation to https://github.com/pypa/setuptools_scm/issues/482.


jamesbraza:
I am not sure why my `write_to` file  (see linked issue https://github.com/pypa/setuptools_scm/issues/482) did not include a `version_tuple`.

From the unit test `test_dump_version`, I can see that `version_tuple` would also contain `'` inside, which would also trip up `Black`.  Do you think this should be fixed as well?

@RonnyPfannschmidt let me know what you think, and feel free to close this PR if you're not interested.

RonnyPfannschmidt:
I like it

jamesbraza:
Alright, it's open for review.

Lmk if there's any more stuff you'd like to see in this @RonnyPfannschmidt

src/setuptools_scm/__init__.py
```diff
@@ -22,8 +22,8 @@
 # coding: utf-8
 # file generated by setuptools_scm
 # don't change, don't track in version control
-version = {version!r}
-version_tuple = {version_tuple!r}
+version = "{version!s}"
+version_tuple = {version_tuple!s}
```
RonnyPfannschmidt, line 26: The old variable name will have to keep the old value to prevent breaking users
The new valve will need a new name

RonnyPfannschmidt (changes requested):
Overall well done, unfortunately we have to us a new variable name for backward compatibility

Thanks!

src/setuptools_scm/__init__.py
jamesbraza: You mean as far as the `version_tuple`?  The value is not actually a string, it's still a tuple.   I just made the conversion a `s` so that the tuple written contains `""`  instead  of `''`.

I am not entirely sure what you meant to be honest.

RonnyPfannschmidt: A template like `short_version ={version_tuple[:2]}` will silently go incorrect with the change of the template variable from tuple to str

jamesbraza: Hi @RonnyPfannschmidt I apologize, but I honestly don't know what you mean.

The `version_tuple` is still a `tuple` in the printed file.  I just used a `s` conversion in the template.  Note that in the template, `version` is surrounded with quotes, and `version_tuple` is not.

All my code did was change:

```python
# Old
version_tuple = (1, 0, 'dev42')
short_version = {version_tuple[:2]}
print(short_version)   # Prints {(1, 0)}

# New
version_tuple = (1, 0, "dev42")  # All that happens is it went from 'dev42' to "dev42", it's still a tuple though
short_version = {version_tuple[:2]}
print(short_version)   # Prints {(1, 0)}
```

RonnyPfannschmidt: as the input is no longer `version_tuple` but `version_tuple_str` its not a tuple one can format as tuple

RonnyPfannschmidt: let me re-phase - if the old template before your changes was passed in as template, it should still result-in correct code
currently as the in template type changed from tuple to str it does not

jamesbraza: Oh, now I understand what you mean.  Yeah lol I was just adding a unit test case and comments to clarify that it's actually a tuple.

Okay, then this becomes harder to solve, because the `!r` conversion in `{version!r}` calls `repr`, which will have the result include single quotes.

Alternate solution:
- Retroactively modify the file written, replacing any `'foo'` with `"foo"`

Maybe this PR makes the `dump_version` code and `test_dump_version` become clunky.

What do you think?  I can implement whatever

testing/test_basic_api.py
```diff
@@ -109,9 +109,10 @@ def test_dump_version(tmpdir):
     expected_version_tuple_line = 'version_tuple = (1, 0, "dev42")'
     assert expected_version_tuple_line in lines
 
-    version_tuple = ""  # If exec() fails, the test will fail, as str is not tuple
-    exec(lines[lines.index(expected_version_tuple_line)])
-    assert isinstance(version_tuple, tuple)
+    # Assert that the version_tuple is actually a tuple
+    exec_locals = {}
```
RonnyPfannschmidt, line 113: that's absolutely the wrong test,
the input type to the format sting has to be a tuple instead of a string,
for the issue at hand it is absolutely irrelevant what the output is . the type being passed to the formatting of the template has to be set correct

PR #485, opened 2020-10-30
Add `no-guess-dev` implementation.

s0undt3ch:
The `no-guess-dev` implementation does no next version guessing, it just
adds `.devN` in case there are new commits after the tag.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+v4.2.0
+======
+
+* Add `no-guess-dev` which does no next version guessing, just adds `.post1.devN` in
+  case there are new commits after the tag
+
 v4.1.2
 =======
 
--- a/README.rst
+++ b/README.rst
@@ -497,6 +497,7 @@ Version number construction
         that matches the most recent tag up to the minor segment. Otherwise if on a
         non-release branch, increments the minor segment and sets the micro segment to
         zero, then appends :code:`.devN`.
+    :no-guess-dev: Does no next version guessing, just adds :code:`.post1.devN`
 
 ``setuptools_scm.local_scheme``
     Configures how the local part of a version is rendered given a
--- a/src/setuptools_scm/version.py
+++ b/src/setuptools_scm/version.py
@@ -300,6 +300,13 @@ def release_branch_semver(version):
     return release_branch_semver_version(version)
 
 
+def no_guess_dev_version(version):
+    if version.exact:
+        return version.format_with("{tag}")
+    else:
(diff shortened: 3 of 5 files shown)


src/setuptools_scm/version.py
```diff
@@ -300,6 +300,13 @@ def release_branch_semver(version):
     return release_branch_semver_version(version)
 
 
+def no_guess_dev_version(version):
```
RonnyPfannschmidt, line 303: this drops in a version that sorts formally wrong, postrelease_version implementation does this with a version that sorts formally right buut doesnt add dev tags

to be correct this should add `post1.devN` to denote a formally correct sorting of the generated version after the last tag

RonnyPfannschmidt (changes requested):
i like the intent, for sorting correctness we have to use a post-release tagging

src/setuptools_scm/version.py
s0undt3ch: Thanks for the review. I'll update.

s0undt3ch:
So, after taking a closer look at the post release implementation, and proper version comparison by `packaging.version.parse()`, I think this implementation doesn't make much sense without guessing the next version.

I did update the PR, but honestly, it doesn't make much sense to merge it.
I'll leave it to your decision.

RonnyPfannschmidt:
@s0undt3ch imho this variant is now slightly better than the postrelease one as it doesnt pretend to be a release in the sense that the post-release one does incompletely

to merge the docs have to be adopted, i can do this later when i get to it (currently its lucky that i find time to do)

s0undt3ch:
Ok. If you leave guidance to what you want the docs to be like, I'll update.

CHANGELOG.rst
```diff
@@ -1,3 +1,9 @@
+v4.2.0
+======
+
+* Add `no-guess-dev` which does no next version guessing, just adds `.devN` in
```
RonnyPfannschmidt: this bit needs to match the new layout

README.rst
```diff
@@ -497,6 +497,7 @@ Version number construction
         that matches the most recent tag up to the minor segment. Otherwise if on a
         non-release branch, increments the minor segment and sets the micro segment to
         zero, then appends :code:`.devN`.
+    :no-guess-dev: Does no next version guessing, just adds :code:`.devN`
```
RonnyPfannschmidt: dito

CHANGELOG.rst
s0undt3ch: Done.

RonnyPfannschmidt (approved):

PR #486, opened 2020-10-30
Relax the default git decribe command

s0undt3ch:
Even though single digit versions support was removed, the git describe command did not need to be reverted.


RonnyPfannschmidt:
good catch, thanks

PR #492, opened 2020-11-25
Add configuration options to customize version and local scheme

### LINKED ISSUE #419: support defining a simple dict of scheme format strings for get_version  in pyproject.toml tool.setuptools_scm
for example:
```toml
# pyproject.toml

[tool.setuptools_scm.scheme]
no_distance_and_clean = "{tag}"
distance_and_clean = "{next_version}.dev{distance}+{scm_letter}{revision_hash}"
no_distance_and_not_clean = "{tag}+dYYYYMMDD"
distance_and_not_clean =  "{next_version}.dev{distance}+{scm_letter}{revision_hash}.dYYYYMMDD"
```

see also https://github.com/pypa/setuptools_scm/issues/418

whitequark:
Fixes #419.

## Motivation

Let me start by saying that setuptools_scm is one of my favorite Python packages ever! :) Not only it is robust when used as intended, but it can also be stretched very far with minimal effort and support even very strange configurations--for example several of my projects actually compose the wheel version from several git repositories at once.

Most of this flexibility however requires executable code, and the trend lately is towards declarative configuration. Unfortunately, it turns out that every single one of my projects that uses setuptools_scm tweaks its schemes with Python code, which is at odds with the ecosystem.

Fortunately, I think I managed to pin down a fairly elegant way to solve the entire class of "setuptools_scm's output is almost but not quite right" problem.

## Background

This PR actually has a third iteration of this feature. I started out by adding a new kind of schemes, `custom`, which would then always format the version using the custom scheme dictionary and only that. However, the code was very hacky (the existing entry points do not accept additional configuration), the interface was not very convenient (having an empty format string as a default, having a reimplementation of the default scheme as a default, and blowing up without all four keys are all annoying), and ultimately that entire approach was inferior because is mutually exclusive with guessing the next version.

I briefly considered having two conditions (`exact` and `inexact`) for the version scheme, unlike the local scheme, but that was a bad idea for similar reasons.

Overriding the base scheme in fine grained conditions for both versions and local parts seems like the right choice to me here.

## Results

 * I think it's a good sign that all of the built-in schemes except for version guessing ones are representable as custom schemes.
 * I like [this commit](https://github.com/nmigen/nmigen/commit/0892894da3a6c2791bfb70730596a0183f85f1a9).
 * I think it's useful that `guess-next-dev` can be stopped from increasing the minor version when it runs on a dirty worktree with zero distance to a tag with just `"custom_version_scheme": {"dirty_tag": "{tag}"}`. (Why does it do that? That seems really inconvenient for testing out local changes to a checkout of a released package.)

## Unresolved questions

 * Naming?
 * How does `exact` differ from `zerodistance`? It seems like the difference is intentional but I couldn't figure out in what situation would one ever get `zerodistance` instead of `exact`.

### DIFF
--- a/src/setuptools_scm/version.py
+++ b/src/setuptools_scm/version.py
@@ -307,6 +307,13 @@ def no_guess_dev_version(version):
         return version.format_with("{tag}.post1.dev{distance}")
 
 
+def postrelease_version(version):
+    if version.exact:
+        return version.format_with("{tag}")
+    else:
+        return version.format_with("{tag}.post{distance}")
+
+
 def _format_local_with_time(version, time_format):
 
     if version.exact or version.node is None:
@@ -335,11 +342,24 @@ def get_no_local_node(_):
     return ""
 
 
-def postrelease_version(version):
-    if version.exact:
-        return version.format_with("{tag}")
-    else:
-        return version.format_with("{tag}.post{distance}")
+def customize_scheme(base_scheme, version, custom_scheme):
+    if custom_scheme is not None:
+        if version.distance:
+            if "clean_dev" in custom_scheme and not version.dirty:
+                trace("custom clean dev", custom_scheme["clean_dev"])
+                return version.format_with(custom_scheme["clean_dev"])
+            if "dirty_dev" in custom_scheme and version.dirty:
+                trace("custom dirty dev", custom_scheme["dirty_dev"])
+                return version.format_with(custom_scheme["dirty_dev"])
+        else:
+            if "clean_tag" in custom_scheme and not version.dirty:
+                trace("custom clean tag", custom_scheme["clean_tag"])
+                return version.format_with(custom_scheme["clean_tag"])
+            if "dirty_tag" in custom_scheme and version.dirty:
+                trace("custom dirty tag", custom_scheme["dirty_tag"])
+                return version.format_with(custom_scheme["dirty_tag"])
+
+    return base_scheme(version)
 
 
 def format_version(version, **config):
@@ -353,8 +373,12 @@ def format_version(version, **config):
     local_scheme = callable_or_entrypoint(
         "setuptools_scm.local_scheme", config["local_scheme"]
     )
-    main_version = version_scheme(version)
+    main_version = customize_scheme(
+        version_scheme, version, config["custom_version_scheme"]
+    )
     trace("version", main_version)
-    local_version = local_scheme(version)
+    local_version = customize_scheme(
+        local_scheme, version, config["custom_local_scheme"]
+    )
     trace("local_version", local_version)
     return main_version + local_version
--- a/src/setuptools_scm/version.py
+++ b/src/setuptools_scm/version.py
@@ -357,4 +357,4 @@ def format_version(version, **config):
     trace("version", main_version)
     local_version = local_scheme(version)
     trace("local_version", local_version)
-    return version_scheme(version) + local_scheme(version)
+    return main_version + local_version
--- a/README.rst
+++ b/README.rst
@@ -336,9 +336,15 @@ The currently supported configuration keys are:
     Configures how the local version number is constructed; either an
     entrypoint name or a callable.
 
+:custom_version_scheme:
+    Overrides format strings for the local version number; a dictionary.
+
 :local_scheme:
     Configures how the local component of the version is constructed; either an
     entrypoint name or a callable.
+
+:custom_local_scheme:
+    Overrides format strings for the local component of the version; a dictionary.
 
 :write_to:
     A path to a file that gets replaced with a file containing the current
@@ -515,6 +521,19 @@ Version number construction
     :dirty-tag: adds ``+dirty`` if the current workdir has changes
     :no-local-version: omits local version, useful e.g. because pypi does
                        not support it
+
+``setuptools_scm.custom_version_scheme``, ``setuptools_scm.custom_local_scheme``
+    Provides a greater degree of control over how the version number is rendered.
+    A dictionary specifying format strings that override the behavior of
+    ``setuptools_scm.version_scheme`` and ``setuptools_scm.local_scheme``,
+    correspondingly, in various conditions.
+
+    Recognized keys:
+
+    :clean_tag: format for zero distance to a tag and clean workdir
+    :dirty_tag: format for zero distance to a tag and dirty workdir
+    :clean_dev: format for non-zero distance to a tag and clean workdir
+    :dirty_dev: format for non-zero distance to a tag and dirty workdir
 
 
 Importing in ``setup.py``
--- a/src/setuptools_scm/__init__.py
+++ b/src/setuptools_scm/__init__.py
@@ -134,7 +134,9 @@ def _do_parse(config):
 def get_version(
     root=".",
     version_scheme=DEFAULT_VERSION_SCHEME,
+    custom_version_scheme=None,
     local_scheme=DEFAULT_LOCAL_SCHEME,
+    custom_local_scheme=None,
     write_to=None,
     write_to_template=None,
     relative_to=None,
@@ -163,7 +165,9 @@ def _get_version(config):
         version_string = format_version(
             parsed_version,
             version_scheme=config.version_scheme,
+            custom_version_scheme=config.custom_version_scheme,
             local_scheme=config.local_scheme,
+            custom_local_scheme=config.custom_local_scheme,
         )
         dump_version(
             root=config.root,
--- a/src/setuptools_scm/config.py
+++ b/src/setuptools_scm/config.py
@@ -45,7 +45,9 @@ class Configuration(object):
         relative_to=None,
         root=".",
         version_scheme=DEFAULT_VERSION_SCHEME,
+        custom_version_scheme=None,
         local_scheme=DEFAULT_LOCAL_SCHEME,
+        custom_local_scheme=None,
         write_to=None,
         write_to_template=None,
         tag_regex=DEFAULT_TAG_REGEX,
@@ -61,7 +63,9 @@ class Configuration(object):
 
         self.root = root
         self.version_scheme = version_scheme
+        self.custom_version_scheme = custom_version_scheme
         self.local_scheme = local_scheme
+        self.custom_local_scheme = custom_local_scheme
         self.write_to = write_to
         self.write_to_template = write_to_template
         self.parentdir_prefix_version = parentdir_prefix_version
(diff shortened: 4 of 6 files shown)


src/setuptools_scm/version.py
```diff
@@ -335,11 +342,24 @@ def get_no_local_node(_):
     return ""
 
 
-def postrelease_version(version):
-    if version.exact:
-        return version.format_with("{tag}")
-    else:
-        return version.format_with("{tag}.post{distance}")
+def customize_scheme(base_scheme, version, custom_scheme):
+    if custom_scheme is not None:
```
RonnyPfannschmidt, line 346: this is the reason why scheme customization breaks next version guessing

the custom scheme discards that data/opportunity

RonnyPfannschmidt (commented):
thanks for spearheading this

i believe in order to make this sensibly doable the next version guessing has  to be separated from the formatting

there is a intentional internal difference between exact and zero distance

basically setuptools_scm intentionally considers any working directory with a change on something the scm tracks on the way to a new  version

so the presumption setuptools_scm makes, that any track-able change is intended for the next version

RonnyPfannschmidt:
to elaborate a bit further, i believe there need to be some decisions on how the version/its formatting is put together in future in order to allow for a good and comprehensible declarative configuration that's not a hack

RonnyPfannschmidt:
a extra source of confusion is that, there is different behavior between local version component formatting and general version formatting

whitequark:
Right, I see your point. I was actually thinking about refactoring that code and I have a potential approach in mind, but I thought that it might be preferable to do this in smaller pieces.

RonnyPfannschmidt:
@whitequark yes, its defintitively a chain of changes needed, key reason i decided not to touch it anytime soon is, that currently i am pretty stretched already and the feature is not something i personally need

whitequark:
Ack. I don't promise it but I may end up getting this done.

> basically setuptools_scm intentionally considers any working directory with a change on something the scm tracks on the way to a new version

Yes, I understand that. But what situation would have `distance=0` *and* `dirty=False`?

RonnyPfannschmidt:
checked out a tag, preparing to publish a release would have distance 0 and dirty=False

whitequark:
Ohh I see. Then I think custom schemes should probably be expanded to 5 conditions.

whitequark:
> checked out a tag, preparing to publish a release would have distance 0 and dirty=False

Are you sure? I've just tried it and when checking out `master` with `v0.3` pointing to `master` I get this:
```
cmd 'git describe --dirty --tags --long --match *[0-9]*'
out b'v0.3-0-g59ef6e6\n'
cmd 'git rev-parse --abbrev-ref HEAD'
out b'master\n'
tag v0.3
tag 'v0.3' parsed to {'version': 'v0.3', 'prefix': '', 'suffix': ''}
version pre parse v0.3
version <Version('0.3')>
version v0.3 -> 0.3
scm version <ScmVersion 0.3 d=None n=g59ef6e6 d=False b=master>
```

Meanwhile, when checking out `v0.3` itself I get this:
```
cmd 'git describe --dirty --tags --long --match *[0-9]*'
out b'v0.3-0-g59ef6e6\n'
cmd 'git rev-parse --abbrev-ref HEAD'
out b'HEAD\n'
tag v0.3
tag 'v0.3' parsed to {'version': 'v0.3', 'prefix': '', 'suffix': ''}
version pre parse v0.3
version <Version('0.3')>
version v0.3 -> 0.3
scm version <ScmVersion 0.3 d=None n=g59ef6e6 d=False b=HEAD>
```

RonnyPfannschmidt:
@whitequark I was sure, but you showed that that was wrong, I'll figure where I went wrong later today 
Thanks for putting it to the test

RonnyPfannschmidt:
the distance is only zero, when the working directory is dirty
```
[rpfannsc@ronny-t490-local pytest]$ SETUPTOOLS_SCM_DEBUG=1 python setup.py --version
WARNING: The wheel package is not available.
root '/home/rpfannsc/Projects/pytest-dev/pytest'
looking for ep setuptools_scm.parse_scm /home/rpfannsc/Projects/pytest-dev/pytest
found ep .git = setuptools_scm.git:parse
cmd 'git rev-parse --show-toplevel'
out b'/home/rpfannsc/Projects/pytest-dev/pytest\n'
real root /home/rpfannsc/Projects/pytest-dev/pytest
cmd 'git describe --dirty --tags --long --match *.*'
out b'6.1.2-0-g4e8b50ba0\n'
cmd 'git rev-parse --abbrev-ref HEAD'
out b'HEAD\n'
tag 6.1.2
tag '6.1.2' parsed to {'version': '6.1.2', 'prefix': '', 'suffix': ''}
version pre parse 6.1.2
version <Version('6.1.2')>
version 6.1.2 -> 6.1.2
scm version <ScmVersion 6.1.2 d=None n=g4e8b50ba0 d=False b=HEAD>
config {'version_scheme': 'guess-next-dev', 'local_scheme': 'node-and-date'}
ep ('setuptools_scm.version_scheme', 'guess-next-dev')
ep found: guess-next-dev
ep ('setuptools_scm.local_scheme', 'node-and-date')
ep found: node-and-date
version 6.1.2
local_version 
6.1.2
[rpfannsc@ronny-t490-local pytest]$ ls
AUTHORS  bench  changelog  CHANGELOG.rst  CITATION  codecov.yml  CODE_OF_CONDUCT.md  CONTRIBUTING.rst  doc  extra  LICENSE  OPENCOLLECTIVE.rst  pip-wheel-metadata  __pycache__  pyproject.toml  README.rst  RELEASING.rst  scripts  setup.cfg  setup.py  src  testing  TIDELIFT.rst  tox.ini
[rpfannsc@ronny-t490-local pytest]$ echo >> LICENSE 
[rpfannsc@ronny-t490-local pytest]$ SETUPTOOLS_SCM_DEBUG=1 python setup.py --version
WARNING: The wheel package is not available.
root '/home/rpfannsc/Projects/pytest-dev/pytest'
looking for ep setuptools_scm.parse_scm /home/rpfannsc/Projects/pytest-dev/pytest
found ep .git = setuptools_scm.git:parse
cmd 'git rev-parse --show-toplevel'
out b'/home/rpfannsc/Projects/pytest-dev/pytest\n'
real root /home/rpfannsc/Projects/pytest-dev/pytest
cmd 'git describe --dirty --tags --long --match *.*'
out b'6.1.2-0-g4e8b50ba0-dirty\n'
cmd 'git rev-parse --abbrev-ref HEAD'
out b'HEAD\n'
tag 6.1.2
tag '6.1.2' parsed to {'version': '6.1.2', 'prefix': '', 'suffix': ''}
version pre parse 6.1.2
version <Version('6.1.2')>
version 6.1.2 -> 6.1.2
scm version <ScmVersion 6.1.2 d=0 n=g4e8b50ba0 d=True b=HEAD>
config {'version_scheme': 'guess-next-dev', 'local_scheme': 'node-and-date'}
ep ('setuptools_scm.version_scheme', 'guess-next-dev')
ep found: guess-next-dev
ep ('setuptools_scm.local_scheme', 'node-and-date')
ep found: node-and-date
version 6.1.3.dev0
local_version +g4e8b50ba0.d20201128
6.1.3.dev0+g4e8b50ba0.d20201128
```

whitequark:
> the distance is only zero, when the working directory is dirty

I see. So the reason that happens is these lines:

https://github.com/pypa/setuptools_scm/blob/303defef04aa1a60dc74a4e52fef0bcf05c3d6be/src/setuptools_scm/version.py#L137-L138

However, would it not be clearer and more elegant to remove the distinction between `distance=0` and `distance=None` (idealy to the former), and rewrite these lines:

https://github.com/pypa/setuptools_scm/blob/303defef04aa1a60dc74a4e52fef0bcf05c3d6be/src/setuptools_scm/version.py#L159-L161

to say:

```python
    @property
    def exact(self):
        return self.distance == 0 and not self.dirty
```

What do you think?

Also, your finding means that the 4 custom scheme keys I came up will be enough.

RonnyPfannschmidt:
i believe its fair to say that the original names for the local scheme in #419 where made without the internals in mind (in all fairness, they arent documented that well

i believe the 3 cases to consider are `exact`, `clean` and `dirty` for formatting

currently guessing the next version is integrated deeply into the version scheme, i believe if the version guessing is separated from the version schemes, customization should be way more easy

a simple way to do that could be to extract guessing the next version out so that the format variable for the version number formatting turns into something that can either lazily or eagerly guess the next version

whitequark:
> i believe there need to be some decisions on how the version/its formatting is put together in future in order to allow for a good and comprehensible declarative configuration that's not a hack

Okay, I thought about it and I have several suggestions that should be technically viable. In order of increasing invasiveness:
  1. Remove `custom_version_scheme`, sidestepping that problem entirely.
  2. Make the version object itself responsible for applying the customization so that it can change the format string *after* the guess is made. I would probably do this by inheriting from `ScmVersion` and overriding `format_with` to call the superclass with the custom `fmt` string. This would require almost no other changes, except that the "exact" path of `simplified_semver_version` would have to be changed to return `return version.format_next_version(guess_next_simple_semver, retain=SEMVER_LEN, increment=False)`.
  3. Split guessing and formatting into distinct steps. This way, there would be three version schemes (`guess-next-dev`, `post-release`, and `no-guess-dev`), and three guess algorithms (`non-semver`, `python-simplified-semver`, and `release-branch-semver`). In this scheme, `guessed` would become a field of `ScmVersion` (or perhaps a subclass), filled in before any scheme rendering code. The `guess-next-dev` scheme would always return `version.format_with("{guessed}")` (because `simplified_semver_version` may return a different version for the exact case than the tag, both paths have to go through guessing).

I'm happy to do (1) or (2). (It is unfortunate that (1) won't make all of my configurations declarative, but I could live with that.)

I'm generally on board with (3) or a variation on that topic, but you would need to decide how to deprecate the configuration of the guess algorithm through the version scheme entry point. I'm also not sure if the churn is worth it.

whitequark:
Ahh, you replied before I had a chance to post the answer I was writing. But it should be still relevant.

RonnyPfannschmidt:
My aim would be a eventual 3, with enabling 2 in the interim

This weekend is a bit packed with personal /family activities, so unfortunately I won't be able to make much effort for the preps

I'm under the impression, that it's possible to implement the full version scheming without too much effort I'll try to verify that idea this weekend

RonnyPfannschmidt:
i took a initial look at the refactoring required, and it seems way larger than i hoped it would be
i'll have to dedicate another time-frame to see if/how it would make sense

whitequark:
Thanks for looking into it!

RonnyPfannschmidt:
im closing this one to trim down the queue

i still haven't gotten to work on the details, but i suspect it will look very different than initially imagined (as im learning more about the details)

again than you for tying this out and giving valuable input

PR #504, opened 2020-12-11
Require setuptools >=42 in pyproject.toml

tiran:
I noticed that ``test_pyproject_support`` and ``test_pyproject_support_with_git`` are broken with setuptools 41.6 on Fedora 32. Tests and feature are working correctly with setuptools 42 and newer. ``tox.ini`` already requires setuptools >= 42.


RonnyPfannschmidt (approved):
Thanks! , will be part of a release this weekend

Issue #507, opened 2020-12-13
distribution name is not passed from pyproject.toml support


RonnyPfannschmidt:
it simply needs to be passed to the configuration

fallout of pytest-dev/pytest#8133
required for  pytest-dev/pytest#8134

PR #508, opened 2020-12-13
Wip reshape trace

RonnyPfannschmidt:
while fixing #507 i revamped trace to help me, but i realized this may be a breaking change
so i decided to put it into a wip pr for later pickup


RonnyPfannschmidt:
closing as unwanted
will either go back to logging to migrate to something like eliot

PR #514, opened 2021-01-20
Use slimmer and more modern APIs

### LINKED ISSUE #513: Port `iter_matching_entrypoints` to `importlib.metadata`
When calling `_version_from_entrypoints` → [`iter_matching_entrypoints`](https://github.com/pypa/setuptools_scm/blob/74b17470912c83e658cfd18e66eec654be87fcd7/src/setuptools_scm/discover.py), we should use `importlib.metadata` instead of `pkg_resources`, as `pkg_resources` isn’t the standard way to do this, and is also buggy in this context (see pypa/setuptools#2531)

The changes required would be pretty minimal:

requirements:
```py
importlib_metadata; python_version < "3.8"
```

code:
```py
try:
    from importlib.metadata import entry_points
except ImportError:
    from importlib_metadata import entry_points
```

flying-sheep:
So `setuptools` actually depends on `packaging` in newer versions, so this uses `importlib.metadata` and `packaging` on 3.8+ and `setuptools`’ `pkg_resources` otherwise.

@RonnyPfannschmidt: Are the test failures because of an upstream change? AFAIK `pkg_resources` just reexports code identical to `packaging.version.parse`, so I don’t think this DeprecationWarning wouldn’t happen in the code before my changes.

Fixes #513


src/setuptools_scm/version.py
```diff
@@ -6,11 +6,14 @@
 import os
 
 from .config import Configuration
-from .utils import trace, string_types
+from .utils import trace, string_types, iter_entry_points
 
-from pkg_resources import iter_entry_points
+try:
+    # Newer setuptools/pkg_resources releases just use that
+    from packaging.version import parse as pkg_parse_version
```
RonnyPfannschmidt: setuptools uses vendoring of a version thats not broken

setup.cfg
```diff
@@ -31,7 +31,8 @@ classifiers=
 zip_safe = true
 python_requires= >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
 install_requires=
-    setuptools
+    setuptools; python_version < "3.8"
+    packaging; python_version >= "3.8"
```
RonnyPfannschmidt: since  the intended use-case is with setuptools as of now, i dont see any demonstrated reason to bring in packaging, as far as i can tell all it does is break things

before packaging can/should be considered, there should be a tightening of the supported setuptools version in install_requires

src/setuptools_scm/version.py
flying-sheep: How is it broken? Setuptools currently use 20.4, the newest version is 20.8

Can you please give me an actionable suggestion for improving this PR?

setup.cfg
flying-sheep: So what do you suggest? Pinning the `packaging` version to what `setuptools` uses? (20.4)

RonnyPfannschmidt: Basically until setuptools drops the vendoring we should not add packaging as dependency

flying-sheep: So what do you suggest?

RonnyPfannschmidt: Migration of the version class specification from the module to passing a callable to the configuration

Then the setuptools integration can pass it in and other new integrations could pass the packaging on sooner

RonnyPfannschmidt: I'll provide a better example when I get back home

RonnyPfannschmidt:
i'll revisit this after the coming release of setuptools_scm 6 which drops python2

flying-sheep:
good to hear!

RonnyPfannschmidt:
Looking good, I'll take a look at merge when I get back to the computer

RonnyPfannschmidt:
thanks!

PR #519, opened 2021-02-01
DOC Mention slow import times for pkg_resources 

rth:
For retrieving the package version at runtime, the readme makes it sound as though `importlib_metadata` and `pkg_resources` are equivalent. Setuptools is almost always installed, so it's not that different from being in stdlib.

pkg_resources is however fairly slow to import https://github.com/pypa/setuptools/issues/926 and should not be recommended for this use case. This adds a note about pkg_resources import times.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -220,7 +220,8 @@ Alternatively, you can use ``pkg_resources`` which is included in
         # package is not installed
        pass
 
-This does place a runtime dependency on ``setuptools``.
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt (approved):

ssbarnea (approved):

RonnyPfannschmidt:
merging as the errors are from the new setup-tools that breaks easy_install off

PR #520, opened 2021-02-08
Use show prefix instead of show toplevel for finding

### LINKED ISSUE #415: Detect POSIX-style paths on Windows
I have a `git` binary installed on Windows (the MSYS2 variant, specifically) that does not understand Windows-style paths. When trying to install software using `setuptools_scm`, this leads to cryptic errors such as the following:

```
William@DESKTOP-H0PMN4M MINGW64 ~/Projects/FPGA/nmigen
$ python setup.py develop
  File "setup.py", line 10, in <module>
    setup(
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "C:/msys64/mingw64/lib/python3.8/distutils/core.py", line 108, in setup
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/dist.py", line 447, in __init__
    _Distribution.__init__(self, {
  File "C:/msys64/mingw64/lib/python3.8/distutils/dist.py", line 292, in __init__
    self.finalize_options()
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/dist.py", line 740, in finalize_options
    ep.load()(self)
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/dist.py", line 747, in _finalize_setup_keywords
  File "c:/msys64/home/william/projects/fpga/nmigen/.eggs/setuptools_scm-3.5.0-py3.8.egg/setuptools_scm/integration.py", line 17, in version_keyword
    dist.metadata.version = _get_version(config)
  File "c:/msys64/home/william/projects/fpga/nmigen/.eggs/setuptools_scm-3.5.0-py3.8.egg/setuptools_scm/__init__.py", line 147, in _get_version
    parsed_version = _do_parse(config)
  File "c:/msys64/home/william/projects/fpga/nmigen/.eggs/setuptools_scm-3.5.0-py3.8.egg/setuptools_scm/__init__.py", line 103, in _do_parse
    version = _version_from_entrypoints(config) or _version_from_entrypoints(
  File "c:/msys64/home/william/projects/fpga/nmigen/.eggs/setuptools_scm-3.5.0-py3.8.egg/setuptools_scm/__init__.py", line 63, in _version_from_entrypoints
    version = _call_entrypoint_fn(root, config, ep.load())
  File "c:/msys64/home/william/projects/fpga/nmigen/.eggs/setuptools_scm-3.5.0-py3.8.egg/setuptools_scm/__init__.py", line 44, in _call_entrypoint_fn
    return fn(root, config=config)
  File "c:/msys64/home/william/projects/fpga/nmigen/.eggs/setuptools_scm-3.5.0-py3.8.egg/setuptools_scm/git.py", line 98, in parse
    wd = GitWorkdir.from_potential_worktree(config.absolute_root)
  File "c:/msys64/home/william/projects/fpga/nmigen/.eggs/setuptools_scm-3.5.0-py3.8.egg/setuptools_scm/git.py", line 33, in from_potential_worktree
    if not samefile(real_wd, wd):
  File "C:\msys64\mingw64/lib/python3.8/genericpath.py", line 100, in samefile
    s1 = os.stat(f1)
FileNotFoundError: [WinError 3] The system cannot find the path specified: '/home/William/Projects/FPGA/nmigen'
```

The error comes from the path `git` returns from the `rev-parse` command:

```
William@DESKTOP-H0PMN4M MINGW64 ~/Projects/FPGA/nmigen
$ git rev-parse --show-toplevel
/home/William/Projects/FPGA/nmigen
```

One potential solution is to pipe the output of `git rev-parse` through `cygpath` is we detect that the `git` being run is MSYS2 (or otherwise doesn't understand Windows-style paths), as unfortunately conversion between POSIX and Windows path is an [involved](https://github.com/Alexpux/Cygwin/blob/master/winsup/utils/cygpath.cc) problem:

```
William@DESKTOP-H0PMN4M MINGW64 ~/Projects/FPGA/nmigen
$ git rev-parse --show-toplevel | cygpath -w -f -
C:\msys64\home\William\Projects\FPGA\nmigen
```

The following workaround allows me to successfully run the `egg-info`, `develop`, and `install` successfully (with `warning: no previously-included files found matching '.travis.yaml'`). Obviously this patch isn't acceptable as-is, and this may not be the correct repo to bring this issue up, depending on the scope of supported `git` variants. However, I figure it's worth a shot.

```
William@DESKTOP-H0PMN4M MINGW64 ~/Projects/py-packages/setuptools_scm
$ git diff
diff --git a/src/setuptools_scm/file_finder_git.py b/src/setuptools_scm/file_finder_git.py
index 8b81d2c..f68c91a 100644
--- a/src/setuptools_scm/file_finder_git.py
+++ b/src/setuptools_scm/file_finder_git.py
@@ -17,6 +17,13 @@ def _git_toplevel(path):
                 universal_newlines=True,
                 stderr=devnull,
             )
+
+            out = subprocess.check_output(
+                ["cygpath", "-w", out],
+                cwd=(path or "."),
+                universal_newlines=True,
+                stderr=devnull,
+            )
         trace("find files toplevel", out)
         return os.path.normcase(os.path.realpath(out.strip()))
     except subprocess.CalledProcessError:
diff --git a/src/setuptools_scm/git.py b/src/setuptools_scm/git.py
index afefa34..43cc6c5 100644
--- a/src/setuptools_scm/git.py
+++ b/src/setuptools_scm/git.py
@@ -27,6 +27,9 @@ class GitWorkdir(object):
     @classmethod
     def from_potential_worktree(cls, wd):
         real_wd, _, ret = do_ex("git rev-parse --show-toplevel", wd)
+        if ret:
+            return
+        real_wd, _, ret = do_ex("cygpath -w {}".format(real_wd), wd)
         if ret:
             return
         trace("real root", real_wd)

William@DESKTOP-H0PMN4M MINGW64 ~/Projects/py-packages/setuptools_scm
$
```

naveen521kk:
This should fix Cygwin type git which will return POSIX path
when `git rev-parse --show-toplevel` is called which isn't parsed
properly. This call `git rev-parse --show-prefix` which return
relative paths which is later parsed.

Fixes https://github.com/pypa/setuptools_scm/issues/415

This idea came from a patch in msys2, https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-setuptools-scm/0001-git-Use-show-prefix-instead-of-show-toplevel-for-fin.patch, and ported to python2.

### DIFF
--- a/src/setuptools_scm/file_finder_git.py
+++ b/src/setuptools_scm/file_finder_git.py
@@ -11,13 +11,35 @@ log = logging.getLogger(__name__)
 
 def _git_toplevel(path):
     try:
+        cwd = os.path.abspath(path or ".")
         with open(os.devnull, "wb") as devnull:
             out = subprocess.check_output(
-                ["git", "rev-parse", "--show-toplevel"],
-                cwd=(path or "."),
+                ["git", "rev-parse", "--show-prefix"],
+                cwd=cwd,
                 universal_newlines=True,
                 stderr=devnull,
             )
+        out = out.strip()
+        if not out:
+            out = cwd
+        else:
+            cwd_parents = []
+            out_parents = ["."]
+            while True:
+                if os.path.abspath(os.path.join(cwd, os.pardir)) != cwd:
+                    cwd = os.path.abspath(os.path.join(cwd, os.pardir))
+                    cwd_parents.append(cwd)
+                else:
+                    break
+            while True:
+                if os.path.basename(out) != "":
+                    out = os.path.join(out, os.pardir)
+                    out_parents.append(out)
+                else:
+                    break
+            print(cwd_parents)
+            print(out_parents)
+            out = str(cwd_parents[len(out_parents) - 1])
         trace("find files toplevel", out)
         return os.path.normcase(os.path.realpath(out.strip()))
     except subprocess.CalledProcessError:
--- a/src/setuptools_scm/file_finder_git.py
+++ b/src/setuptools_scm/file_finder_git.py
@@ -19,7 +19,7 @@ def _git_toplevel(path):
                 universal_newlines=True,
                 stderr=devnull,
             )
-        out = out.strip()[:-1] #remove the trailing pathsep
+        out = out.strip()[:-1]  # remove the trailing pathsep
         if not out:
             out = cwd
         else:
@@ -27,7 +27,7 @@ def _git_toplevel(path):
             # ``cwd`` is absolute path to current working directory.
             # the below method removes the length of ``out`` from
             # ``cwd``, which gives the git toplevel
-            out = cwd[:len(cwd)-len(out)-1]
+            out = cwd[: len(cwd) - len(out) - 1]
         trace("find files toplevel", out)
         return os.path.normcase(os.path.realpath(out.strip()))
     except subprocess.CalledProcessError:
--- a/src/setuptools_scm/file_finder_git.py
+++ b/src/setuptools_scm/file_finder_git.py
@@ -29,7 +29,8 @@ def _git_toplevel(path):
             # ``cwd``, which gives the git toplevel
             assert cwd.replace("\\", "/").endswith(out)
             # In windows cwd contains ``\`` which should be replaced by ``/``
-            # for this assertion to work.
+            # for this assertion to work. Length of string isn't changed by replace
+            # ``\\`` is just and escape for `\`
             out = cwd[: -len(out)]
         trace("find files toplevel", out)
         return os.path.normcase(os.path.realpath(out.strip()))
--- a/src/setuptools_scm/file_finder_git.py
+++ b/src/setuptools_scm/file_finder_git.py
@@ -19,25 +19,15 @@ def _git_toplevel(path):
                 universal_newlines=True,
                 stderr=devnull,
             )
-        out = out.strip()
+        out = out.strip()[:-1] #remove the trailing pathsep
         if not out:
             out = cwd
         else:
-            cwd_parents = []
-            out_parents = ["."]
-            while True:
-                if os.path.abspath(os.path.join(cwd, os.pardir)) != cwd:
-                    cwd = os.path.abspath(os.path.join(cwd, os.pardir))
-                    cwd_parents.append(cwd)
-                else:
-                    break
-            while True:
-                if os.path.basename(out) != "":
-                    out = os.path.join(out, os.pardir)
-                    out_parents.append(out)
-                else:
-                    break
-            out = str(cwd_parents[len(out_parents) - 1])
+            # Here, ``out`` is a relative path to root of git.
+            # ``cwd`` is absolute path to current working directory.
+            # the below method removes the length of ``out`` from
+            # ``cwd``, which gives the git toplevel
+            out = cwd[:len(cwd)-len(out)-1]
         trace("find files toplevel", out)
         return os.path.normcase(os.path.realpath(out.strip()))
     except subprocess.CalledProcessError:
--- a/src/setuptools_scm/file_finder_git.py
+++ b/src/setuptools_scm/file_finder_git.py
@@ -27,7 +27,7 @@ def _git_toplevel(path):
             # ``cwd`` is absolute path to current working directory.
             # the below method removes the length of ``out`` from
             # ``cwd``, which gives the git toplevel
-            assert cwd[len(cwd) - len(out) :].replace("\\", "/") == out
+            assert cwd.replace("\\", "/").endswith(out)
             # In windows cwd contains ``\`` which should be replaced by ``/``
             # for this assertion to work.
             out = cwd[: len(cwd) - len(out) - 1]
--- a/src/setuptools_scm/file_finder_git.py
+++ b/src/setuptools_scm/file_finder_git.py
@@ -37,8 +37,6 @@ def _git_toplevel(path):
                     out_parents.append(out)
                 else:
                     break
-            print(cwd_parents)
-            print(out_parents)
             out = str(cwd_parents[len(out_parents) - 1])
         trace("find files toplevel", out)
         return os.path.normcase(os.path.realpath(out.strip()))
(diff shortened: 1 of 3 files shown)


naveen521kk:
I think `easy_install` is removed from setuptools, so that failure seems irrelevant. If interested I can set up Github Actions to test on msys2-git also.

RonnyPfannschmidt:
an action that tests it would be most appreciated,, im absolutely non-fluent with the windows details and dont want to break them

RonnyPfannschmidt:
self-install needs a setuptools version pin, i'll merge if it fails anyway

naveen521kk:
sure I will create an action to test on msys2 soon.

src/setuptools_scm/file_finder_git.py
```diff
@@ -11,13 +11,33 @@
...
                 universal_newlines=True,
                 stderr=devnull,
             )
+        out = out.strip()
+        if not out:
+            out = cwd
+        else:
+            cwd_parents = []
+            out_parents = ["."]
+            while True:
```
RonnyPfannschmidt: after crosschecking with how this is handled,
i believe this can be significantly simplified

presuming the prefix is a relative path, talking the length of the prefix off the full path should yield a path ending with a path separator

the pathlib using code was a bit unfortunate

naveen521kk: I just try to exactly what that patch tried to do. Let me check for this soon.

naveen521kk: Done in https://github.com/pypa/setuptools_scm/pull/520/commits/f383bd83ea57a15e69119278a4c4df786a4a9ed3

also, the tests pass, thanks for this idea :)

src/setuptools_scm/git.py
```diff
@@ -26,9 +27,15 @@ def do_ex(self, cmd):
...
-        real_wd, _, ret = do_ex("git rev-parse --show-toplevel", wd)
+        wd = os.path.abspath(wd)
+        real_wd, _, ret = do_ex("git rev-parse --show-prefix", wd)
+        real_wd = real_wd[:-1]  # remove the trailing pathsep
         if ret:
             return
+        if not real_wd:
+            real_wd = wd
+        else:
+            real_wd = wd[: len(wd) - len(real_wd) - 1]
```
RonnyPfannschmidt: lovely trimdown
please assert the real_wd is actually the tail of the path for pedantic/paranoya

naveen521kk: Fixed in https://github.com/pypa/setuptools_scm/pull/520/commits/b9cdb54319c2f51548dbebfbbab50fbe09406d84

src/setuptools_scm/file_finder_git.py
```diff
@@ -27,6 +27,9 @@ def _git_toplevel(path):
             # ``cwd`` is absolute path to current working directory.
             # the below method removes the length of ``out`` from
             # ``cwd``, which gives the git toplevel
+            assert cwd[len(cwd) - len(out):].replace('\\','/') == out
```
RonnyPfannschmidt: please use the `str.endswith` method for better readability

also please take a look at https://www.python.org/dev/peps/pep-0616/#specification for a simpler expression of the suffix dropping

naveen521kk: > please use the str.endswith method for better readability

:+1: 

> also please take a look at https://www.python.org/dev/peps/pep-0616/#specification for a simpler expression of the suffix dropping

ah! negative indexing, let me do, forgot that they do exist... thanks

naveen521kk: Done https://github.com/pypa/setuptools_scm/pull/520/commits/c4d06b0c82b6f85233c715ff262d52bc3aa8bdf7 https://github.com/pypa/setuptools_scm/pull/520/commits/eae87b09a67699d10f63fceda3c47e8f371e8a1d

src/setuptools_scm/git.py
```diff
@@ -26,9 +27,18 @@ def do_ex(self, cmd):
...
-        real_wd, _, ret = do_ex("git rev-parse --show-toplevel", wd)
+        wd = os.path.abspath(wd)
+        real_wd, _, ret = do_ex("git rev-parse --show-prefix", wd)
+        real_wd = real_wd[:-1]  # remove the trailing pathsep
         if ret:
             return
+        if not real_wd:
+            real_wd = wd
+        else:
+            assert wd.replace("\\", "/").endswith(real_wd)
```
RonnyPfannschmidt, line 38: for nitpicking maybe lets add a comment that the length is not changed by the replace,
as the escape always trips me up, alternatively maybe a raw string could also help

also great job, this is merge-able as is, do you want to squish some commits?

RonnyPfannschmidt (approved):

src/setuptools_scm/git.py
naveen521kk: > alternatively maybe a raw string could also help

ok then I will use raw strings

> also great job, this is merge-able as is, do you want to squish some commits?

maybe squash while merge?

naveen521kk: no raw string doesn't help, because it raises `SyntaxError: EOL while scanning string literal`

naveen521kk: let me add a comment.

RonnyPfannschmidt: i  can do that if you are fine with it (my personal stance is that  i only change history created by others if they request it or consent to it )

i'll enable squash merges for this repo now ^^

naveen521kk: done https://github.com/pypa/setuptools_scm/pull/520/commits/92c5422207df8b376232ace6d311575d75feb415

naveen521kk: > i can do that if you are fine with it (my personal stance is that i only change history created by others if they request it or consent to it )

It's ok I don't mind, it could some problems if squash it myself, rather than GitHub squashing it.

RonnyPfannschmidt:
thanks again!

lazka:
Thanks for supporting MSYS2!

naveen521kk:
@RonnyPfannschmidt is it possible to expect a release soon? It breaks pretty much any project which uses this and has a `pyproject.toml`.
Thanks.

RonnyPfannschmidt:
Thanks for the ping, I'll sort the changelog this evening and push it out

PR #526, opened 2021-03-04
Prepare breakages - drop old setuptools/python

RonnyPfannschmidt:


setup.cfg
```diff
@@ -15,12 +15,10 @@ classifiers=
     Intended Audience :: Developers
     License :: OSI Approved :: MIT License
     Programming Language :: Python
-    Programming Language :: Python :: 2.7
-    Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.5
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
```
jugmac00, line 21: There is also a "Python 3 only" classifier, just in case.

`Programming Language :: Python :: 3 :: Only`

henryiii: setup-cfg-fmt (via pre-commit, generally) is useful for this sort of cleanup, it puts the only classify in for you if applicable. :)

henryiii:
Why were eggs built in the first place? Pip hasn't even been able to install them for (some rather unspecified long period of time). Good to see them go! :)

RonnyPfannschmidt:
@henryiii for easy_install, it still worked in a number of legacy setups

But it's gone now and setuptools can do wheels

PR #529, opened 2021-03-11
CalVer by date

ikerreyes:
Implementation of CalendarVersioning (<year>.<month>.<day>.<patch>). It should support years with 4 and 2 digits. Added as a version_scheme, it will:

- if there is not exact match, use current date. If already used, add +1 to the <patch>
- if there is exact match, leave as it is
- distance is kept as default when formatting (.devX)


testing/test_version.py
```diff
@@ -170,3 +173,68 @@ def test_version_bump_bad():
     ):
 
         guess_next_version(tag_version="2.0.0-alpha.5-PMC")
+
+
+def date_to_str(date_=None, days_offset=0, fmt="%y.%-m.%-d"):
+    date_ = date_ or date.today()
+    date_ = date_ - timedelta(days=days_offset)
+    return date_.strftime("%y.%-m.%-d")
```
RonnyPfannschmidt:
```suggestion
    return format(date_, fmt)
```

src/setuptools_scm/version.py
```diff
@@ -309,6 +309,69 @@ def no_guess_dev_version(version):
...
+        (
+            r"^(?P<date>(?P<year>\d{2}|\d{4})(?:\.\d{1,2}){2})"
+            r"(?:\.(?P<patch>\d*)){0,1}?$"
+        ),
+        str(ver),
+    )
+    return match
+
+
+def guess_next_date_ver(version, node_date=None):
```
RonnyPfannschmidt: version should contain the date, so i think its fine to use `version.node_date or date.today()` down there

src/setuptools_scm/version.py
```diff
@@ -309,6 +309,69 @@ def no_guess_dev_version(version):
...
+    match = date_ver_match(version)
+    if match is None:
+        raise ValueError(
+            "{version} does not correspond to a valid versioning date, "
+            "please correct or use a custom version scheme".format(version=version)
+        )
+    # keep same formatting style
+    # "%y.%-m.%-d" will not work on windows:
+    # https://stackoverflow.com/questions/904928/python-strftime-date-without-leading-0
+    fmt = "{{dt:%{year_fmt}}}.{{dt.month}}.{{dt.day}}".format(
```
RonnyPfannschmidt: with

```
t = datetime.date.today()
"{node_date:{date_fmt}}.{patch}".format(node_date=t, date_fmt="%y.%m.%d", patch=3, distance=213)
```
should get things really nice

the `date_fmt` should also be a part of the default parameters for the function to ensure its overridable

src/setuptools_scm/version.py
ikerreyes: I think it doesn't because the `format_next_version` function only passes the version tag, rather than the full version object. Maybe I should be using `format_next_version` to call `guess_next_date_ver`?

RonnyPfannschmidt: ah yes, please use that, format_next_version was made exactly for that use-case

src/setuptools_scm/version.py
ikerreyes: should then the check to find whether to use 4 or 2 digit years computed before calling `guess_next_date_ver`?

RonnyPfannschmidt: Yes

PR #533, opened 2021-03-16
chore: bump pyupgrade to  Python 3.6+

henryiii:
~~If you use pre-commit.ci, this should automatically add the changes (if there are any),~~ No, so I ran it manually.

Edit: black should come after pyupgrade.


RonnyPfannschmidt:
thanks for sorting that trough

PR #546, opened 2021-03-19
Fix rst formatting issue in README

wbolster:


RonnyPfannschmidt:
thanks for the contribution, i just took note that his part of the docs needs to be gone, it no longer applies in recent setuptools_scm

thanks for bringing my oversight to my attention

wbolster:
even better 😂

RonnyPfannschmidt:
removed in #550

Issue #524, opened 2021-02-24
Incorrect removal of dash in case of pre-release tag string ?


smarie:
I found this today: `1.0.0-rc1` becomes `1.0.0rc1`, see details here https://github.com/smarie/python-getversion/issues/10

The underlying error is due to `pkg_resources` that removes the pre-release dash automatically in its string representation since version [0.6a9](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#0.6a9).

However as suggested in [my conclusions](https://github.com/smarie/python-getversion/issues/10#issuecomment-784987499) I think that the issue should probably better be fixed in the default scheme of setuptools_scm because faithfulness to git tags and compliance with `semver` for example, is probably a higher concern for you than for `pkg_resources`.

Thanks again for this great package !

RonnyPfannschmidt:
At first glance i think setuptools scm is correct as it handles standard compliant python version numbers

Those are not in line with semver

RonnyPfannschmidt:
After reading your conclusion I'll review the details and either fix this in a major release or document the choice

smarie:
Great ! Thanks for your reactivity @RonnyPfannschmidt . In the meantime I'll ship `getversion` with a patch using a custom version scheme

RonnyPfannschmidt:
https://www.python.org/dev/peps/pep-0440/#pre-release-separators

I believe that a decision on whether or not to normalise should be made

Im leaning towards always normalising


@jaraco any opinion on the way

smarie:
Note that a conservative move would be that you include a patched version scheme in the "alternate but built-in" version schemes of setuptool_scm : see [my current dirty hack](https://github.com/smarie/python-getversion/blob/ad3f31017bb58c76398010690f44bb0cdde7a43b/getversion/plugin_setuptools_scm.py#L35)

jaraco:
In general, I try to avoid normalizing user inputs, to honor the user's indication, and only to normalize in storage or comparison. However, the consensus from the Python community seems to be to prefer normalization early, to essentially disregard the user's intention and replace it with a normal form.

It does seem to me to be a gap that the most popular versioning scheme (semver) is in conflict with the packaging normalization rules.

I like the idea of providing an "alternate but built-in" version that's lenient to semver syntax. That would provide an opportunity to explore an implementation and to give users a nice escape hatch. I'd probably not recommend this except for very popular and compatible forms like semver.

smarie:
[...]

PR #555, opened 2021-04-11
Add condition to skip version parser

abitrolly:
Ref. #524

- [x] add config.skip_version_parser
- [x] add condition to skip parsing
- [ ] test that it works

### DIFF
--- a/src/setuptools_scm/version.py
+++ b/src/setuptools_scm/version.py
@@ -188,7 +188,10 @@ def meta(
             "meta invoked without explicit configuration,"
             " will use defaults where required."
         )
-    parsed_version = _parse_tag(tag, preformatted, config)
+    if config.skip_version_parser:
+        parsed_version = tag
+    else:
+        parsed_version = _parse_tag(tag, preformatted, config)
     trace("version", tag, "->", parsed_version)
     assert parsed_version is not None, "Can't parse version %s" % tag
     return ScmVersion(
--- a/src/setuptools_scm/config.py
+++ b/src/setuptools_scm/config.py
@@ -65,6 +65,7 @@ class Configuration:
         parse=None,
         git_describe_command=None,
         dist_name=None,
+        skip_version_parser=False
     ):
         # TODO:
         self._relative_to = relative_to
@@ -82,6 +83,7 @@ class Configuration:
         self.tag_regex = tag_regex
         self.git_describe_command = git_describe_command
         self.dist_name = dist_name
+        self.skip_version_parser = skip_version_parser
 
     @property
     def fallback_root(self):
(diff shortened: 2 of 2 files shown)


src/setuptools_scm/version.py
```diff
@@ -188,7 +188,10 @@ def meta(
             "meta invoked without explicit configuration,"
             " will use defaults where required."
         )
-    parsed_version = _parse_tag(tag, preformatted, config)
+    if config.skip_version_parser:
```
RonnyPfannschmidt, line 191: the issue already outlined a acceptable structural solution, this isn't

a random flag to opt out not just of the normalization of the version , but also all other cleanup is not going to cut it
also integration tests are required for new features

i strongly suggest to write tests and to consider a Version subclass that normalizes less

RonnyPfannschmidt (changes requested):

src/setuptools_scm/version.py
abitrolly: I don't quite get the reference. Can you explain what's wrong with the flag?

RonnyPfannschmidt: as implemented and placed it takes out a number of features in surprising ways

in the issue before it was discussed to have a Version subclass that prevents normalization of valid versions and enables broader version usage

abitrolly: I don't understand you. This is an optional opt-out from the check. The role of SCM plugin in my pipeline is to pass the version from tagged commit to setuptools. Choosing which tag is suitable for release is done on CI level.

RonnyPfannschmidt: Your specific use case is not the main goal of setuptools_scm

I stated a way that would allow you to implement your use case in a way that matches the project

Im not going to take ownership of the stopgap solution you propose

Currently Im very thin spread  on time and can't discus /fight over this

abitrolly: Well, true, its your project and your time. It is not me who should stop you from having fun. )

For my project I switched to hardcoded version in the .py file and `attr:` tag. It will take less time for managing this manually than "fighting" for this change. When I get bored, I will inject the tag into the build process in my CI.

Thanks for spending the time with me though. I now close this PR as it doesn't look like a fit for this project.

abitrolly: Reopening so that other people could weight in their opinion.

RonnyPfannschmidt:
closing as there is work on a correct implementation in #584

abitrolly:
That's good to know. I am looking for a way how to run `git describe --contains HEAD` to extract version without using any plugins. So far I haven't found if `pyproject.toml` supports that function.

Issue #222, opened 2018-03-14
create a command/function that prints the next version


RonnyPfannschmidt:
followup to #124

PR #560, opened 2021-04-23
Add command to print next version

jacobolofsson:
Fix for #222

Adds a "increment" version scheme, that guesses the next version, and
does not take distance in to account (no devN postfix).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,6 +61,7 @@ setuptools_scm.parse_scm_fallback =
     setup.py = setuptools_scm.hacks:fallback_version
 setuptools_scm.version_scheme =
     guess-next-dev = setuptools_scm.version:guess_next_dev_version
+    increment = setuptools_scm.version:increment_version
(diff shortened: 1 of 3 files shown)


RonnyPfannschmidt (changes requested):
thanks for trying to tackle this issue,

unfortunately the solution is at the wrong place

the proposed scheme doesn't take into account other version schemes and is not available from other invocations/scemes in a practical manner

instead of creating a new version scheme, the returned version should drop the dev/local part in some way (in case of a clean workdir that is)

ideally this should be available as setup command as well as a opt-in for the get_version call (which in turn can be argument-ed with a env var later on)

RonnyPfannschmidt:
closing due to lack of feedback

jacobolofsson:
The next time I have some free time on my hands I might take a crack at this again, but ill reopen it when(if?) I get there.

RonnyPfannschmidt:
@jacobolofsson thanks for the update

PR #573, opened 2021-06-01
Accept branches start with 'v' as being a release branch.

### LINKED ISSUE #571: Treat v-prefixed branch as release for release-branch-semver?
The `release-branch-semver` scheme appears to match what we do in Matplotlib pretty closely. Since we make tags like `v1,2,3`, our branches are named like `v1.2.x`. This unfortunately is not treated as a release branch because of the `v` prefix in the branch name, even though the `v` prefix is dropped from the tag.

Now, we've never made a release branch with `setuptools-scm`, so this is something we could change, but it'd require updating a bunch like docs, CI, etc.

Does this make sense to handle this prefixing in the `release-branch-semver`?

QuLogic:
This seems to be the most straightforward way to implement it, as it matches the `VERSION_REGEX` in `packaging` that just has `v?{other stuff}`.
 
Fixes #571.

### DIFF
--- a/src/setuptools_scm/version.py
+++ b/src/setuptools_scm/version.py
@@ -275,11 +275,15 @@ def release_branch_semver_version(version):
         # Does the branch name (stripped of namespace) parse as a version?
         branch_ver = _parse_version_tag(version.branch.split("/")[-1], version.config)
         if branch_ver is not None:
+            branch_ver = branch_ver["version"]
+            if branch_ver[0] == "v":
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (approved):

RonnyPfannschmidt (approved):
well done, thanks

RonnyPfannschmidt:
yikes, sorry i let that one sit

PR #576, opened 2021-06-01
docs: recommend _version.py as a nicer default

### LINKED ISSUE #575: docs: recommend _version.py
The example shows `<package>/version.py`. I think it be a little better to have the example `<package>/_version.py`. Then users could do `from ._version import version as __version__`, or whatever, and not be affected by changes to the output template as much.

I'm asking since I'd normalized on using version.py everywhere ([Scikit-HEP developer docs](https://scikit-hep.org/developer/packaging#git-tags-official-pypa-method), [scikit-hep/cookie](https://github.com/scikit-hep/cookie), and lots of our packages) since it was the way the example had it, but now another developer is unhappy that it's public, and I agree with him.

henryiii:
This renames the recommendation to `_version.py`, to avoid publishing the contents of "version.py" as externally visible API. Ideally, one should do `from ._version import version as __version__` to follow common pratice; I could add that as an example if you'd like it. But, regardless, I think this is slightly nicer as a recommendation.

Closes #575.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,7 +70,7 @@ to be supplied to ``get_version()``. For example:
     # pyproject.toml
 
     [tool.setuptools_scm]
-    write_to = "pkg/version.py"
+    write_to = "pkg/_version.py"
 
 
 ``setup.py`` usage (deprecated)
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
Thanks!

PR #584, opened 2021-06-17
Support for custom and non-normalizing version classes

### LINKED ISSUE #524: Incorrect removal of dash in case of pre-release tag string ?

smarie:
Fixes #524


smarie:
Very rough because I have to go, but already a first working step. Let's take this again tomorrow.

Remaining todo:

 - documentation
 - one test explicitly for git release candidates ("1.0.1-rc1")
 - a utility class `NonNormalizedVersion`

smarie:
All set apart from flake8/black possible remaining errors

src/setuptools_scm/version.py
```diff
@@ -20,6 +20,38 @@
...
+    """
+
+    def __init__(self, tag):
+        self.tag = tag
+
+    def __repr__(self):
+        return self.tag
+
+
+def _get_version_cls(config):
```
RonnyPfannschmidt: this has to go

src/setuptools_scm/config.py
```diff
@@ -82,6 +84,8 @@ def __init__(
         self.tag_regex = tag_regex
         self.git_describe_command = git_describe_command
         self.dist_name = dist_name
+        self.version_cls = version_cls
```
RonnyPfannschmidt: this should fill in a actual class at the config level

src/setuptools_scm/config.py
```diff
@@ -82,6 +84,8 @@ def __init__(
...
+        self.normalize = normalize
```
RonnyPfannschmidt: this should integrate with version_cls in the way that when setting it to true, it should choose the NonNormalized version class and/or assert its either already being used or its not replacing a different choice

src/setuptools_scm/version.py
```diff
@@ -168,7 +201,8 @@ def format_next_version(self, guess_next, fmt="{guessed}.dev{distance}", **kw):
 def _parse_tag(tag, preformatted, config):
     if preformatted:
         return tag
-    if not isinstance(tag, Version):
+    version_cls = _get_version_cls(config)
```
RonnyPfannschmidt: by making the version class a mandatory attribute of the config this will no longer be necessary

src/setuptools_scm/version.py
```diff
@@ -20,6 +20,38 @@
...
+class NonNormalizedVersion:
+    """A non-normalizing version handler.
+
+    You can use this class to disable all version verification and normalization.
+    For example you can use this to avoid git release candidate version tags
+    ("1.0.0-rc1") to be normalized to "1.0.0rc1". Only use this if you fully
+    trust the version tags.
+    """
+
+    def __init__(self, tag):
```
RonnyPfannschmidt: i consider it mandatory to ensure the version is a valid python version number here, i would recommend to subclass Version and storing the non-normalized value on  for repr/str use

RonnyPfannschmidt (changes requested):
good start, please consider the structural nitpicks

src/setuptools_scm/version.py
smarie: ok

src/setuptools_scm/version.py
smarie: ok

src/setuptools_scm/version.py
smarie: Ok I move it to the constructor of `Configuration`. Note however that this imposes to import `Version` and define `NonNormalizedVersion` in the `config.py` package (less "purity" but more practicality)

src/setuptools_scm/__init__.py
```diff
@@ -188,3 +191,22 @@ def _get_version(config):
         )
 
         return version_string
+
+
+# Public API
+__all__ = [
```
smarie, line 197: I included `NonNormalizedVersion` at the package level but realized (from a flake8 warning stating symbol not used) that there was no `__all__`. I tried to propose one to "officialize" the public symbols.

testing/test_basic_api.py
```diff
@@ -135,3 +135,24 @@ def parse(root):
...
+
+    # you can not use normalize=False and version_cls at the same time
+    with pytest.raises(ValueError):
+        setuptools_scm.get_version(normalize=False, version_cls=MyVersion)
+
+    # TODO unfortunately with PRETEND_KEY the preformatted flag becomes True
+    #  which bypasses our class. which other mechanism would be ok to use here
+    #  to create a test?
+    # monkeypatch.setenv(setuptools_scm.PRETEND_KEY, "1.0.1")
+    # assert setuptools_scm.get_version(version_cls=MyVersion) == "1"
```
smarie, line 158: I am not able to find a proper way to pass a version that I control, without entering in the "preformatted=True" bypass (and therefore not using the custom class). If you have an idea let me know. I would basically copy here the same kind of test that is done in git test file below.

If you think that this is overkill, I can just remove this comment

testing/test_git.py
```diff
@@ -104,6 +104,70 @@ def test_version_from_git(wd):
...
+    assert wd.get_version(normalize=False) == "17.33.0-rc"
+    assert wd.get_version(version_cls=NonNormalizedVersion) == "17.33.0-rc"
+    assert (
+        wd.get_version(version_cls="setuptools_scm.NonNormalizedVersion")
+        == "17.33.0-rc"
+    )
+
+
+@pytest.mark.parametrize("with_class", [False, type, str])
+def test_git_version_unnormalized_setuptools(with_class, tmpdir, wd, monkeypatch):
```
smarie, line 117: This is the test that is of interest to me, but it could be moved to the "basic api" test if I have a way to inject the "1.0.0-rc1" version. Indeed there is nothing really specific to git here. See previous comment.

smarie:
This one should be good to go. I inserted a few comments on parts where feedback is needed.

RonnyPfannschmidt (approved):
looks good, i'll get to merging either later today or on the weekend

thanks!

src/setuptools_scm/__init__.py
RonnyPfannschmidt: lets go for a "symbol" to use

default to normalized, provide non-normalized, and let people go wild with importable names

RonnyPfannschmidt: also id like to go for privatizing more of the api

testing/test_basic_api.py
RonnyPfannschmidt: pretend-key is unable to pass versions, predent key is specifically intended to go for preformatted

testing/test_git.py
RonnyPfannschmidt: it may be possible to jsut use one of the meta tests, we can do this as a followup

src/setuptools_scm/config.py
```diff
@@ -137,3 +171,28 @@ def from_file(cls, name="pyproject.toml", dist_name=None):
...
+        # store raw for str
+        self._raw_version = version
+
+    def __str__(self):
+        # return the non-normalized version (parent returns the normalized)
+        return self._raw_version
+
+    def __repr__(self):
+        # same pattern as parent
+        return f"<NonNormalizedVersion({repr(str(self))})>"
```
RonnyPfannschmidt:
```suggestion
        return f"<NonNormalizedVersion({self._raw_version!r})>"
```

smarie:
Perfect, thanks @RonnyPfannschmidt  ! Oh, another thing that is worth looking at : string support, see comment.

src/setuptools_scm/config.py
```diff
@@ -83,6 +93,30 @@ def __init__(
...
+            elif isinstance(version_cls, str):
+                try:
+                    # Not sure this will work in old python
+                    import importlib
+
+                    pkg, cls_name = version_cls.rsplit(".", 1)
+                    version_cls_host = importlib.import_module(pkg)
+                    version_cls = getattr(version_cls_host, cls_name)
+                except:  # noqa
+                    raise ValueError(f"Unable to import version_cls='{version_cls}'")
```
smarie, line 117: I added support for `version_cls` as a string, because I suspect that usage inside pyproject.toml will require it.

testing/test_basic_api.py
RonnyPfannschmidt: the basic way to do this is currently to use a number of inner apis (aka setting up config and meta, 

maybe izts time to ease testing a bit

RonnyPfannschmidt:
sorry for the extra week

smarie:
No worries, thanks very much for your time  @RonnyPfannschmidt !

smarie:
Hi @RonnyPfannschmidt , this is probably a very busy time for you, sorry to bother you with this. I was just wondering if you have a very rough target date for next version release including this feature ? Thanks a lot in advance !

RonnyPfannschmidt:
Thanks for the ping, unfortunately im pretty swamped atm, im not sure if I can get it in this week, but im relatively sure i can schedule it for the next week

smarie:
This is perfect Ronny. Thanks!

PR #589, opened 2021-07-11
Use UTF-8 encoding when reading pyproject.toml

domdfcoding:
This avoids a `UnicodeDecodeError` when reading `pyproject.toml` files containing unicode characters on Windows. These characters may be present in a project's metadata, such as its description or an author's name.

```python
>>> from setuptools_scm.config import Configuration
>>> Configuration.from_file()
  File "<stdin>", line 1, in <module>
  File "lib\site-packages\setuptools_scm\config.py", line 137, in from_file
    defn = __import__("toml").load(strm)
  File "lib\site-packages\toml\decoder.py", line 156, in load
    return loads(f.read(), _dict, decoder)
  File "lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 242: character maps to <undefined>
```

Furthermore, the [TOML specification](https://toml.io/en/v1.0.0#spec) requires files to be encoded in UTF-8.

### DIFF
--- a/testing/test_git.py
+++ b/testing/test_git.py
@@ -280,7 +280,7 @@ def test_git_archive_export_ignore(wd, monkeypatch):
 
 @pytest.mark.issue(228)
 def test_git_archive_subdirectory(wd, monkeypatch):
-    wd("mkdir foobar")
+    os.mkdir(wd.cwd / "foobar")
     wd.write("foobar/test1.txt", "test")
     wd("git add foobar")
     wd.commit()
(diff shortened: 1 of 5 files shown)


testing/test_git.py
```diff
@@ -280,7 +280,7 @@ def test_git_archive_export_ignore(wd, monkeypatch):
 
 @pytest.mark.issue(228)
 def test_git_archive_subdirectory(wd, monkeypatch):
-    wd("mkdir foobar")
+    os.mkdir(wd.cwd / "foobar")
```
domdfcoding, line 283: I had to change this as otherwise the test fails with `FileNotFoundError: [WinError 2] The system cannot find the file specified` because it can't find `mkdir`.

PR #598, opened 2021-08-05
Adding search_parent_directories to recursively search parent directory for git-root

### LINKED ISSUE #591: Search parent directories
Support that you have a library directory:
```
/path/to/mylib/.git
/path/to/mylib/my/path/myscript.py
```
and support that I want to call 
```
python /path/to/mylib/my/path/myscript.py
```
and get the version string based on `/path/to/mylib/.git`. 

Currently I use
```python
import git
root = git.Repo(os.path.dirname(__file__), search_parent_directories=True).working_tree_dir
myversion = get_version(root=root)
```

Now, it would be great to have the same `search_parent_directories` directly in *setuptools_scm*.

tdegeus:
Fixes https://github.com/pypa/setuptools_scm/issues/591

### DIFF
--- a/src/setuptools_scm/config.py
+++ b/src/setuptools_scm/config.py
@@ -75,6 +75,7 @@ class Configuration:
         dist_name=None,
         version_cls=None,
         normalize=True,
+        search_parent_directories=False,
     ):
         # TODO:
         self._relative_to = relative_to
@@ -92,6 +93,8 @@ class Configuration:
         self.tag_regex = tag_regex
         self.git_describe_command = git_describe_command
         self.dist_name = dist_name
+        self.search_parent_directories = search_parent_directories
+        self.parent = None
 
         if not normalize:
             # `normalize = False` means `version_cls = NonNormalizedVersion`
--- a/src/setuptools_scm/discover.py
+++ b/src/setuptools_scm/discover.py
@@ -2,11 +2,54 @@ import os
 from .utils import trace, iter_entry_points
 
 
-def iter_matching_entrypoints(path, entrypoint):
-    trace("looking for ep", entrypoint, path)
-    for ep in iter_entry_points(entrypoint):
-        if os.path.exists(os.path.join(path, ep.name)):
-            if os.path.isabs(ep.name):
-                trace("ignoring bad ep", ep)
-            trace("found ep", ep)
-            yield ep
+def walk_potential_roots(root, search_parents=True):
+    """
+    Iterate though a path and each of its parents.
+    :param root: File path.
+    :param search_parents: If ``False`` the parents are not considered.
+    """
+
+    if not search_parents:
+        yield root
+        return
+
+    tail = root
+
+    while tail:
+        yield root
+        root, tail = os.path.split(root)
+
+
+def match_entrypoint(root, name):
+    """
+    Consider a ``root`` as entry-point.
+    :param root: File path.
+    :param name: Subdirectory name.
+    :return: ``True`` if a subdirectory ``name`` exits in ``root``.
+    """
+
+    if os.path.exists(os.path.join(root, name)):
+        if not os.path.isabs(name):
+            return True
+        trace("ignoring bad ep", name)
+
+    return False
+
+
+def iter_matching_entrypoints(root, entrypoint, config):
+    """
+    Consider different entry-points in ``root`` and optionally its parents.
+    :param root: File path.
+    :param entrypoint: Entry-point to consider.
+    :param config: Configuration,
+        read ``search_parent_directories``, write found parent to ``parent``.
+    """
+
+    trace("looking for ep", entrypoint, root)
+
+    for wd in walk_potential_roots(root, config.search_parent_directories):
+        for ep in iter_entry_points(entrypoint):
+            if match_entrypoint(wd, ep.name):
+                trace("found ep", ep, "in", wd)
+                config.parent = wd
+                yield ep
--- a/src/setuptools_scm/git.py
+++ b/src/setuptools_scm/git.py
@@ -104,6 +104,27 @@ def fail_on_shallow(wd):
         )
 
 
+def get_working_directory(config):
+    """
+    Return the working directory (``GitWorkdir``).
+    """
+
+    if config.parent:
+        return GitWorkdir.from_potential_worktree(config.parent)
+
+    if config.search_parent_directories:
+        return search_parent(config.absolute_root)
+
+    wd = GitWorkdir.from_potential_worktree(config.absolute_root)
+
+    if wd is not None:
+        return wd
+
+    from .hg_git import GitWorkdirHgClient
+
+    return GitWorkdirHgClient.from_potential_worktree(config.absolute_root)
+
+
 def parse(root, describe_command=None, pre_parse=warn_on_shallow, config=None):
     """
     :param pre_parse: experimental pre_parse action, may change at any time
@@ -111,13 +132,11 @@ def parse(root, describe_command=None, pre_parse=warn_on_shallow, config=None):
     if not config:
         config = Configuration(root=root)
 
-    wd = GitWorkdir.from_potential_worktree(config.absolute_root)
-    if wd is None:
-        from .hg_git import GitWorkdirHgClient
+    wd = get_working_directory(config)
 
-        wd = GitWorkdirHgClient.from_potential_worktree(config.absolute_root)
     if wd is None:
         return
+
     if pre_parse:
         pre_parse(wd)
 
@@ -171,3 +190,30 @@ def _git_parse_describe(describe_output):
     tag, number, node = describe_output.rsplit("-", 2)
     number = int(number)
     return tag, number, node, dirty
+
+
+def search_parent(dirname):
+    """
+    Walk up the path to find the `.git` directory.
+    :param dirname: Directory from which to start searching.
+    """
+
+    # Code based on:
+    # https://github.com/gitpython-developers/GitPython/blob/main/git/repo/base.py
+
+    curpath = os.path.abspath(dirname)
+
+    while curpath:
+
+        try:
+            wd = GitWorkdir.from_potential_worktree(curpath)
+        except Exception:
+            wd = None
+
+        if wd is not None:
+            return wd
+
+        curpath, tail = os.path.split(curpath)
+
+        if not tail:
+            return None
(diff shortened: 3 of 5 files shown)


src/setuptools_scm/git.py
```diff
@@ -171,3 +171,64 @@ def _git_parse_describe(describe_output):
     tag, number, node = describe_output.rsplit("-", 2)
     number = int(number)
     return tag, number, node, dirty
+
+
+def is_git_dir(d):
```
RonnyPfannschmidt: please note that the from_potential_worktree helper in the git worktree already finds the root using git directly (this works correctly on windows and in special situations on windows)

src/setuptools_scm/config.py
```diff
@@ -93,6 +95,11 @@ def __init__(
         self.git_describe_command = git_describe_command
         self.dist_name = dist_name
 
+        if search_parent_directories:
```
RonnyPfannschmidt: the search should happen in the wd parse functions, else the config object is misconfigured, and that makes it harder to reason about

also this is a git only helper, so currently it is broken for non- git like mercurial

RonnyPfannschmidt: lets put it on config as a value, then the parse function can deal with the details per scm

RonnyPfannschmidt (changes requested):
thank you for a initial attempt

as this is a git only feature, it should primarily be handled in the git tooling

the parse function takes the config, and that should pass the option on


its not necessary to add hg support, but it shouldn't happen in config


good work in preparing the details and the test

src/setuptools_scm/git.py
tdegeus: Great, so I should just replace
```python
if is_git_dir(dirname):
```
with 
```python
GitWorkdir.from_potential_worktree(dirname) is not None:
```
and get rid of the `is_git_dir` helper?

src/setuptools_scm/config.py
tdegeus: Honestly I'm a little bit lost in finding the parse functions ;)

tdegeus: Ah maybe I get it. Do you mean something like `_do_parse`?

tdegeus:
@RonnyPfannschmidt Concerning your second comment: not entirely sure that I understand what you are doing, could you give some pointers based on the edit?

src/setuptools_scm/git.py
RonnyPfannschmidt: there is a samefile check that needs to ensure the parent searching is working correctly, that needs to be conditional

src/setuptools_scm/config.py
RonnyPfannschmidt: i meant the parse function in the git module (right around the place that does the call to from potential workdir

src/setuptools_scm/git.py
tdegeus: So I use `GitWorkdir.from_potential_worktree`, but I don't like that I needed the `try: ... except: ...` because of `git` command that fails

src/setuptools_scm/config.py
tdegeus: I added it to `git.parse` and that works well. What I however don't understand is that I don't think that this function is called from `get_version`?

RonnyPfannschmidt (approved):

tdegeus:
So indeed, as the CI shows, it does not seem that `git.parse` is called (or at least not in time: form the code I don't fully get how `git.parse` is called...).

RonnyPfannschmidt:
@tdegeus it seems it just returns none,

the helper in the _do_parse function tries all parse functions

i can dig into this on the weeken at earliest

tdegeus:
I did some debugging @RonnyPfannschmidt . The problem is that it is not the result of `git.parse` that will determine if it will be used. Rather, there is a naive try here
https://github.com/pypa/setuptools_scm/blob/dfb4bea08dc03dcd560caf7e00b6b8aa837b2a33/src/setuptools_scm/discover.py#L8
with `path` equal to the initially specified (absolute) root. 

## Example

Consider that I have my package here
```
/path/to/mypackage/.git
```
and that I'm running from
```
/path/to/mypackage/src
```
then 
https://github.com/pypa/setuptools_scm/blob/dfb4bea08dc03dcd560caf7e00b6b8aa837b2a33/src/setuptools_scm/discover.py#L8
decides not to try parsing with *git* because 
```
/path/to/mypackage/src/.git
```
does not exist.

So even though `git.parse` does work and will recognise the right working directory, it is simply not considered. 

I think it is better that you comment how to improve this. I can try to propose something naive though

tdegeus:
Here is a proposed work-around, let me know what you think!

RonnyPfannschmidt:
:facepalm: my appologoes, i completely forgot about that sanity check, which is not sane in the face of upward search

RonnyPfannschmidt:
all on all considered, the best solution given the necessary workaround, might actually be to just walk the parent directories in that function 

then the search for parents would work for all scms and for git as well by default

what do you think about hat

tdegeus:
I think that would be fine! Do you propose to keep walking the parent directories still as an option ?

RonnyPfannschmidt:
@tdegeus yes, it should be a opt in for scripts - a future release may decide to enable it by default for the get_version function

right now people use it in pre-configured situations where always enabling it may cause a breakage/wrong result

so the path forwards is have it a opt in, and maybe later have it as a opt out for script usage

RonnyPfannschmidt:
@tdegeus btw, i recommend installing pre-commit and useing it for this repo

i observe you having to fight the linting in the CI
 pre-commit is a tool that helps you sort this out before you even commit

tdegeus:
> @tdegeus btw, i recommend installing pre-commit and useing it for this repo
> 
> i observe you having to fight the linting in the CI
> pre-commit is a tool that helps you sort this out before you even commit

I will look into it ;)

tdegeus:
@RonnyPfannschmidt I implemented the search for the parent directly in `discovery`. What might make things nicer is to store the path in `config` on not re-search. I don't think `Configure` has an API to that end. Or am I looking over it?

src/setuptools_scm/discover.py
```diff
@@ -2,15 +2,40 @@
...
+            return None
+
+
+def iter_matching_entrypoints(path, entrypoint, config):
+
     trace("looking for ep", entrypoint, path)
-    if not search_parent_directories:
+
+    if config.search_parent_directories:
         for ep in iter_entry_points(entrypoint):
```
RonnyPfannschmidt, line 51: unfortunately this loop has a mistake that can break discovery in some cases

for example if scm repositories of different scms for whatever reason are contained within each other,
then the wrong outer one may be found 

unfortunately its necessary to walk the potential roots one by one, and check each level for all entrypoints


i proposes a walk_potential_roots(root, search_parents) , that returns root if search_parents if false,

tdegeus: Yeah I realised this as well 😴, I will correct for it!

tdegeus:
@RonnyPfannschmidt This should do it, ready for your review.

(I noticed in the CI `188 passed, 1 skipped, 1 xfailed, 1 xpassed` isn't that worrisome?)

tdegeus:
P.S. the `walk_roots` could be simpler, and I think that that is what you have in mind. This is what I was thinking about
```python
def walk_potential_roots(root, search_parents = True):

    if not search_parents:
        yield root
        return

    while True:
        root, tail = os.path.split(root)
        if not tail:
            return
        yield root
```
but one needs the unsplitted `root` on the first iteration with `search_parents = True`, but that I couldn't think of how to do. Maybe you have a better idea?

RonnyPfannschmidt:
Always yield root, assign root to tail before the loop and then have a while tail loop in a if block

RonnyPfannschmidt (approved):
looks good now, i'll aim for merge and release on the weekend

tdegeus:
Thanks @RonnyPfannschmidt  that was indeed what needed to be done.

I couldn't help myself to leave an elegant code ;)

tdegeus:
BTW I exclusively use git so I don't feel super confident to do it, but the other `parse` functions could be extended as well, at least using `config.parent`.

RonnyPfannschmidt:
@tdegeus thats ok  do you want to re-base a little to "beautify" the commits or should i merge

tdegeus:
Done @RonnyPfannschmidt  !

RonnyPfannschmidt:
all failures seem to be related to a unfortunate integration of hg-git, it may require some exra help, im ok with temporary disabling it so i can bugfix this as followup

RonnyPfannschmidt:
@tdegeus befor you do anything else, gimme a few to locally debug

tdegeus:
Sure!

src/setuptools_scm/git.py
```diff
@@ -104,20 +104,39 @@ def fail_on_shallow(wd):
...
+        return search_parent(config.absolute_root)
+
+    wd = GitWorkdir.from_potential_worktree(config.absolute_root)
+
+    if wd is not None:
+        return wd
+
+    from .hg_git import GitWorkdirHgClient
+
+    return GitWorkdirHgClient.from_potential_worktree(config.absolute_root)
```
tdegeus, line 125: @RonnyPfannschmidt 

Concerning the failing CI : in one of the final commits I made I added this function under pressure from *flake8*. I'm quite convinced that the old implementation is unchanged, but it would be good to double check.

src/setuptools_scm/discover.py
```diff
@@ -2,11 +2,54 @@
...
+        if not os.path.isabs(name):
+            return True
+        trace("ignoring bad ep", name)
+
+    return False
+
+
+def iter_matching_entrypoints(root, entrypoint, config):
+    """
+    Consider different entry-points in ``root`` and optionally its parents.
```
tdegeus, line 41: Also, since the last passing CI (it is a bit annoying that it does not run on every commit because I'm a first time contributor here) I cleaned up a bit here.

RonnyPfannschmidt: @tdegeus i identified a bug in hg_git, i separated thins, release incoming

RonnyPfannschmidt:
@tdegeus thanks for providing this, i'll shuffle some details around to harden the hg-git stuff

tdegeus:
@RonnyPfannschmidt Do you have some idea when there will be release that includes this?

RonnyPfannschmidt:
@tdegeus the original plan was roughly about 11 days ago, then life happened, i have some time within the next 2-3 days to finalize the release and hope to get it done by mid week

tdegeus:
Awesome ! It's just nice to know that something is coming in the near future !

Issue #608, opened 2021-08-27
Missing module name tomli


RouquinBlanc:
Since the release of setuptools_scm 6.1.1 a few minutes ago, our jenkins jobs are failing on the following error:

```
Traceback (most recent call last):
  File "/var/lib/jenkins/workspace/nidar-orchid_master/setup.py", line 7, in <module>
    setuptools.setup(
  File "/root/.pyenv/versions/3.9.5/lib/python3.9/site-packages/setuptools/__init__.py", line 153, in setup
    return distutils.core.setup(**attrs)
  File "/root/.pyenv/versions/3.9.5/lib/python3.9/distutils/core.py", line 108, in setup
    _setup_distribution = dist = klass(attrs)
  File "/root/.pyenv/versions/3.9.5/lib/python3.9/site-packages/setuptools/dist.py", line 434, in __init__
    _Distribution.__init__(self, {
  File "/root/.pyenv/versions/3.9.5/lib/python3.9/distutils/dist.py", line 292, in __init__
    self.finalize_options()
  File "/root/.pyenv/versions/3.9.5/lib/python3.9/site-packages/setuptools/dist.py", line 743, in finalize_options
    ep(self)
  File "/var/lib/jenkins/workspace/nidar-orchid_master/.eggs/setuptools_scm-6.1.1-py3.9.egg/setuptools_scm/integration.py", line 52, in infer_version
    config = Configuration.from_file(dist_name=dist_name)
  File "/var/lib/jenkins/workspace/nidar-orchid_master/.eggs/setuptools_scm-6.1.1-py3.9.egg/setuptools_scm/config.py", line 187, in from_file
    defn = _load_toml(data)
  File "/var/lib/jenkins/workspace/nidar-orchid_master/.eggs/setuptools_scm-6.1.1-py3.9.egg/setuptools_scm/config.py", line 59, in _lazy_tomli_load
    from tomli import loads
ModuleNotFoundError: No module named 'tomli'
```

Maybe a missing dependency somewhere?

Limess:
We're also seeing this issue and it's blocking all of our pip installs + builds

```
DEPRECATION: A future pip version will change local packages to be built in-place without first copying to a temporary directory. We recommend you use --use-feature=in-tree-build to test your packages with this new behavior before it becomes the default.
       pip 21.3 will remove support for this functionality. You can find discussion regarding this at https://github.com/pypa/pip/issues/7555.
        ERROR: Command errored out with exit status 1:
[...]

PR #609, opened 2021-08-27
fix tomli dependency,

PythonicNinja:
related to https://github.com/pypa/setuptools_scm/issues/608

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,7 +28,7 @@ classifiers =
 packages = find:
 install_requires =
     setuptools>=45
-    tomli~=1.0.0
+    tomli~=1.0
 python_requires = >=3.6
 package_dir =
     =src
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,7 @@ classifiers =
 packages = find:
 install_requires =
     setuptools>=45
+    tomli~=1.0.0
 python_requires = >=3.6
 package_dir =
     =src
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -28,6 +28,7 @@ classifiers =
 packages = find:
 install_requires =
     setuptools>=45
+    tomli~=1.0.0
```
RonnyPfannschmidt: i believe by now the pin has to be corrected to `tomli~=1.0`

RonnyPfannschmidt (changes requested):

setup.cfg
SebastianRehfeldt: Is it required to keep tomli in `[options.extras_require]` or should it be removed there?

PythonicNinja: ok changing 👍

PythonicNinja: @RonnyPfannschmidt done

PythonicNinja: I would say that if it's required we should also remove it from `extras_require`

PythonicNinja: done

RonnyPfannschmidt: actually it should stay only in the extra, and the helper to load it should error if its not found only and only if a `tool.setuptools_scm` section is actually in `pyproject.toml`

RonnyPfannschmidt:
unfortunately i'll be away from the computer for about 1-2 hours now for childcare, see you later

RonnyPfannschmidt (approved):

RonnyPfannschmidt:
since pip is used for install anyway, lets go for a hard dependency anyway, i'll update the docs and do a release afterwards

PR #610, opened 2021-08-27
prepare 6.2.0 release

RonnyPfannschmidt:


henryiii:
Removing the extra produces a warning when directly installed, not sure if it's visible as a `pyproject.toml` requirement. You could leave an empty extra to keep the warning out.

`setuptools_scm[toml]>4` would produce a warning on 6.2.0, for example, but anything less than `setuptools_scm>6.2` would be broken without the extra.

PR #616, opened 2021-08-29
combined preparation for the 6.3.0 release and a regression resolution

RonnyPfannschmidt:
this fixes the support for old setuptools while not starting to claim to support old setuptools

* fix #612: depend on packaging to ensure version parsing parts
* fix #611: correct the typo that hid away the toml extra and add it in ``setup.py`` as well
* fix #615: restore support for the git_archive plugin which doesn't pass over the config

Issue #611, opened 2021-08-27
setuptools-scm 6.2.0 has no such extra feature 'toml


Kyle-Verhoog:
:wave: we just started seeing this in our docs build: https://readthedocs.org/projects/ddtrace/builds/14573515/

`/home/docs/checkouts/readthedocs.org/user_builds/ddtrace/envs/latest/bin/python ./setup.py install --force`

```
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/ddtrace/envs/latest/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2739, in requires
    deps.extend(dm[safe_extra(ext)])
KeyError: 'toml'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "./setup.py", line 150, in <module>
    setup(
  File "/home/docs/checkouts/readthedocs.org/user_builds/ddtrace/envs/latest/lib/python3.8/site-packages/setuptools/__init__.py", line 152, in setup
    _install_setup_requires(attrs)
  File "/home/docs/checkouts/readthedocs.org/user_builds/ddtrace/envs/latest/lib/python3.8/site-packages/setuptools/__init__.py", line 147, in _install_setup_requires
    dist.fetch_build_eggs(dist.setup_requires)
  File "/home/docs/checkouts/readthedocs.org/user_builds/ddtrace/envs/latest/lib/python3.8/site-packages/setuptools/dist.py", line 785, in fetch_build_eggs
    resolved_dists = pkg_resources.working_set.resolve(
  File "/home/docs/checkouts/readthedocs.org/user_builds/ddtrace/envs/latest/lib/python3.8/site-packages/pkg_resources/__init__.py", line 780, in resolve
    new_requirements = dist.requires(req.extras)[::-1]
  File "/home/docs/checkouts/readthedocs.org/user_builds/ddtrace/envs/latest/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2741, in requires
    raise UnknownExtra(
pkg_resources.UnknownExtra: setuptools-scm 6.2.0 has no such extra feature 'toml'
```

haven't gotten a chance to dig in further - will update if I get the chance!

RonnyPfannschmidt:
i identified a setuptools bug about provides_extra not being added when using `setup.cfg`, only when using `setup.py`

vss888:
Is there any work around the bug?

sagrawal-code:
> Is there any work around

[...]

PR #618, opened 2021-08-31
setup.cfg: fix typo preventing toml extra from actually existing

eli-schwartz:
Fixup for #610 which makes it actually take effect.


RonnyPfannschmidt:
thanks, however this fix is already included in #616

eli-schwartz:
Ah, okay... I saw that in the list of PRs but overlooked it because it didn't sound like it had anything to do with the issue, and #611 claimed that it was a bug in setuptools itself so I got the impression that it wasn't being fixed. I was maybe too hasty... thanks for pointing that out.

PR #629, opened 2021-09-05
Fix TypeError in version.py

musicmrman99:
I was using this with the setuptools_scm_git_archive plugin and came across a `TypeError: expected string or bytes-like object` in version.py:20. I had a quick look at the code and found that it seems to convert to a string only when it's already a string - I'm guessing that was an accident :) After double-checking, something (possibly setuptools_scm_git_archive) was passing in a `packaging.version.Version` instance, so the TypeError makes sense.

I didn't make an issue, as it looks like a simple fix.

### DIFF
--- a/src/setuptools_scm/version.py
+++ b/src/setuptools_scm/version.py
@@ -16,7 +16,7 @@ SEMVER_LEN = 3
 
 
 def _parse_version_tag(tag, config):
-    tagstring = tag if not isinstance(tag, str) else str(tag)
+    tagstring = tag if isinstance(tag, str) else str(tag)
     match = config.tag_regex.match(tagstring)
 
     result = None


RonnyPfannschmidt:
Good find, i believe we should support passing version objects however

musicmrman99:
Through string conversion, this patch enables support for non-string objects, including support for the latest versions of most of the possible Version classes mentioned in `_version_cls.py` (which I'm guessing is this project's centralised definition of 'Version'), as well as any custom class passed in:

- The latest versions of `packaging.version.Version` and `pkg_resources._vendor.packaging.version.Version` can both convert themselves to a string in a standard format that's parsable by the default config's `tag_regex` (quick refs: https://github.com/pypa/packaging/blob/main/packaging/version.py#L293 and https://github.com/pypa/setuptools/blob/main/pkg_resources/_vendor/packaging/version.py#L305). I wouldn't know about any compatibility problems from older versions of the packages, though.
- A `__str__()` could be added to the fallback class in `_version_cls.py` that is parsable. (Thinking about it, we should probably do this anyway.)
- The docs say that any class passed for `version_cls` to `use_scm_version` must have an `__str__()` that returns the normalised version string, so custom classes should (?) also be compatible with the default config's `tag_regex`.

Regardless, do you think it would be better to support the `Version` type more directly (eg. `try` for a Version first to avoid formatting to string, only to re-parse with a regex), or keep plain string conversion (as fixed by this patch)?

beckermr:
I really could use this patch. Can't we delay the discussion of version objects to another PR?

beckermr:
bump here @RonnyPfannschmidt - can we merge this?

PR #634, opened 2021-09-14
build/test fixes

RonnyPfannschmidt:
- enhance mercurial detection in hg_git tests
- drop devpi testenv
- drop flake8 env for pre-commit
- trim down the toml extra and only make it need modern setuptools
- add test requirements as extra
- run tox selfcheck on all tox python versions
- drop unneeded flake8 excludes
- use build to build the artifacts in the chec_dist env
- drop the dist env, setup.py shouldn't be used for that
- minimal acceptance test for dummy pretend versions

### DIFF
--- a/src/setuptools_scm/__init__.py
+++ b/src/setuptools_scm/__init__.py
@@ -4,8 +4,8 @@
 """
 import os
 import warnings
-from logging import getLogger as _get_logger
 
+from ._version_cls import _version_as_tuple
 from ._version_cls import NonNormalizedVersion
 from ._version_cls import Version
 from .config import Configuration
@@ -73,7 +73,7 @@ def _version_from_entrypoints(config: Configuration, fallback=False):
             return version
 
 
-def dump_version(root, version, write_to, template=None):
+def dump_version(root, version: str, write_to, template: "str | None" = None):
     assert isinstance(version, str)
     if not write_to:
         return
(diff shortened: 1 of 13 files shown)


src/setuptools_scm/__init__.py
```diff
@@ -86,13 +87,19 @@ def dump_version(root, version, write_to, template=None):
                 os.path.splitext(target)[1], target
             )
         )
+    try:
+        parsed_version = Version(version)
+    except Exception:
```
FFY00: Shouldn't this be `packaging.version.InvalidVersion`? You can re-export `InvalidVersion` from your version helper module.

src/setuptools_scm/__init__.py
```diff
@@ -86,13 +87,19 @@ def dump_version(root, version, write_to, template=None):
...
 
-    parsed_version = Version(version)
-    version_fields = parsed_version.release
-    if parsed_version.dev is not None:
-        version_fields += (f"dev{parsed_version.dev}",)
-    if parsed_version.local is not None:
-        version_fields += (parsed_version.local,)
+        log = _get_logger("setuptools_scm")
+        log.exception("failed to parse version %s", version)
+        version_fields = (0, 0, 0)
```
FFY00: Wouldn't it be best to use the version as-is -- `version_fields = (version,)`? Afterall, it is perfectly possible that this is a legacy version, and I don't think we should fallback to `0.0.0` unless there is something actually requiring us to (if there is, please let me know).

FFY00 (changes requested):

src/setuptools_scm/__init__.py
RonnyPfannschmidt: instead of re-exporting i put the tuple helper into the fle as well

PR #644, opened 2021-10-21
Escape glob pattern in git describe to avoid shell expansion.

### LINKED ISSUE #643: DEFAULT_DESCRIBE fails if working dir contains file with digit in name (missing shell escape)
The default describe invocation doesn't quote the match pattern leading to is being expanded by bash if a file with that name exists or `shopt -s nullglob` is in effect.

Should be as simple as quoting the pattern. Will add a pr

tapetersen:
To fix #643. I do realize now that this isn't a real problem as it's used here since the command is not actually run through a shell but it did bite me when copying it to a CI script to check the version and shouldn't hurt.

Feel free to close the PR and issue if you don't think it's necessary though.

### DIFF
--- a/src/setuptools_scm/git.py
+++ b/src/setuptools_scm/git.py
@@ -16,7 +16,15 @@ from .version import meta
 # If testing command in shell make sure to quote the match argument like
 # '*[0-9]*' as it will expand before being sent to git if there are any matching
 # files in current directory.
-DEFAULT_DESCRIBE = ["git", "describe", "--dirty", "--tags", "--long", "--match", "*[0-9]*"]
+DEFAULT_DESCRIBE = [
+    "git",
+    "describe",
+    "--dirty",
+    "--tags",
+    "--long",
+    "--match",
+    "*[0-9]*",
+]
 
 
 class GitWorkdir(Workdir):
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
Good idea, copy to shell is important

tapetersen:
It does seem to break in Windows though (the tests run with GitHub actions) so I guess it's not really mergable. I can't see a way to ensure its escaped as shell if copied without the quotes being passed to git when run in python without shell which I assume is the reason for the fail on Windows. 

Another reason I prefer python to shell I think but I don't see a way to improve the code with respect to copying to shell without breaking the use in the module then. 

Maybe it warrants a comment for warning and it may have been clearer of it was split as a list instead off a single string that it doesn't work as expected of copied to shell but the rest of the module is in they style as well so I don't think that is the way.

Thanks for a great project either way you decide to resolve it

RonnyPfannschmidt:
I shall investigate after I recover from my cold

RonnyPfannschmidt:
@tapetersen windows and linux use different quoting options

i think i will have to switch to list based commands that dont involve a shell

tapetersen:
@RonnyPfannschmidt I've updated it to warn and use  pre-split list instead of the string (Not necessary as the code uses shlex.split on posix otherwise so same effect there) but should be clearer with the caveat.

RonnyPfannschmidt (approved):
will mere later and add to the changelog

PR #646, opened 2021-10-29
Consider configuration when running `__main__`

abravalheri:
Since the usage with `setup.py` is deprecated (according to `setuptools_scm` README), it would be nice to have a command that give the same results as:

    python setup.py --version

The idea of the changes implemented here is to do that using the `__main__` module.

The methodology consists in using the `find_files` function to look for a `pyproject.toml` file and then read the configuration from there.

If `pyproject.toml` file is not found or it does not contain a `[tool.setuptools_scm]` table, the command works as implemented before (guessing the next version).

### DIFF
--- a/src/setuptools_scm/__main__.py
+++ b/src/setuptools_scm/__main__.py
@@ -1,22 +1,67 @@
-import sys
+import argparse
+import os
+import warnings
 
 from setuptools_scm import _get_version
-from setuptools_scm import get_version
 from setuptools_scm.config import Configuration
+from setuptools_scm.discover import walk_potential_roots
 from setuptools_scm.integration import find_files
 
 
 def main() -> None:
-    files = list(sorted(find_files("."), key=len))
-    try:
-        pyproject = next(fname for fname in files if fname.endswith("pyproject.toml"))
-        print(_get_version(Configuration.from_file(pyproject)))
-    except (LookupError, StopIteration):
-        print("Guessed Version", get_version())
+    opts = _get_cli_opts()
+    root = opts.root or "."
 
-    if "ls" in sys.argv:
-        for fname in files:
+    try:
+        pyproject = opts.config or _find_pyproject(root)
+        root = opts.root or os.path.relpath(os.path.dirname(pyproject))
+        config = Configuration.from_file(pyproject)
+        config.root = root
+    except (LookupError, FileNotFoundError) as ex:
+        # no pyproject.toml OR no [tool.setuptools_scm]
+        warnings.warn(f"{ex}. Using default configuration.")
+        config = Configuration(root)
+
+    print(_get_version(config))
+
+    if opts.command == "ls":
+        for fname in find_files(config.root):
             print(fname)
+
+
+def _get_cli_opts():
+    prog = "python -m setuptools_scm"
+    desc = "Print project version according to SCM metadata"
+    parser = argparse.ArgumentParser(prog, description=desc)
+    # By default, help for `--help` starts with lower case, so we keep the pattern:
+    parser.add_argument(
+        "-r",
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (commented):
I like the idea,
Im still recovering from a cold and aim to dig into the details mid next week

Thanks for preparing this!

abravalheri:
No problems, take your time. I wish you a god recovery :)

src/setuptools_scm/__main__.py
```diff
@@ -1,13 +1,21 @@
...
 
+from setuptools_scm import _get_version
 from setuptools_scm import get_version
+from setuptools_scm.config import Configuration
 from setuptools_scm.integration import find_files
 
 
 def main() -> None:
-    print("Guessed Version", get_version())
+    files = list(sorted(find_files("."), key=len))
```
abravalheri: This operation is a bit expensive (just to find the `pyproject.toml` at the root), however if we want to consider all SCM including 3rd party extensions, it becomes a bit tricky.

Alternatively `file_finder_hg._hg_toplevel` or `file_finder_git._git_toplevel` could be used, however that will not include other SCM that might be supported via extensions...

A natural question would be: is this command required to be very optimised? Can't the user afford a bit of time here just to check how `setuptools_scm` is returning the version? (I imagine this command would just be used for debugging purposes, once the project is packaged, the version is already set in the distribution file...)

abravalheri: This could be an alternative implementation:

```python
import os
import sys

from setuptools_scm import _get_version, get_version
from setuptools_scm.config import Configuration
from setuptools_scm.discover import walk_potential_roots
from setuptools_scm.integration import find_files


def main() -> None:
    try:
        pyproject = _find_config(".")
        print(_get_version(Configuration.from_file(pyproject)))
    except (FileNotFoundError, LookupError) as ex:
        print(str(ex))  # e.g. "./pyproject.toml does not contain a tool.setuptools_scm section"
        print(f"Guessed Version {get_version()}")

    if "ls" in sys.argv:
        for fname in find_files("."):
            print(fname)


def _find_config(parent):
    for directory in walk_potential_roots(parent):
        pyproject = os.path.join(directory, "pyproject.toml")
        if os.path.exists(pyproject):
            return pyproject
    raise FileNotFoundError("pyproject.toml not found")

```

RonnyPfannschmidt: lets add a flag to make config non-optional and a option to tell where pyproject.toml is

additionally a warning can happen if no pyproject.toml was found

abravalheri:
Thank you very much for the review @RonnyPfannschmidt.
I added some `argparse` options more or less in the direction that you pointed out.

(In the end it made more sense for me to have 2 complementary options one for the root dir other for the config, since it simplified the logic)

I don't know if I understood correctly what you mean by:
> lets add a flag to make config non-optional and a option to tell where pyproject.toml

Does it make sense to have 1 CLI flag + 1 CLI option for that?

FlorianWilhelm:
Hi @RonnyPfannschmidt, I hope you are doing good :-) 

The more I tell people to stop using `setup.py` for everything, like `python setup.py bdist_wheel`, the more I miss an alternative to `python setup.py --version`. What do you think about @abravalheri's PR? I think it would be really helpful as setuptools moves more and more towards `pyproject.toml`.

RonnyPfannschmidt:
It looks good at first glance, i hope to be able to get back to things this week

RonnyPfannschmidt:
thanks again, i finally got back to this after having a chat with @FlorianWilhelm today

abravalheri:
Thank you very much for the review @RonnyPfannschmidt !

PR #649, opened 2021-11-15
Add instructions for usage with docker

### LINKED ISSUE #568: document painless git usage in buildkit docker usage
For future reference (building only works in recent docker versions with buildkit support) you can avoid having `.git` in one of your layers by temporarily bind-mounting it during build time [[docs]](https://github.com/moby/buildkit/blob/v0.8.3/frontend/dockerfile/docs/syntax.md):

```dockerfile
RUN --mount=source=.git,target=.git,type=bind \
    pip install --no-cache-dir -e .
```

_Originally posted by @ddelange in https://github.com/pypa/setuptools_scm/issues/77#issuecomment-844927695_

physikerwelt:
Closes #568

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -231,7 +231,8 @@ version that is used to cache the results of the pip install process:
 
     FROM python
     COPY pyproject.toml
-    RUN SETUPTOOLS_SCM_PRETEND_VERSION=ignore pip install -e .[test]
+    ARG PSEUDO_VERSION=1
+    RUN SETUPTOOLS_SCM_PRETEND_VERSION=${PSEUDO_VERSION} pip install -e .[test]
     RUN --mount=source=.git,target=.git,type=bind pip install -e .
 
 Note that running this Dockerfile requires docker with BuildKit enabled
--- a/README.rst
+++ b/README.rst
@@ -214,7 +214,8 @@ Usage from Docker
 
 By default, docker will not copy the ``.git``  folder into your container.
 Therefore, builds with version inference might fail.
-Consequently, you can use the following snipped to infer the version from the host os without copying the entire ``.git`` folder to your Dockerfile.
+Consequently, you can use the following snipped to infer the version from
+the host os without copying the entire ``.git`` folder to your Dockerfile.
 
 .. code:: dockerfile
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -209,6 +209,32 @@ The underlying reason is, that services like *Read the Docs* sometimes change
...
+Consequently, you can use the following snipped to infer the version from the host os without copying the entire ``.git`` folder to your Dockerfile.
+
+.. code:: dockerfile
+
+    RUN --mount=source=.git,target=.git,type=bind \
+        pip install --no-cache-dir -e .
+
+However, this build step introduces a dependency to the state of your local
+.git folder the build cache.
+To optimize build caching, one can use ``SETUPTOOLS_SCM_PRETEND_VERSION=1`` in
```
RonnyPfannschmidt: for safety/santiy across other packages, this should use `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_$PKG_NAME`

in addition a recommendation to have a optional build arg to set that version to something realistic would be very important

RonnyPfannschmidt (changes requested):
thanks for adding this, the advice is solid, but the choice in the version number and the pretend passover needs a extra touch

README.rst
physikerwelt: Thank you. This is an other option I did not consider. My approach was to use the PRETEND_VERSION feature only in the particular layer, thus the version number will not be used, because it is overwritten by the next instruction. In my setup the additional line speeds up the build form 1 minute to 5 seconds. I will elaborate on that.

physikerwelt: I inspected this a bit further. The fist variant
```dockerfile
FROM python
COPY pyproject.toml
RUN --mount=source=.git,target=.git,type=bind \
    pip install --no-cache-dir -e .
```
 generates slightly smaller images

Screenshot from 2021-11-16 10-02-26

However, the second 
```dockerfile
FROM python
COPY pyproject.toml
RUN SETUPTOOLS_SCM_PRETEND_VERSION=1 pip install -e .[test]
RUN --mount=source=.git,target=.git,type=bind pip install -e .
```
is much faster for repetitive builds.
Screenshot from 2021-11-16 09-48-37

I adjusted the documentation slightly. As I am totally new to this project, it would be highly appreciated that you perform further edits on this PR to ensure a coherent README.

README.rst
```diff
@@ -214,26 +214,32 @@ Usage from Docker
...
+.git folder the build cache and triggers the long-running pip install process on every build.
+To optimize build caching, one can use an environment variable to pretend a pseudo
+version that is used to cache the results of the pip install process:
 
 .. code:: dockerfile
 
     FROM python
     COPY pyproject.toml
-    RUN SETUPTOOLS_SCM_PRETEND_VERSION=1 pip install -e .[test]
+    RUN SETUPTOOLS_SCM_PRETEND_VERSION=ignore pip install -e .[test]
```
RonnyPfannschmidt: this versions will break certain cyclic dependencies where a specfic/working/valid version for a dependency is actually needed

physikerwelt: oups. fixed.

README.rst
```diff
@@ -209,6 +209,39 @@ The underlying reason is, that services like *Read the Docs* sometimes change
...
+.git folder the build cache and triggers the long-running pip install process on every build.
+To optimize build caching, one can use an environment variable to pretend a pseudo
+version that is used to cache the results of the pip install process:
+
+.. code:: dockerfile
+
+    FROM python
+    COPY pyproject.toml
+    ARG PSEUDO_VERSION=1
+    RUN SETUPTOOLS_SCM_PRETEND_VERSION=${PSEUDO_VERSION} pip install -e .[test]
```
RonnyPfannschmidt, line 235: random thought, would it help to bind mount the pip cache of the invoking user?

physikerwelt: Yes, some people do that. https://stackoverflow.com/a/58021389
However, is this specific to setuptools? Maybe one can link general recommendations for working with docker and python. I think both options, mounting .git and using the pretend version environment variables are specific for setuptools. The rest depends on the specific use case with pros and cons.
The line in https://github.com/pypa/setuptools_scm/issues/568#issue-896566554 is to my feeling the important bit to share, and effective if one does not care about waiting for pip install to complete. If you like, feel free to remove the second part with the pretend version altogether. I am uncertain, if this is still in the scope of #568.

RonnyPfannschmidt: thanks, im outside for a while longer then i'l ltake a look at mergin this in the evening

physikerwelt:
@RonnyPfannschmidt  Do you need any input from me? Otherwise, I would remove this PR from my watch list.

RonnyPfannschmidt:
finally got around giving this a look over, thanks :+1:

PR #663, opened 2022-01-06
Add feature to ignore dirty state

adrien-berchet:
In some cases I would like to be able to ignore the dirty state of a git repository when guessing the new version. For example, it would solve the following issue with RTD: https://github.com/readthedocs/readthedocs.org/issues/8201 (we would just have to set the environment variable in the RTD settings).
WDYT?

### DIFF
--- a/src/setuptools_scm/__init__.py
+++ b/src/setuptools_scm/__init__.py
@@ -132,6 +132,7 @@ def get_version(
     version_cls=None,
     normalize=True,
     search_parent_directories=False,
+    ignore_dirty=False,
 ):
     """
     If supplied, relative_to should be a file from which root may
--- a/src/setuptools_scm/__init__.py
+++ b/src/setuptools_scm/__init__.py
@@ -132,7 +132,6 @@ def get_version(
     version_cls=None,
     normalize=True,
     search_parent_directories=False,
-    ignore_dirty=False,
 ):
     """
     If supplied, relative_to should be a file from which root may
--- a/src/setuptools_scm/config.py
+++ b/src/setuptools_scm/config.py
@@ -108,7 +108,6 @@ class Configuration:
         self.dist_name = dist_name
         self.search_parent_directories = search_parent_directories
         self.parent = None
-        self.ignore_dirty = ignore_dirty
 
         if not normalize:
             # `normalize = False` means `version_cls = NonNormalizedVersion`
--- a/src/setuptools_scm/config.py
+++ b/src/setuptools_scm/config.py
@@ -108,6 +108,7 @@ class Configuration:
         self.dist_name = dist_name
         self.search_parent_directories = search_parent_directories
         self.parent = None
+        self.ignore_dirty = ignore_dirty
 
         if not normalize:
             # `normalize = False` means `version_cls = NonNormalizedVersion`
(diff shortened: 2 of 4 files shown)


RonnyPfannschmidt:
hi, first thank you for the contribution,

i appreciate the idea, its a little unfortunate you spend time on a **very well done** implementation as we have to coordinate a bit to integrate the idea in the general overall plan for providing configurable overrides to setuptools_scm

i will try to provide more details by either tonight or the weekend

adrien-berchet:
Hi @RonnyPfannschmidt 
Thank you for your feedback! I didn't know there was such plan but it's good to know. I will read the details then :)
And if it is relevant to integrate this PR in the plan I can take more time to update it accordingly.

RonnyPfannschmidt:
@adrien-berchet at the higher level i want to provide a feature for setuptools_scm to override config and/or version fields from env_vars, and specifically per matching distribution name 

the env vars would be `SETUPTOOLS_SCM_OVERRIDE_CONFIG/VERSION_FOR_$PKGNAME`

so the use-case you proposed here would be implemented by something like `SETUPTOOLS_SCM_OVERRIDE_VERSION_FOR_MYPKGNAME="dirty=false"`

im  currently not yet sure which syntax to use

adrien-berchet:
Ok, I think I see.

In my case shouldn't it be `CONFIG` instead of `VERSION`? So for the package `MYPKGNAME` it would be `SETUPTOOLS_SCM_OVERRIDE_CONFIG_FOR_MYPKGNAME = 'ignore_dirty=true'`? Or should we also change the `CONFIG / VERSION` part into something like `SETUPTOOLS_SCM_OVERRIDE_IGNORE_DIRTY_FOR_MYPKGNAME = true`? The latter is probably better so the user can override several variables easily.

And then we have to iterate over all the environment variables to find the ones that start with `SETUPTOOLS_SCM_OVERRIDE_VERSION_FOR_` (or `SETUPTOOLS_SCM_OVERRIDE_CONFIG`) and then check that the end of this env is equal to the current package name, right? Do you already have a function to do this or should I write one?

RonnyPfannschmidt:
@adrien-berchet what i meant was, that with the override for version,, the ignore_dirty would no longer be necessary as the override of dirty would enforce it to be false

RonnyPfannschmidt:
currently there is some example code for the code to use the per package version override

i was planning to get to implement the override before the end of the week

adrien-berchet:
> @adrien-berchet what i meant was, that with the override for version,, the ignore_dirty would no longer be necessary as the override of dirty would enforce it to be false

Ahhh ok ok, that makes more sense indeed.

> i was planning to get to implement the override before the end of the week

Ok, I am going to wait until it gets implemented and use it, so we avoid changing things several times.
I will come back next week to check (no pressure ;-P).

Thanks!

adrien-berchet:
Hi @RonnyPfannschmidt 
I pushed a new version that should be much closer to your overriding convention (I was inspired by `_overrides._read_pretended_version_for`).
For now I could not put the `_read_dirty_for` function into the `_overrides.py` file because of circular imports (because `version.meta` is imported in `_overrides.py`) but we can improve this if you prefer.

RonnyPfannschmidt:
im just slowly getting back to things after the covid quarantaine with the toddler is over

adrien-berchet:
> im just slowly getting back to things after the covid quarantaine with the toddler is over

Oh ok, I hope everything is fine!

RonnyPfannschmidt:
https://github.com/pypa/setuptools_scm/pull/800 introduced generalized overrides which will support leaving out dirty state if necessary

adrien-berchet:
Ah nice, thanks!

PR #675, opened 2022-01-28
Stated which version of setuptools is required in the error message

couling:
At the moment the error message is very confusing if you have `setuptools>=42,<45`.  The message only mentions needing version `>= 42` but is triggered if `< 45`.  I don't know why `>= 45` is required. but at least we can state It's required in the error message.


RonnyPfannschmidt (approved):
Thanks

PR #676, opened 2022-01-28
Fix setuptools_scm installation instruction for pyproject.toml

asvetlov:
I believe, `[toml]` extra dependency is crucial to make the recipe work.
At least I was unable to setup SCM using `pyproject.toml` only (without mentioning `setuptools_scm` in `setup.py`/`setup.cfg`).


RonnyPfannschmidt:
Thanks

PR #678, opened 2022-02-01
Fix for folks using black.

hameerabbasi:
Due to the repr format specifier used in the Python template, strings use
single-quotes instead of double-quotes. `black` really seems to hate those. Example generated file:

```python
# coding: utf-8
version = '0.8.8.dev2+gaa94c5a'
version_tuple = (0, 8, 8, 'dev2', 'gaa94c5a')
```

This adds a line that will cause `black` to ignore the file completely. Unfortunately, I couldn't find a quick way to replace the single-quotes reliably.


henryiii:
FYI, if you run black via pre-commit (the app, not the git hook), it will only run on version tracked files. You can also specify ignore files in your config file, I think - but I always use pre-commit.

hameerabbasi:
I use `pytest-black` which doesn't have a way to ignore files I believe.

henryiii:
It should respect pyproject.toml config? You can ignore there.

henryiii:
(Note that this pr is bad, there already is an unneeded in Python 3 utf-8 comment there.

hameerabbasi:
> It should respect pyproject.toml config? You can ignore there.

I'll try this and report back.

hameerabbasi:
It seems to work nicely, thanks!

Quansight-Labs/uarray#272

PR #681, opened 2022-02-07
Suppress GPG signature when getting HEAD committer date (#548)

### LINKED ISSUE #548: GPG Signature causes: Preparing wheel metadata: finished with status 'error'
**By creating this issue, I want to create awareness about the following behavior:** 

I came across the following Error: pre-commit fails when setting up the black environment.

```
          ...
          File "/private/var/folders/zs/ydvqpbz95pb3_l60_2vk2k0h0000gn/T/pip-build-env-p1l6kgqx/overlay/lib/python3.9/site-packages/setuptools_scm/git.py", line 67, in get_head_date
            return datetime.strptime(date_part, r"%Y-%m-%d").date()
          File "/usr/local/opt/asdf/installs/python/3.9.2/lib/python3.9/_strptime.py", line 568, in _strptime_datetime
            tt, fraction, gmtoff_fraction = _strptime(data_string, format)
          File "/usr/local/opt/asdf/installs/python/3.9.2/lib/python3.9/_strptime.py", line 349, in _strptime
            raise ValueError("time data %r does not match format %r" %
        ValueError: time data 'gpg: Signature made Wed Aug 26 17:50:45 2020 CES' does not match format '%Y-%m-%d'
```
(see [Black Issue 2059](https://github.com/psf/black/issues/2059))

With help from ichard26, I solved the problem by setting `log.showSignature = False` in my gitconfig.

In my optinion it is a good idea to avoid this problem in the future, by making the library independed of showSignature.

puxlit:
When users have `log.showSignature` enabled, `get_head_date` fails with `ValueError: time data 'gpg: Signature made [...]' does not match format '%Y-%m-%d'`, as observed in #548, #621, psf/black#2059, and so on.

Short of rewriting `get_head_date` to use plumbing commands, we can account for this edge case for git v2.10 and higher (see git/git@369dc4081c836bc17ee1debaf6688eb098359760) with either `git log --no-show-signature` or `git -c log.showSignature=false log`. (The latter has the advantage of not failing due to an unrecognised argument for earlier versions of git.)

This isn't a comprehensive fix, but is hopefully an improvement.

Closes #682, fixes #548.

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -50,11 +50,16 @@ jobs:
         # environment variable is too long. Consequently, we need to shorten PATH to
         # something minimal before we can install GnuPG. For further details, see
         # <https://github.com/actions/virtual-environments/issues/2876>.
+        #
+        # Additionally, we'll explicitly set `gpg.program` to ensure Git for Windows
+        # doesn't invoke the bundled GnuPG, otherwise we'll run into
+        # <https://dev.gnupg.org/T5504>. See also: <https://dev.gnupg.org/T3020>.
         run: |
-          $env:PATH = "C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\ProgramData\Chocolatey\bin"
+          $env:PATH = "C:\Program Files\Git\bin;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\ProgramData\Chocolatey\bin"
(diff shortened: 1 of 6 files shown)


RonnyPfannschmidt:
Do you have a idea for a tesrcase  for this one

puxlit:
Hmm, how would you feel about adding GnuPG as a test dependency? :badpokerface:

RonnyPfannschmidt:
Sounds acceptable, we should xfail the test in case it's missing tho

RonnyPfannschmidt:
excellent and detailed work, thanks !!

i plan to merge and release later today

 would you like to add the changelog entry?

puxlit:
No worries, thank you too! I've given the changelog entry a stab. 🙂

PR #689, opened 2022-02-21
Stop suggesting to put wheel in `pyproject.toml`

webknjaz:
This is a known misconception that it's necessary. It's actually been
automatically exposed by `setuptools.build_meta` since it got
introduced.

Ref: https://github.com/pypa/setuptools/pull/3056.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,7 +41,7 @@ built step by specifying it as one of the build requirements.
 
     # pyproject.toml
     [build-system]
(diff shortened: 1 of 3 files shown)


webknjaz:
@RonnyPfannschmidt that single failing CI check is caused by the deprecation of a PyPy interpreter version:

> Error: PyPy 3.6 not found

webknjaz:
> > Error: PyPy 3.6 not found

https://github.com/pypa/setuptools_scm/pull/690 should fix that.

RonnyPfannschmidt:
thanks!

PR #690, opened 2022-02-21
Use `pypy-3.x` notation in GHA workflows

webknjaz:
The old `pypy3` syntax has been deprecated IIRC.


webknjaz:
I think this could fix https://github.com/pypa/setuptools_scm/pull/689#issuecomment-1047158269.

RonnyPfannschmidt:
thanks!

PR #702, opened 2022-04-08
Delay setting default Configuration root

### LINKED ISSUE #691: Using 'root' in `pyproject.toml` file raises a TypeError when invoked from `__main__.py`
Hello,

I'm using the following versions

    setuptools          50.3.2
    setuptools-scm      6.4.2

I've a `pyproject.toml` file which setuptools_scm section is

       [tool.setuptools_scm]                                                                                                                                                                                                            
       write_to = "_version.txt"                                                                                                                                                                                                  
       root = "../.."                                                                                                                                                                                                                   
       relative_to = "__file__" 

When used with `python3 -m build` it works. However,  when I run just tool to see the output version, I get the following

    $ python -m setuptools_scm       
      File "<some-path>/lib/python3.7/runpy.py", line 193, in _run_module_as_main
        "__main__", mod_spec)
      File "<some-path>/lib/python3.7/runpy.py", line 85, in _run_code
        exec(code, run_globals)
      File "<some-path>/lib/python3.7/site-packages/setuptools_scm/__main__.py", line 83, in <module>
        main()
      File "<some-path>/lib/python3.7/site-packages/setuptools_scm/__main__.py", line 18, in main
        config = Configuration.from_file(pyproject, root=root)
      File "<some-path>/lib/python3.7/site-packages/setuptools_scm/config.py", line 216, in from_file
        return cls(dist_name=dist_name, **section, **kwargs)
    TypeError: type object got multiple values for keyword argument 'root'

The variable "root" get defined twice, first by `__main__.py` when setting the default 'root' or the one from the command line (`--root`); and the second time by the user (me) when I add 'root' to my `pyproject.toml` file

GDYendell:
I have experienced the same issue reported in #691. This change is intended to fix it.

As far as I can tell, in projects where `pyproject.toml` is not in the root, `pip install` will not work without `root` (and `relative_to`) defined in the `pyproject.toml`, but if it is defined in `pyproject.toml` then `python -m setuptools_scm` will not work - it gives a duplicate keyword argument error:

```
▶ python -m setuptools_scm 
  File "/dls_sw/prod/tools/RHEL7-x86_64/Python3/3-7-2dls1/prefix/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/dls_sw/prod/tools/RHEL7-x86_64/Python3/3-7-2dls1/prefix/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/scratch/development/setuptools_scm/src/setuptools_scm/__main__.py", line 83, in <module>
    main()
  File "/scratch/development/setuptools_scm/src/setuptools_scm/__main__.py", line 18, in main
    config = Configuration.from_file(pyproject, root=root)
  File "/scratch/development/setuptools_scm/src/setuptools_scm/config.py", line 216, in from_file
    return cls(dist_name=dist_name, **section, **kwargs)
TypeError: type object got multiple values for keyword argument 'root'
```

The problem is that it falls back to the setting root to be the same directory as `pyproject.toml` if `--root` is not passed as an option and passes that into `Configuration.from_file`, so `root` appears in both `**section` and `**kwargs` in `cls()` call.

This change moves the setting of this default down into `Configuration.from_file`, so that it can check if it already has a root from the config file. The effect is that `--root` will supercede the config file, but if `--root` is not passed and it exists in the config file, it will use that (where currently it throws an error). The behaviour of using `--root` (or `"."` if it is not given) to search for the config file and to create the default Configuration is maintained.

Fixes #691

### DIFF
--- a/src/setuptools_scm/__main__.py
+++ b/src/setuptools_scm/__main__.py
@@ -10,12 +10,10 @@ from setuptools_scm.integration import find_files
 
 def main() -> None:
     opts = _get_cli_opts()
-    root = opts.root or "."
 
     try:
-        pyproject = opts.config or _find_pyproject(root)
-        root = opts.root or os.path.relpath(os.path.dirname(pyproject))
-        config = Configuration.from_file(pyproject, root=root)
+        pyproject = opts.config or _find_pyproject(opts.root or ".")
+        config = Configuration.from_file(pyproject, root=opts.root)
     except (LookupError, FileNotFoundError) as ex:
         # no pyproject.toml OR no [tool.setuptools_scm]
         print(
@@ -24,7 +22,7 @@ def main() -> None:
             f" Reason: {ex}.",
             file=sys.stderr,
         )
-        config = Configuration(root=root)
+        config = Configuration(root=opts.root or ".")
 
     version = _get_version(config)
     assert version is not None
--- a/src/setuptools_scm/config.py
+++ b/src/setuptools_scm/config.py
@@ -179,7 +179,8 @@ class Configuration:
     @classmethod
     def from_file(
         cls,
-        name: str = "pyproject.toml",
+        name: str = "./pyproject.toml",
+        root = None,
         dist_name=None,  # type: str | None
         _load_toml=_lazy_tomli_load,
         **kwargs,
@@ -212,6 +213,14 @@ class Configuration:
                 dist_name = defn["project"].get("name")
         if dist_name is None:
             dist_name = _read_dist_name_from_setup_cfg()
+
+        if root is not None:
+            # Supercede config file root with parameter
+            section["root"] = root
+        else:
+            if "root" not in section:
+                # Set root to directory of pyproject.toml
+                section["root"] = os.path.relpath(os.path.dirname(name))
 
         return cls(dist_name=dist_name, **section, **kwargs)
 
--- a/src/setuptools_scm/config.py
+++ b/src/setuptools_scm/config.py
@@ -180,7 +180,7 @@ class Configuration:
     def from_file(
         cls,
         name: str = "./pyproject.toml",
-        root = None,
+        root=None,
         dist_name=None,  # type: str | None
         _load_toml=_lazy_tomli_load,
         **kwargs,


RonnyPfannschmidt (changes requested):
at first glance this looks good, i suspect the the new rootfinding fails in certain cases on windows  due to drive differences

baodrate:
FYI @GDYendell, I fixed the tests and added handling for a previously neglected edge case (`root` specified in `pyproject.toml` should be relative to config file). I made a PR against your branch here:

https://github.com/GDYendell/setuptools_scm/pull/1

tests are passing. You can see the results on my fork:

https://github.com/qubidt/setuptools_scm/actions/runs/2191926795

GDYendell:
Thanks @qubidt !

@RonnyPfannschmidt shall I merge, or perhaps just cherry pick, these additional changes or do you want to take a look at it separately first?

RonnyPfannschmidt:
I just returned from a 2 week vacation and haven't gotten to this yet in detail

I like the general idea, but have to verify some details on how to integrate things (the details around creating config objects are quite more messy than I'd like and your changes made that even more apparent)

baodrate:
@RonnyPfannschmidt any update on considering this PR or an alternative solution to the issue?

RonnyPfannschmidt:
I'm currently just not able to commit as much time to setuptools _acm as I'd like to

RonnyPfannschmidt:
i finally did the review, indeed when root is overridden, then it should be ignored/not passed down

the approach implemented here unfortunately is not entirely correct, but the deferring a a a concept is the right idea

RonnyPfannschmidt:
i cant reiterate this for the next release, but this is the next change i will sort out

RonnyPfannschmidt:
i started to recreate the behaviour starting with creating unitests for the cli - i hope i can complete that pr today,

RonnyPfannschmidt:
closing in favor of #731 which was heavily inspired by this one, but also adds tests and a warning for root enforcement (cli root superseeding config root)

thanks for providing this initial version which greatly cut down the time i needed to understand the problem

GDYendell:
Thanks @RonnyPfannschmidt ! :star:

Issue #707, opened 2022-04-13
`fatal: unsafe repository ('...' is owned by someone else)` and impact of CVE-2022-24765 fix for `git`


KOLANICH:
Hi. Today I have noticed that editable installation of packages from root stopped working. It is because of git version with fix for [`CVE-2022-24765`](https://github.blog/2022-04-12-git-security-vulnerability-announced/) has been released and delivered.

It writes that `LookupError: setuptools-scm was unable to detect version for ...`.

Investigation and adding dumping resulted in pinpointing the error to `.git.from_potential_worktree`, where `git` errors with

```
fatal: unsafe repository ('...' is owned by someone else)

To add an exception for this directory, call:
	git config --global --add safe.directory
```

.

1. I know that editable installs from root are insecure and that just adding files to such dirs allows privelege escallation.
2. I know that `pip` must not execute wheel building elevated even if called from root, it should drop its rights for the time it builds wheel, and it is a flaw in `pip` it doesn't do that.

But those are not flaws in `setuptools_scm`. Let's concentrate on the flaws on `setuptools_scm` related to the issue.

* #705
* #706

KOLANICH:
The workaround to this issue is to use `sudo fakeroot` instead of `sudo`. `fakeroot` makes `git` think that the repo is owned by `root`, and so the security check is satisfied.

pitrou:
I suspect many people are affected by this. It happens for example when running a Docker container where the source tree is mounted from the host.

pitrou:
@henryiii @RonnyPfannschmidt What would be your take on this?

RonnyPfannschmidt:
@pitrou thanks for the linked issue,

I believe it may be good/necessary to pass that config option to git

I'll make it a Fokus roughly mid next week after completing git archive support

KOLANICH:
>I believe it may be good/necessary to pass that config option to git

The doc says that config option takes no effect when passed through command line. Modifying configs is IMHO inacceptable. IMHO the solution with `fakeroot` and/or using bindings to more flexible git libs is more optimal.

pitrou:
@KOLANICH Can you point to the doc that says that?

pitrou:
Ok, found it: https://git-scm.com/docs/git-config/2.36.0#Documentation/git-config.txt-safedirectory

pelson:
[...]

PR #708, opened 2022-04-25
Use --git-dir to avoid issues with CVE-2022-24765 mitigation

chrisburr:
This is a possible solution to https://github.com/pypa/setuptools_scm/issues/707 by explicitly passing `--git-dir=$PWD/.git` (which bypasses auto-discovery and therefore the vulnerability).

I've tried to add a test for this in b8dcd7a25d00443495bc91037eb6e5d9a8796c43. Obviously it has to change ownership to a non-existant user so it needs to use `sudo` but I don't particularly like it. I can refactor if you have better ideas or drop the test entirely if you'd prefer.

### DIFF
--- a/src/setuptools_scm/git.py
+++ b/src/setuptools_scm/git.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from os.path import isfile
 from os.path import join
 from os.path import samefile
+from shlex import quote
 
 from .config import Configuration
 from .scm_workdir import Workdir
@@ -36,7 +37,11 @@ class GitWorkdir(Workdir):
     def from_potential_worktree(cls, wd):
         require_command(cls.COMMAND)
         wd = os.path.abspath(wd)
-        real_wd, _, ret = do_ex("git rev-parse --show-prefix", wd)
+        git_dir = join(wd, ".git")
+        real_wd, _, ret = do_ex(
+            f"git --git-dir={quote(git_dir)} rev-parse --show-prefix",
+            wd,
+        )
         real_wd = real_wd[:-1]  # remove the trailing pathsep
         if ret:
             return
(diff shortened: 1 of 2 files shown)


src/setuptools_scm/git.py
```diff
@@ -54,23 +59,35 @@ def from_potential_worktree(cls, wd):
...
         return cls(real_wd)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._git_dir = join(self.path, ".git")
+
     def is_dirty(self):
-        out, _, _ = self.do_ex("git status --porcelain --untracked-files=no")
+        out, _, _ = self.do_ex(
+            f"git --git-dir={quote(self._git_dir)} "
```
RonnyPfannschmidt: Let's use the list form here

testing/test_git.py
```diff
@@ -92,6 +94,42 @@ def test_parse_call_order(wd):
     git.parse(str(wd.cwd), git.DEFAULT_DESCRIBE)
 
 
+@pytest.mark.issue("https://github.com/pypa/setuptools_scm/issues/707")
+def test_not_owner(wd):
```
RonnyPfannschmidt, line 99: Let's mark this as xfail run=False for now

RonnyPfannschmidt (commented):
If the parameters are needed everywhere, let's consider a helper method that calls git subcommands

henryiii:
> If the parameters are needed everywhere, let's consider a helper method that calls git subcommands

I almost suggested this! Something like `self._do_git(*cmds)`. FYI, I've used this method to fix a CMakeLists.txt git usage (unrelated to setuptools_scm) in GooFit after seeing it here. Thanks!

chrisburr:
Thanks for the review @RonnyPfannschmidt.

I think these changes have also fixed the issue with the Windows CI in [2e54cc2](https://github.com/pypa/setuptools_scm/pull/708/commits/2e54cc2baea1c707147d416dbd21910b4c89f98b).

henryiii:
Gentle reminder, this is broken on some systems using newer git until this patch is made. (cibuildwheel should be fine by matching ownership, but multibuild still has mismatched owners).

RonnyPfannschmidt:
i just merged this, will be part of the next release

PR #719, opened 2022-06-21
Fix Missing importlib_metadata in <=3.7

### LINKED ISSUE #718: setuptools_scm 7.0.0 fails with Python 3.7
When using `setuptools_scm` in a CI pipeline to install a package I'm getting:

```
ModuleNotFoundError: No module named 'importlib.metadata'
```
compare https://github.com/audeering/audonnx/runs/6985662108?check_suite_focus=true

which is a common `importlib` bug described for example at https://github.com/newville/asteval/issues/93

kojiromike:
Fixes #718

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,8 @@
+v7.0.1
+=======
+
+* fix #718: Avoid `ModuleNotFoundError` by requiring importlib_metadata in python < 3.8
+
 v7.0.0
 =======
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,6 +31,7 @@ install_requires =
     setuptools
     tomli>=1.0.0 # keep in sync
     typing-extensions
+    importlib-metadata;python_version < '3.8'
 python_requires = >=3.7
 package_dir =
     =src


RonnyPfannschmidt:
this looks like the fix,

i need to create a way to test this type of possible failure

kojiromike:
@RonnyPfannschmidt I have a local dockerfile that I just finished testing this fix in; however, it will take me a little while to figure out why the 3.7 tests in this repo didn't already catch the error.

kojiromike:
Pushing a hypothetical change that might fail in 3.7 to show if tests work.

kojiromike:
@RonnyPfannschmidt I think this is ready for testing now. It will fail in <3.8, correctly, if you remove the conditional requirement. I didn't want to spam this PR too much, but you can see that failure over here: https://github.com/kojiromike/setuptools_scm/actions/runs/2536366008 (I canceled the workflow after the necessary tests ran.)

cesarlr (approved):

kojiromike:
Still no good on the testing front. I guess a PEP517 build will use the pyproject requirements, including importlib_metadata and thus passing, even in python 3.7. But it seems a non-PEP517 build using pip or `setup.py` will not install the setup dependencies before trying to run `setup.py`, where it all falls apart.

kojiromike:
I've put things back to just the fix, as I'm unable to find a way to automatically test right now.

cesarlr (approved):

RonnyPfannschmidt:
Lovely, thanks for testing out the details, if you add a changelog entry i can directly go for merge +release after dinner

kojiromike:
@RonnyPfannschmidt done, Genieße dein Abendessen

PR #729, opened 2022-06-24
Hg / pip compatibility

paugier:
I open this PR to discuss about these changes that would avoid the problem describe in https://github.com/pypa/pip/issues/10635.

This is a now a draft implementation of solution 2 described in https://github.com/pypa/pip/issues/10635#issuecomment-966973120.

I see that there are already a function to clean up the environment of Git specific environment variables. A possibility would be to merge my new function `avoid_pip_isolation` and `no_git_env` to one unique function cleaning the environment to call the programs.

I wonder if it would be necessary to have a mechanism to deactivate this cleaning?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,5 +2,6 @@
 requires = [
     "setuptools>=45",
     "packaging>=20.0",
+    "typing_extensions",
 ]
 build-backend = "setuptools.build_meta"
--- a/src/setuptools_scm/utils.py
+++ b/src/setuptools_scm/utils.py
@@ -63,7 +63,7 @@ def ensure_stripped_str(str_or_bytes: str | bytes) -> str:
         return str_or_bytes.decode("utf-8", "surrogateescape").strip()
 
 
-def _run(cmd: _t.CMD_TYPE, cwd: _t.PathT) -> subprocess.CompletedProcess[bytes]:
+def _run(cmd: _t.CMD_TYPE, cwd: _t.PathT) -> subprocess.CompletedProcess[str]:
     return subprocess.run(
         cmd,
         capture_output=True,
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -2,5 +2,6 @@
 requires = [
     "setuptools>=45",
     "packaging>=20.0",
+    "typing_extensions",
```
paugier, line 5: I had to add this to avoid a bug in the installation of setuptools_scm (Python 3.9).

paugier:
The first commit is actually only about improving debug logging.

src/setuptools_scm/utils.py
```diff
@@ -63,43 +84,48 @@ def ensure_stripped_str(str_or_bytes: str | bytes) -> str:
...
         env=dict(
-            no_git_env(os.environ),
+            avoid_pip_isolation(no_git_env(os.environ)),
             # os.environ,
             # try to disable i18n
             LC_ALL="C",
             LANGUAGE="",
             HGPLAIN="1",
         ),
+        text=True,
```
RonnyPfannschmidt, line 103: Oh, seems like I missed that opportunity before :)

src/setuptools_scm/utils.py
```diff
@@ -63,43 +84,48 @@ def ensure_stripped_str(str_or_bytes: str | bytes) -> str:
...
     )
 
 
 def do_ex(cmd: _t.CMD_TYPE, cwd: _t.PathT = ".") -> _CmdResult:
-    trace("cmd", repr(cmd))
-    trace(" in", cwd)
+    if isinstance(cmd, list):
+        cmd_4_trace = " ".join(cmd)
```
RonnyPfannschmidt: This needs to use shlex quoting to avoid commands one can't safely run copy pasted

src/setuptools_scm/utils.py
```diff
@@ -63,43 +84,48 @@ def ensure_stripped_str(str_or_bytes: str | bytes) -> str:
...
+        cmd_4_trace = cmd
+    trace("----\ncmd:", repr(cmd_4_trace))
+    trace(" in:", cwd)
     if os.name == "posix" and not isinstance(cmd, (list, tuple)):
         cmd = shlex.split(cmd)
 
     res = _run(cmd, cwd)
     if res.stdout:
-        trace("out", repr(res.stdout))
+        trace("out:\n", res.stdout)
```
RonnyPfannschmidt: Ideally those add a indent level

RonnyPfannschmidt (changes requested):
I like what I see there

A good opt out for the env cleanup is unclear to me as well

I think it's a good idea to keep pip and git cleanup in separate functions

pyproject.toml
RonnyPfannschmidt: is this python 3.9 only or python 3.9+?

paugier: I tried only with 3.9 and I had an import error without this line. However, I don't understand well how we are supposed to use [typing_extensions](https://pypi.org/project/typing-extensions/), since "The typing_extensions module serves two related purposes". I don't know if it is also useful for python > 3.9. It seems that it depends on what exactly is used in the code.

RonnyPfannschmidt (approved):
looks good to me thanks

anything you want to add, or should i aim for a swift release?

pyproject.toml
RonnyPfannschmidt: @paugier ah, its my fault - i missed to make the _t import in hg/hg_git support optional, if you like you can replicate what i did for git,

PR #736, opened 2022-06-28
cleanup pyproject loading and allow cli relative roots to be specified

RonnyPfannschmidt:
this also tries to address concerns by @qubidt in #691 / #702

please take a look if this resolves the issue

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,8 @@ v7.0.4
 
 * fix #727: correctly handle incomplete archivals from setuptools_scm_git_archival
 * fix #691: correctly handle specifying root in pyproject.toml
+* correct root override check condition (to ensure absolute path matching)
+* allow root by the cli to be considered relative to the cli (using abspath)
 
 v7.0.3
 =======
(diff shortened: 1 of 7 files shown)


baodrate (approved):
looks good! passes all the cases I can come up with

RonnyPfannschmidt:
@qubidt can you share those, I'd like to have them part of the testsuite

baodrate:
I only wrote a few, you can see them in the PR I made against the original #702:

https://github.com/GDYendell/setuptools_scm/pull/1/files#diff-6f46d110c3d6f83624aca4044339075d63fdb10acd2b1f0c2ade98204c0a34ec

Of which one (`test_repo_cmdline_root_relative_cwd` I *think*) fails against the current master but passes with this PR. If you want to add them to the feel free to copy them. Don't go out of your way to merge my commit or anything, lol

most of the other cases I was referring to were just manual testing I did w/ variations of `.`/`..`/relative/linked paths for `-r` and `-c`. Nothing particularly interesting or useful

RonnyPfannschmidt:
thanks for the update, time for merge and release then

PR #744, opened 2022-07-09
Fix #743: Allow entrypoints in `parse_scm`

### LINKED ISSUE #743: There is no way to use a custom "parse" function with `pyproject.toml`-based configuration
Since `pyproject.toml` can't import Python functions directly

```
[tool.setuptools_scm]
use_scm = { parse="some_name" }
```

tries to use the string `"some_name"` as a function, causing a traceback. If the string is defined as the name of an entrypoint of `setuptools_scm.parse`, it might make sense to use it instead. Does this make sense? Would a PR be useful?

moshez:
Unit test  added.

Full test:

```
$ git tag
1.0.0
$ python -m build --no-isolation --sdist|tail -1
...
Successfully built UNKNOWN-1.0.0.tar.gz
$  cat pyproject.toml 
[build-system]
requires = ["setuptools", "setuptools_scm"]
build-backend = "setuptools.build_meta"

[tool.setuptools_scm]
parse = ".git"
```

### DIFF
--- a/src/setuptools_scm/__init__.py
+++ b/src/setuptools_scm/__init__.py
@@ -82,7 +82,7 @@ def dump_version(
 
 
 def _get_parse_function(
-    parse: Union[MaybeConfigFunction, str],
+    parse: MaybeConfigFunction | str,
 ) -> MaybeConfigFunction:
     if callable(parse):
         return parse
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt:
practically with this people to publish plugins as they couldn't spell out their parse functions as local import

RonnyPfannschmidt:
closing this - after closing the issue aready - the correct way with setuptools to pass code is in setup.py

PR #748, opened 2022-07-11
Use stdlib tomllib instead of tomli for Python 3.11+

### LINKED ISSUE #740: Use tomllib instead of tomli when present
tomli got into standard library with 3.11. It is no longer necessary for this package to depend on tomli when on 3.11 or newer. Please make the dependency conditional and use tomllib when available.

hugovk:
Fixes https://github.com/pypa/setuptools_scm/issues/740.


RonnyPfannschmidt:
lovely, thanks !

PR #757, opened 2022-08-05
tests: Fix test_git_archhival_from_unfiltered assumptions

### LINKED ISSUE #750: Buggy last test in 7.0.5's `testing/test_git.py` ?
While packaging 7.0.5 for Debian, I found the last test of `testing/test_git.py` a bit strange:
1. it's named `test_git_archhival_from_unfiltered` -- why two h ?
2. it tries to load a file which isn't currently installed by the package... and hence fails! Either that file is mandatory and it should be installed or optional and the test should handle it.

Cheers,

J.Puydt

stanislavlevin:
This test relied on several things:

- editable installation and thereby, '.git_archival.txt' is always
  accessible. But some Linux distros (e.g. Debian) run tests out
  of a git tree.
- no expansion happens for '.git_archival.txt' (see .gitattributes),
  i.e. the packaged sources are not produced by git archive.

Fixes: https://github.com/pypa/setuptools_scm/issues/750

### DIFF
--- a/testing/test_git.py
+++ b/testing/test_git.py
@@ -24,7 +24,6 @@ from setuptools_scm import integration
 from setuptools_scm import NonNormalizedVersion
 from setuptools_scm.file_finder_git import git_find_files
 from setuptools_scm.git import archival_to_version
-from setuptools_scm.utils import data_from_mime
 from setuptools_scm.utils import do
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
Thanks!

PR #762, opened 2022-08-19
Support UTF-8 in setup.cfg files

jaraco:


src/setuptools_scm/_integration/setuptools.py
```diff
@@ -14,7 +14,7 @@ def read_dist_name_from_setup_cfg(
     parser = configparser.ConfigParser()
 
     if isinstance(input, (os.PathLike, str)):
-        parser.read([input])
+        parser.read([input], encoding="utf-8")
```
RonnyPfannschmidt, line 17: based on encoding behaviour, would it make sense to use 

```
with open(input, encoding="utf-8", errors="replace") as fp:
   parser.read_file(fp)
```

RonnyPfannschmidt (approved):
nice catch, thanks

would like to get feedback on the question before merge

src/setuptools_scm/_integration/setuptools.py
jaraco: In my opinion, it's better to use defaults unless there's a reason not to use the defaults, and I'm not aware of any reason. And since [Setuptools doesn't replace](https://github.com/pypa/setuptools/blob/feed34f981cd7035d787e1cb4825ea58a423152d/setuptools/dist.py#L712), neither should `setuptools_scm`. If there are users currently depending on the current behavior on Windows only, their project wouldn't install with Setuptools anyway, so the simpler form is best.

jaraco:
Are there any plans to release this change? It's still blocking https://github.com/jaraco/configparser/issues/58.

RonnyPfannschmidt:
I hope to get to it in the next week, I'm slightly swamped atm

jaraco:
It looks like I have access to cut releases. I'll examine the project and see if I have enough context to cut a release.

PR #765, opened 2022-08-29
Remove default "coding: utf-8" in .py template

mwtoews:
This PR removes the first line of the `.py` template `# coding: utf-8`.

As of Python 3.0, the default encoding Python source files is UTF-8 ([PEP 3120](https://peps.python.org/pep-3120/)).


RonnyPfannschmidt (approved):
thanks!

PR #769, opened 2022-10-11
Rename license_file to license_files

jelmer:
Rename license_file to license_files

This fixes a deprecation warning.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,7 +7,7 @@ url = https://github.com/pypa/setuptools_scm/
 author = Ronny Pfannschmidt
 author_email = opensource@ronnypfannschmidt.de
 license = MIT
-license_file = LICENSE
+license_files = LICENSE
 classifiers =
     Development Status :: 5 - Production/Stable
     Intended Audience :: Developers
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,6 +7,7 @@ url = https://github.com/pypa/setuptools_scm/
 author = Ronny Pfannschmidt
 author_email = opensource@ronnypfannschmidt.de
 license = MIT
+license_file = LICENSE
 license_files = LICENSE
 classifiers =
     Development Status :: 5 - Production/Stable


RonnyPfannschmidt (approved):

RonnyPfannschmidt:
thanks!

RonnyPfannschmidt:
@jelmer can you run pre-commit for this one, it seems like a msisconfig is currently preventing auto-push

@asottile are there any docs on how to alow pre-commit to push, i believe i missed some setting and/or a pypa policy is preventing it atm

RonnyPfannschmidt:
nm, found it

RonnyPfannschmidt:
or not :/

asottile:
@RonnyPfannschmidt hmmm the only think I could think of is if OP unchecked the "allow maintainers to push" box (which is what the run says too)

jelmer:
@RonnyPfannschmidt done, all green now :)

RonnyPfannschmidt:
@asottile i believe we just found a bug in setup-cfg-format related to a setuptools change

I can't promise to get back to that this week

asottile:
there's a bunch of dupes in the tracker, it's currently the intended behavior

RonnyPfannschmidt:
@asottile the current intended behaviour seems to be at odds with the deprecation in setuptools then, should there be some gating wrt supported versions?

asottile:
I'm going to let you read the issues first rather than rehashing everything here 👍

RonnyPfannschmidt:
@asottile oh yikes, support timelines crossing the toml warzone - would i be able to persuade you with a pr fix where it skips adding license-file if license-files is set? and/or a cli option for the hook?

asottile:
perhaps something like that -- though as written that'd break something like `babi-grammars` which uses `license_files` for vendored things.  perhaps if it's _exactly_ `license-files = {the same thing it would set license_file to currently}` then that's fine

RonnyPfannschmidt:
@asottile so my intent would be to opt out of setting license-file if its missing only if license-files is set

asottile:
yeah I think the reasonable behaviour is:

1. if `license_files` is set to the same value of what `setup-cfg-fmt` today would set `license_file` to -- do nothing
2. otherwise do the current behaviour

at a future date when the setuptools version where license-files works is old enough then change the behaviour to set-add the license file to `license_files` but I can't do that now without breaking old stuff

RonnyPfannschmidt:
with the migration to pyproject.toml this became obsolete before i got back to it

thanks for working on it

at least we got https://github.com/asottile/setup-cfg-fmt/pull/166 out of it

PR #772, opened 2022-10-14
Fix "not enough values to unpack" when parsing .git_archival.txt from a tagged commit; also fix #759

### LINKED ISSUE #759: The suggested contents of `.git_archival.txt` do not include the `--match` flag (like the default git describe)
I noticed a small discrepancy between the default git describe in `setuptools_scm` and the suggested contents of `.git_archival.txt` - the first one uses `--match *[0-9]*` while the second doesn't even though both support it:
https://github.com/pypa/setuptools_scm/blob/e1283177b23ccf254739aa8292448154c54741c8/src/setuptools_scm/git.py#L42-L43

`%(describe:tags=true,match=*[0-9]*)` can be used to get an equivalent match in `.git_archival.txt`.

I also see that the `.git_archival.txt` in the root of the repository and in `src/setuptools/.git_archival.txt` both do not include the `:tags=true` part, I'm unsure if that's intentional but it doesn't seem like this repository uses annotated tags (or at least not since v6.3.1).

dlitz:
Given `.git_archival.txt` content like this:

    node: 3634907645428b9542cfb4fd1ceef08f14526cb8
    node-date: 2022-10-13T20:38:09-06:00
    describe-name: v0.0.5
    ref-names: HEAD -> main, tag: v0.0.5

setuptools_scm fails with an exception like this:

```console
$ python3 -m setuptools_scm
  [..snip..]
  File ".../setuptools_scm/git.py", line 312, in parse_archival
    return archival_to_version(data, config=config)
  File ".../setuptools_scm/git.py", line 286, in archival_to_version
    tag, number, node, _ = _git_parse_describe(archival_describe)
  File ".../setuptools_scm/git.py", line 245, in _git_parse_describe
    tag, number, node = describe_output.rsplit("-", 2)
ValueError: not enough values to unpack (expected 3, got 1)
```

The problem here is that the describe-name field is "v0.0.5" instead of
something like "v0.0.5-0-g3634907", so the tuple unpacking fails.

So, it works correctly with untagged versions of the repo:

```console
$ tmpdir=$(mktemp -d)
$ git archive HEAD | tar -C "$tmpdir" -xf -
$ ( cd "$tmpdir" ; python3 -m setuptools_scm )
0.0.1.dev7+gb03bd8b
```

But if we tag the repo and create another archive, the exception occurs:

```console
$ git tag v0.0.1 HEAD
$ tmpdir=$(mktemp -d)
$ git archive HEAD | tar -C "$tmpdir" -xf -
$ ( cd "$tmpdir" ; python3 -m setuptools_scm )
  [..snip..]
  File "/usr/lib/python3/dist-packages/setuptools_scm/git.py", line 316, in parse_archival
    return archival_to_version(data, config=config)
  File "/usr/lib/python3/dist-packages/setuptools_scm/git.py", line 290, in archival_to_version
    tag, number, node, _ = _git_parse_describe(archival_describe)
  File "/usr/lib/python3/dist-packages/setuptools_scm/git.py", line 245, in _git_parse_describe
    tag, number, node = describe_output.rsplit("-", 2)
ValueError: not enough values to unpack (expected 3, got 1)
```

Version information:

```console
$ pip show setuptools_scm | grep ^Version
Version: 7.0.5
$ dpkg -s python3-setuptools-scm | grep ^Version
Version: 7.0.5-1
```

The first commit of the pull-request proposes a fix for that.

The second commit fixes #759.

### DIFF
--- a/.git_archival.txt
+++ b/.git_archival.txt
@@ -1,4 +1,4 @@
 node: $Format:%H$
 node-date: $Format:%cI$
-describe-name: $Format:%(describe)$
+describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
 ref-names: $Format:%D$
--- a/README.rst
+++ b/README.rst
@@ -316,7 +316,7 @@ and copy-paste this into it::
 
     node: $Format:%H$
     node-date: $Format:%cI$
-    describe-name: $Format:%(describe:tags=true)$
+    describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
     ref-names: $Format:%D$
 
 Create the ``.gitattributes`` file in the root directory of your repository
--- a/src/setuptools_scm/.git_archival.txt
+++ b/src/setuptools_scm/.git_archival.txt
@@ -1,4 +1,4 @@
 node: $Format:%H$
 node-date: $Format:%cI$
-describe-name: $Format:%(describe)$
+describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
 ref-names: $Format:%D$
--- a/src/setuptools_scm/git.py
+++ b/src/setuptools_scm/git.py
@@ -232,9 +232,11 @@ def _git_parse_inner(
     )
 
 
-def _git_parse_describe(describe_output: str) -> tuple[str, int, str, bool]:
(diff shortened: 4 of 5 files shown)


RonnyPfannschmidt (approved):
lovely thanks!

PR #783, opened 2022-11-10
Use fallback_version in repos when git/hg is not available.

### LINKED ISSUE #549: Use fallback_version if SCM isn't found
I'm in some situation (testing my package in Docker) where git isn't available, and thought I could just use `fallback_version` argument, but it turns out I still get an exception `OSError: 'git' was not found`. 

I found I can overcome this by defining the `SETUPTOOLS_SCM_PRETEND_VERSION` environment variable, but I thought it would be easier to use `fallback_version` in this case as well?

I'm not sure I understood the different cases where one should use `SETUPTOOLS_SCM_PRETEND_VERSION` or `fallback_version`?

anntzer:
Closes #549.

### DIFF
--- a/testing/test_basic_api.py
+++ b/testing/test_basic_api.py
@@ -98,6 +98,24 @@ setup(use_scm_version={"fallback_version": "12.34"})
     assert res == "12.34"
 
 
+@pytest.mark.parametrize("scm", ["git", "hg"])
+def test_fallback_scm_absent(
+    scm: str, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    monkeypatch.delenv("SETUPTOOLS_SCM_DEBUG")
+    p = tmp_path / "sub/package"
+    p.mkdir(parents=True)
+    p.joinpath("setup.py").write_text(
+        """from setuptools import setup
+setup(use_scm_version={"fallback_version": "12.34"})
+"""
+    )
+    monkeypatch.setenv("PATH", "")
+    p.joinpath(f".{scm}").mkdir()
+    res = do([sys.executable, "setup.py", "--version"], p)
+    assert res == "12.34"
+
+
 def test_empty_pretend_version(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.delenv("SETUPTOOLS_SCM_DEBUG")
     monkeypatch.setenv("SETUPTOOLS_SCM_PRETEND_VERSION", "")
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,9 @@
 
+next
+====
+
+* fallback_version is also used in repositories if the corresponding SCM command is not available
+
 v7.0.5
 =======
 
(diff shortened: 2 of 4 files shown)


testing/test_basic_api.py
```diff
@@ -98,6 +98,22 @@ def test_fallback(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     assert res == "12.34"
 
 
+def test_fallback_scm_absent(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
```
RonnyPfannschmidt: The test is deceptively incomplete as it mixes error and fallback conditions

We should trigger the fallback on error when the scm is missing

I'll have to check out the details on when to fail

RonnyPfannschmidt (changes requested):

testing/test_basic_api.py
anntzer: Sorry, it is not clear to me why you are unhappy about the test.
If it is that .git/.hg are not real repos, I can change that.  The main point of the test is actually that making PATH empty means that subprocess.Popen will not find the git/hg executables and thus fail with a FileNotFoundError; i.e. the test behaves as if git/hg are not installed.

RonnyPfannschmidt: its not the problem that its not real repos

the problem is that the tests exercises both of them at once when setuptools_scm for the normals scm finding should more directly fail

so this should be 2 tests, one or each scm

also the scm version extraction should fail with an error and be handled instead of pretending nothing was found

the errors should bubble and be handled instead of being hidden

anntzer: > so this should be 2 tests, one or each scm

Sure, I changed that, no problem.

> also the scm version extraction should fail with an error and be handled instead of pretending nothing was found
> the errors should bubble and be handled instead of being hidden

I thought that the behavior requested in #549 was that if the git/hg command is not present *and* fallback_version is set, then use fallback_version, instead of erroring out.  (Certainly, that is the behavior that *I* would prefer.  Perhaps @CharlesB2 can confirm whether that's also what he's asking for?)

RonnyPfannschmidt: the scm version finding has to fail, the failure however needs to be handled so that the fallback can happen

anntzer: AFAICT from
https://github.com/pypa/setuptools_scm/blob/a70d44b5621770a4d4868057a9a1503ab7ac952b/src/setuptools_scm/_entrypoints.py#L65-L69
returning None (or another falsy value) *is* the way for a version finder to fail.

RonnyPfannschmidt: That's for not finding a version, when stuff is broken, there's need for actual errors

anntzer: I don't understand what mechanism you want me to use, then.  Note that fallback_version is currently implemented as a normal entrypoint version finder, which just happens to have lowest priority and always work.
An alternative approach would be instead to special-case fallback_version and fall back to it if none of the other entrypoints succeed; I can do that if you prefer.

RonnyPfannschmidt: parse_scm is a different entrypoint than parse_scm_fallback

anntzer: I can see that (and can move fallback_version to the parse_scm_fallback EP if you prefer), but note again that prior to this PR nothing would catch exceptions raised by the setuptools_scm.git.parse EP and that exception would immediately bubble out; I still don't understand where you want me to handle that exception.

RonnyPfannschmidt:
Superseded BZ #901 which uses a different approach

PR #787, opened 2022-11-20
Apply refurb suggestions

DimitriPapadopoulos:
* [FURB123]: Replace `int(x)` with `x`
* [FURB131]: Replace `del x[y]` with `x.pop(y)`


RonnyPfannschmidt:
thanks

PR #792, opened 2022-12-05
Clarify pip+git install example

max-len:
The example does not clarify how to install a specific revision with the required method.


RonnyPfannschmidt (changes requested):
The updated example would break for any repo that started with or updated to the main branch name

max-len:
Well, it is an example for this specific case: The revision name is already given known, when coming from the old syntax.
IMO it should map `old syntax with $var` to `new sytnax with $var` rather than to `new syntax` (implicit var).

One might omit the revision to check out the repo's HEAD. This could be marked with square braces in the example, would you go with that?

PR #803, opened 2023-02-10
Allow Custom Version Scheme Outside setup.py (resolves #781)

### LINKED ISSUE #781: Using a custom version_scheme outside setup.py
setuptools_scm allows to use a custom version_scheme function. When not using `setup.py` this needs to be an entry-point name. However, this doesn't work logically. When the project is installed (e.g. `pip install -e .`), setuptools asks setuptools_scm for the version tp install the project. But the project is not installed at that point and therefore the entry point is not accessible. Then, the install will fail due to `assert main_version is not None`.
Can you please check how to solve this chicken-egg issue?

One idea - Support qualified names: Instead of an entry-point give the qualified name to a function and setuptools_scm will execute it

schang412:
Currently, there is no way to specify a custom version_scheme if you are not using setup.py. When provided a string that does not match the name of a valid entrypoint, the main version is assigned `None` and fails the assert. 

This patch provides a method to specify a qualified name for a function using the same syntax as an console script entrypoint:

```toml
[tool.setuptools_scm]
version_scheme = "my_module.my_file:my_custom_version_scheme_func"
local_scheme = "my_module.my_file:my_custom_local_scheme_func"
```

In order to accomplish this, when going through the available version schemes, if no version scheme is found, it treats the string as a qualified function name, and tries to load it using `importlib.metadata.EntryPoint`. If this does not lead to a valid function, it will maintain current behavior.


RonnyPfannschmidt:
Thanks for having a look at this

I'll have to investigate how that would integrate with the delayed simplify_types branch

schang412:
Looking at the branch in your fork, it seems like the changes just need to be moved to a different file. If you'd like, after merging simplify_types, I can update my PR to reflect the changes necessary.

RonnyPfannschmidt:
I'll enable that

I intend to completely change how version schemes work, so people can pick and choose them with extra configuration instead of dropping into functions most of the time

RonnyPfannschmidt:
its merged, please have a look

schang412:
Hi, I updated my branch. Please let me know if there’s any issues or anything I should change.

RonnyPfannschmidt (commented):
One item that's missing is a unit or acceptance test for the behaviour

schang412:
Sure, I can make one. Is testing/test_config.py the most appropriate place?

RonnyPfannschmidt:
Most likely for now

testing/test_version.py
```diff
@@ -214,6 +214,15 @@ def test_format_version_schemes() -> None:
     )
 
 
+def test_custom_version_schemes() -> None:
+    version = meta("1.0", config=c)
+    format_version(
+        version,
+        local_scheme="no-local-version",
+        version_scheme="setuptools_scm.version:guess_next_dev_version",
```
RonnyPfannschmidt: we should pick a non-default version scheme and check that it is actually used

schang412:
Hi, I updated it to use a non-default version scheme, and merged the latest main branch in but some tests are failing for windows now. I'm not sure why. Curiously, one of the Windows tests passed.

RonnyPfannschmidt (approved):
Thanks

PR #820, opened 2023-03-20
Makes .py output black compliant

### LINKED ISSUE #482: Auto-generated write_to file triggering errors in flake8, pydocstyle, and pylint

ns-rse:
Closes #482

* Replace single for double quotes in output so that linting with Black passes.
* Updates tests.

### DIFF
--- a/src/setuptools_scm/__init__.py
+++ b/src/setuptools_scm/__init__.py
@@ -4,7 +4,7 @@
 """
 from __future__ import annotations
 
-import os
+from pathlib import Path
 import re
 from typing import Any
 from typing import Pattern
@@ -46,8 +46,8 @@ def dump_version(
     template: str | None = None,
 ) -> None:
     assert isinstance(version, str)
-    target = os.path.normpath(os.path.join(root, write_to))
-    ext = os.path.splitext(target)[1]
+    target = Path(root) / Path(write_to)
+    ext = target.suffix
     template = template or TEMPLATES.get(ext)
     from ._trace import trace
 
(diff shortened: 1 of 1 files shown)


src/setuptools_scm/__init__.py
```diff
@@ -46,22 +46,27 @@ def dump_version(
...
+                target,
             )
         )
     version_tuple = _version_as_tuple(version)
 
-    with open(target, "w") as fp:
-        fp.write(template.format(version=version, version_tuple=version_tuple))
+    with target.open("w") as fp:
+        fp.write(
+            template.format(version=version, version_tuple=version_tuple).replace(
```
RonnyPfannschmidt, line 66: Can't accept that as is, as it may break templates that include certain escapes

RonnyPfannschmidt (changes requested):

PR #821, opened 2023-03-20
Makes .py output black compliant

### LINKED ISSUE #482: Auto-generated write_to file triggering errors in flake8, pydocstyle, and pylint

ns-rse:
Closes #482

* Replaces single with double quotes in output.
* Updates tests in light of this.

### DIFF
--- a/src/setuptools_scm/__init__.py
+++ b/src/setuptools_scm/__init__.py
@@ -62,10 +62,11 @@ def dump_version(
     version_tuple = _version_as_tuple(version)
 
     if ext == ".py":
-        template = template.replace("'", '"')
-
-    with target.open("w") as fp:
-        fp.write(template.format(version=version, version_tuple=version_tuple))
+        with target.open("w") as fp:
+            fp.write(template.format(version=version, version_tuple=version_tuple).replace("'", '"'))
+    else:
+        with target.open("w") as fp:
+            fp.write(template.format(version=version, version_tuple=version_tuple))
 
 
 def _do_parse(config: Configuration) -> _t.SCMVERSION | None:
@@ -78,9 +79,7 @@ def _do_parse(config: Configuration) -> _t.SCMVERSION | None:
     if config.parse:
         parse_result = config.parse(config.absolute_root, config=config)
         if isinstance(parse_result, str):
-            raise TypeError(
-                f"version parse result was {str!r}\nplease return a parsed version"
-            )
+            raise TypeError(f"version parse result was {str!r}\nplease return a parsed version")
 
         if parse_result:
             assert isinstance(parse_result, ScmVersion)
@@ -89,9 +88,7 @@ def _do_parse(config: Configuration) -> _t.SCMVERSION | None:
             parsed_version = _version_from_entrypoints(config, fallback=True)
     else:
         # include fallbacks after dropping them from the main entrypoint
-        parsed_version = _version_from_entrypoints(config) or _version_from_entrypoints(
-            config, fallback=True
-        )
+        parsed_version = _version_from_entrypoints(config) or _version_from_entrypoints(config, fallback=True)
 
     return parsed_version
 
--- a/src/setuptools_scm/__init__.py
+++ b/src/setuptools_scm/__init__.py
@@ -4,7 +4,7 @@
 """
 from __future__ import annotations
 
-import os
+from pathlib import Path
 import re
 from typing import Any
 from typing import Pattern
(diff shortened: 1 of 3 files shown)


src/setuptools_scm/__init__.py
```diff
@@ -46,22 +46,27 @@ def dump_version(
...
             )
         )
     version_tuple = _version_as_tuple(version)
 
-    with open(target, "w") as fp:
-        fp.write(template.format(version=version, version_tuple=version_tuple))
+    with target.open("w") as fp:
+        fp.write(
+            template.format(version=version, version_tuple=version_tuple).replace(
+                "'", '"'
```
RonnyPfannschmidt: This unfortunately is unacceptable as it may break custom templates

RonnyPfannschmidt (changes requested):

RonnyPfannschmidt:
Please avoid repeatedly creating new pull requests for the same set of changes

ns-rse:
> Please avoid repeatedly creating new pull requests for the same set of changes

My apologies, I pushed the change too early with an error then cocked up fixing it.

I've now restricted quote substitution to only files that are `.py` which are those that may be linted by Black. I was unsure whether to restrict further to only do so when the `TEMPLATES` derived `.py` are used and would happily take advice on this.

src/setuptools_scm/__init__.py
```diff
@@ -46,21 +46,25 @@ def dump_version(
...
-                os.path.splitext(target)[1], target
+                ext,
+                target,
             )
         )
     version_tuple = _version_as_tuple(version)
 
-    with open(target, "w") as fp:
+    if ext == ".py":
+        template = template.replace("'", '"')
```
RonnyPfannschmidt: this still breaks 3rd party templates while also not solving the repr issue

ns-rse:
Sorry, forgot to check the tests will resolve that now.

src/setuptools_scm/__init__.py
ns-rse: Yes I realised when the tests failed (putting child to bed distracted me). Just worked through and have corrected and checked, tests pass locally. :crossed_fingers:

src/setuptools_scm/__init__.py
```diff
@@ -46,22 +46,31 @@ def dump_version(
...
         )
     version_tuple = _version_as_tuple(version)
 
-    with open(target, "w") as fp:
-        fp.write(template.format(version=version, version_tuple=version_tuple))
+    if ext == ".py":
+        with target.open("w") as fp:
+            fp.write(
+                template.format(version=version, version_tuple=version_tuple).replace(
+                    "'", '"'
```
RonnyPfannschmidt, line 68: this pattern still can break examples

all it needs is a example with a required string thats getting broken by this

this replacement is only safe if it validates before and after generates the same ast

its far safer to have a  str wrapper that adds a format type thats returns a black safe string for tuples or strs

ns-rse: Thanks for the explanation and pointers. "_str wrapper_" is something I'm going to have to go and learn about, probably best closing this for now and I'll make another PR when I've understood the suggestion, got some working code and written tests if required.

PR #848, opened 2023-05-17
Normalize the dist-name, as per PEP503, when looking for overrides

### LINKED ISSUE #829: SETUPTOOLS_SCM_PRETEND_VERSION_FOR_DIST_NAME should escape hyphens
While it is great being able to override SETUPTOOLS_SCM_PRETEND_VERSION for a particular distribution (e.g. working on a CI that doesn't clone full `.git` history or remotely developing inside a docker container through ssh), it currently has an inconvenience: setting the environment variable will likely not be shell-compliant, as most python package names contain dashes/hyphens, and those are not allowed. For instance, if your package is called `hello-world`, the following would not work:
```sh
export SETUPTOOLS_SCM_PRETEND_VERSION_FOR_HELLO-WORLD=1.2.3
# bash: export: `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_HELLO-WORLD=1.2.3': not a valid identifier
```
whereas with
```sh
export SETUPTOOLS_SCM_PRETEND_VERSION_FOR_HELLO_WORLD=1.2.3
pip install .
```
the override gets ignored.

The only workaround I've found is using the following syntax:
```sh
env 'SETUPTOOLS_SCM_PRETEND_VERSION_FOR_HELLO-WORLD=1.2.3' pip install .
```
but it isn't a great user experience, especially since it can't easily be set externally.

Intuitively, it seems like the fix would be to simply replace `dist_name.upper()` in https://github.com/pypa/setuptools_scm/blob/main/src/setuptools_scm/_overrides.py#L19 with `dist_name.upper().replace("-", "_")`

pelson:
Closes #829.

It is worth noting that technically this is a breaking change. I didn't aim for backwards compat, as I figured the behaviour was pretty inconvenient/unfriendly, as per the original issue.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ breaking
   and where hiding dirty states that are now explicitly dirty
 * depend on later importlib for the full selectable api
 * move setuptools integration code to private sub-package
+* use normalized dist names for the SETUPTOOLS_SCM_PRETEND_VERSION_FOR_${DIST_NAME}
 
 features
 --------
--- a/README.rst
+++ b/README.rst
@@ -262,7 +262,7 @@ Note that running this Dockerfile requires docker with BuildKit enabled
 `[docs] <https://github.com/moby/buildkit/blob/v0.8.3/frontend/dockerfile/docs/syntax.md>`_.
 
 To avoid BuildKit and mounting of the .git folder altogether, one can also pass the desired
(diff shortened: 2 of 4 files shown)


RonnyPfannschmidt (changes requested):
Lovely, thanks! 

let's add a note on this breaking change to the changelog and land it

RonnyPfannschmidt (approved):

RonnyPfannschmidt:
thanks!

pelson:
> thanks!

Glad we could ship it!

Thanks for the quick response and positive experience! :+1:

PR #850, opened 2023-05-21
remove typing_extensions runtime dep on >= py3.8

gotmax23:
This makes it easier to bootstrap setuptools_scm and makes builds quicker, as pip doesn't need to download/install an extra package when preforming isolated builds.


RonnyPfannschmidt (approved):
Thanks for preparing this

Issue #561, opened 2021-04-24
opt out file finders


RonnyPfannschmidt:
followup to pypa/setuptools#2652
followup to #190

unfortunately people have to have setuptools_scm installed in their envs, and thus file finders are active even if they don't want them active

we should figure a local opt-out/in as workaround
a opt-in may be possible by adding well chosen global state

aka only working on folders we registered by  a integration before

jsallay:
I've read through the various issues on this subject and I just wanted to add a point that I haven't seen brought up.  I first noticed this issue when I saw that my CI files were being added to the sdist.  I also checked your releases and saw that the setuptools_scm .github folder is included in the sdist.

I worry about this from a security perspective because sometimes people hard-code secrets for CI.  By including it in the sdist, they are now releasing their secrets with the package (even when creating the package from a non-public repo).  I know that it would be better for people to NOT commit secrets to their repos.  But it happens quite often and it isn't obvious to a package creator that those files will be included.

allisonkarlitskaya:
To paraphrase a bit from #847: I really think this should be some kind of opt-in rather than an explicit opt-out.  Every user of setuptools should not be expected to opt-out of every single plugin that they know (or don't know) about just in case it might be installed in their environment.

Two ideas for a low-barrier opt in, which (together?) would probably cover 90%+ of existing users:
 - mention of `setuptools_scm` as a build dependency
 - `[tool.setuptools_scm]` section in `pyproject.toml`, as is done for the version behaviour

That could then be coupled with an explicit opt-out for people who want the version behavior without the file disting behaviour.

RonnyPfannschmidt:
Structurally this is a upstream issue with setuptools, a file finder is utterly blind and without context


Most recent setuptools versions are working on providing a new way to do this, but nobody so far tried to use that

I consider the secret leaking a strawman, any secret that's committed unencrypted to a git repo ough to be considered compromised

Im happy to accept contributions or to help anyone wanting to fix this in setuptools

I'm not going to try to fix this myself within the next 3-5 years

con-f-use:
Btw. with build isolation in wide use, e.g. by pip, why would people have setuptools_scm installed in their project's Python environment these days. Even when installing in editable mode, most tools would build an editable wheel in isolation and install that . As a build dependency in the isolated  build env, setuptools_scm gets automatically downloaded and when not building, I don't see why you'd need it. What am I missing?

[...]

PR #851, opened 2023-05-23
Allow opt-out of find_files behavior. (issue 561)

daveware-nv:
~Changes it so find_files is only enabled with config parameter `enable_find_files` is set to true, otherwise will always return empty list.~

Adds config argument `enable_find_files` that when set to `false` allows `find_files` behavior to be disabled.

Addresses https://github.com/pypa/setuptools_scm/issues/561

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,9 +7,10 @@ or in a SCM managed file.
 
 Additionally ``setuptools_scm`` provides setuptools with a list of
 files that are managed by the SCM (i.e. it automatically adds all of
-the SCM-managed files to the sdist). This is opt-in, by supplying
-config argument: ``enable_find_files``. Once opted in unwanted files
-must be excluded by discarding them via ``MANIFEST.in``.
+the SCM-managed files to the sdist). Unwanted files must be excluded
+by discarding them via ``MANIFEST.in``. It is also possible to
+opt-out of this behavior by setting the ``enable_find_files`` config
+argument to false.
 
 ``setuptools_scm`` supports the following scm out of the box:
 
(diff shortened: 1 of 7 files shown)


RonnyPfannschmidt (changes requested):
for legacy support this has to be opt-in

im researching ways to enable this better by adding a new sdist subcommand instead of obtaining a unrelated possibly miss-configured config object for just the file listing

daveware-nv:
OK, I've switched the default to true.

ljames8:
isn't this PR good to be merged now the requested changes have been made ?

RonnyPfannschmidt:
cant merge this yet as opting out of the file finders possibly opts into version finding + there is a bug

Lenbok:
@RonnyPfannschmidt Do you have more information, e.g. what is the bug that is preventing this being merged?

webknjaz:
@daveware-nv looks like the PR title needs to be fixed to reflect the post-review changes. Could you update it?

RonnyPfannschmidt:
currently there are many merge conflicts - im unable to merge as is

in the next major release i want to switch from file finders to custom setuptools subcommands as currently file finder opt out requires setuptools_scm "opt-in"

jaraco:
> cant merge this yet as opting out of the file finders possibly opts into version finding + there is a bug

In https://github.com/pypa/setuptools_scm/pull/364#discussion_r349931908, I advocated for there being a separate boolean to enable (or disable) version inference. Probably that needs to be implemented (first) so people can opt out of versions inference as well. Either that, or the `tool.setuptools_scm` needs to be deprecated and superseded by a new name that could have different defaults (and doesn't use the section header to imply enablement of a feature).

RonnyPfannschmidt:
closing due to lack of follow up

Issue #840, opened 2023-04-26
Dependency on hatchling is leading to a circular reference


gmc444-b:
`setuptools_scm` v 7.0.1 now uses `hatchling` as a build backend. I'm encountering a circular dependency, where `hatchling` 1.14.0 depends on `pluggy` 1.0.0, which depends on `setuptools_scm`.

Would it be possible to avoid using `hatchling`?

gmc444-b:
I need to correct what I reported above. The dependency comes up with the work going into the 8.0 release, not the 7.0.1 release. But you may still want to be aware of the circular dependency.

PR #852, opened 2023-05-24
fix typing-extensions constraint and don't duplicate runtime deps

gotmax23:


RonnyPfannschmidt (commented):
Lovely, i wasn't aware of that option

Unfortunately I'll have to go back to setuptools due to bootstrapping concerns for now

RonnyPfannschmidt:
#855 migrates back to setuptools to resolve #840 - i wonder if we can replicate this for setuptools, or if we need some extra code

RonnyPfannschmidt:
for now im closing this as the solution for setuptools seems non-obvious

this is still relevant if we ever get the chance for a more complete hatch port when the dependeny loop can be resolved

RonnyPfannschmidt:
thanks for providing this solution, its indeed awesome and its a bummer we have to stay on plain setuptools for a while longer

i'll definitively take it in once we get a chance

PR #854, opened 2023-05-25
pass `ScmVersion` to dump_version to provide scm_version as extra template variable

### LINKED ISSUE #577: QST: how to get the complete git commit hash
Is there a way to get the complete git commit hash using `setuptools_scm`?

### LINKED ISSUE #818: Expand write_to_template to permit commit hash and date
While shifting a project to setuptools_scm, I saw someone ask if they could generate a copyright field dynamically based on the current year. Something like:

```Python
from datetime import datetime
__copyright__ = f"Copyright {datetime.today():%Y}, The Project X developers"
```

Obviously this would be bad, as the copyright would apply to the most recent change to the source code, not time of import. This information is available to setuptools_scm, and it could be useful to permit something like the following:

```Python
from datetime import datetime
from ._version import commit_date

__copyright__ = f"Copyright {datetime.fromisoformat(commit_date):%Y}, The Project X developers"
```

Is there any appetite to permit additional SCM-derived data in the generated file? I would imagine anything available in the `ScmVersion` data class would be potentially useful. Beyond saving a chore, generating more detailed bug report templates are the main other use case I can think of.

Anyway, any appetite for this? It could be done in a separate build tool plugin, but it would be nice to consolidate interaction with the SCM to a single utility.

### LINKED ISSUE #853: parse `SCMVersion` to dump_version
Currently, only `version`(str) and `version_tuple` are exposed in `dump_version`, therefore it is not possible to write additional informations such as `node`, `time`, etc. 

We propose to parse `version` to `dump_version` and rename the `version` to `version_str` to give more flexibility in `dump_version`. 

See also #577

ZhiyuanChen:
fixes #853

related to  #818 
related to #577


src/setuptools_scm/__init__.py
```diff
@@ -155,20 +163,21 @@ def _get_version(config: Configuration) -> str | None:
...
+    version_str = _format_version(
         parsed_version,
         version_scheme=config.version_scheme,
         local_scheme=config.local_scheme,
     )
     if config.write_to is not None:
         dump_version(
             root=config.root,
-            version=version_string,
+            version=parsed_version,
```
RonnyPfannschmidt: i recommend using the name scm_version or parsed_version as input names so it can be distinguished

src/setuptools_scm/__init__.py
```diff
@@ -43,27 +43,35 @@
 
 def dump_version(
     root: _t.PathT,
-    version: str,
+    version: Version,
```
RonnyPfannschmidt: the argument rename is a breaking change change would silently break downstream users
as such

RonnyPfannschmidt (changes requested):
thanks for preparing this 

unfortunately the argument rename is a breaking change that will get downstreams by surprise

as such the old name needs to stay and the new object needs to go in by a different name

there may be need for extra considerations wrt what attributes to put on ScmVersion to help formatting

RonnyPfannschmidt (commented):
this looks much better now, thanks!

for completeness i would recommend to add a unit-test that uses a custom template to dump the node in addition to the normal version and checks for it

with the addition of a changelog note on the feature addition this would be ready to merge then

i presume the pdb calls will go away then

:+1:

testing/test_basic_api.py
```diff
@@ -155,24 +160,34 @@ def test_root_relative_to(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> No
...
+    scm_version = meta(version, config=c)
+    dump_version(tmp_path, version, scm_version, "first.py")
     lines = read("first.py").splitlines()
     assert "__version__ = version = '1.0.dev42'" in lines
     assert "__version_tuple__ = version_tuple = (1, 0, 'dev42')" in lines
 
-    dump_version(tmp_path, "1.0.1+g4ac9d2c", "second.py")
+    version = "1.0.1+g4ac9d2c"
+    scm_version = meta("1.0.1", node="g4ac9d2c", config=c)
+    dump_version(tmp_path, version, scm_version, "second.py")
```
RonnyPfannschmidt: Unfortunately this doesn't Test The new feature

RonnyPfannschmidt (commented):
I took note that adding the parameter as is, does a avoidable breaking change

I recommend changing it to a optional keyword only parameter

As for testing,there is need for a own test that uses a custom template that's actually using data from the SCM version

RonnyPfannschmidt (commented):
Excellent work, thanks

Please confirm if i should squash or merge

I'll give it a last look over once I return to the computer

RonnyPfannschmidt (approved):

ZhiyuanChen:
> Excellent work, thanks
> 
> Please confirm if i should squash or merge
> 
> I'll give it a last look over once I return to the computer

Thanks~ I believe it's better to squash the changes, as most commits are redundant"

RonnyPfannschmidt:
Thanks for enabling this

PR #855, opened 2023-05-26
fix #840 - return to setuptools

### LINKED ISSUE #840: Dependency on hatchling is leading to a circular reference

RonnyPfannschmidt:


nanonyme:
Was it really necessary to mark rich as mandatory dependency instead of extra while doing this? It makes the dependency tree of even more complex. Seems the original issue was about simplifying the dependency tree.

PR #858, opened 2023-05-29
Emphasize permitted version format in packages for publication

### LINKED ISSUE #322: README should mention that PyPI doesn't accept local version identifiers
And maybe suggest some workarounds for that.

Currently I came up with:
1) Setting env var with no local identifier
    ```console
    python -c 'import setuptools_scm; print(setuptools_scm.get_version(local_scheme=lambda v: ""))' | read SETUPTOOLS_SCM_PRETEND_VERSION
    ```
    which is scary because of #321
2) substituting `setuptools_scm.local_scheme` entrypoint with `lambda v: ""`

Which should probably be done only when uploading the dist because in all other cases it's still nice to have local identifier.

Ref: https://github.com/pypa/twine/issues/430
Ref: https://github.com/pypa/python-packaging-user-guide/issues/578

OlenaYefymenko:
The patch adds mention in README that local version identifiers in packages don't allow publishing to PyPI/TestPyPI. Also, it suggests alternative local servers for using local version identifiers.

Resolves #322

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -292,6 +292,15 @@ the tag.
 For Git projects, the version relies on `git describe <https://git-scm.com/docs/git-describe>`_,
 so you will see an additional ``g`` prepended to the ``{revision hash}``.
 
+.. note::
+
+    According to `PEP 440 <https://peps.python.org/pep-0440/#local-version-identifiers>`_,
+    if a version includes a local component, the package cannot be published to public
+    package indexes like PyPI or TestPyPI. The disallowed version segments may
+    be seen in auto-publishing workflows or when a configuration mistake is made.
+
+    However, some package indexes such as devpi or other alternatives allow local
+    versions. Local version identifiers must comply with PEP `PEP 440`_.
 
 Semantic Versioning (SemVer)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


webknjaz:
@RonnyPfannschmidt looks like a twine bug or is the PyPI renderer that limited?

RonnyPfannschmidt:
@webknjaz this is not a bug, it's a intended and well chooosen limit

README.rst
```diff
@@ -292,6 +292,15 @@ the tag.
...
 
+.. note::
+
+    According to `PEP 440 <https://peps.python.org/pep-0440/#local-version-identifiers>`_,
+    if a version includes a local component, the package cannot be published to public
+    package indexes like PyPI or TestPyPI. The disallowed version segments may
+    be seen in auto-publishing workflows or when a configuration mistake is made.
+
+    However, some package indexes such as devpi or other alternatives allow local
+    versions. Local version identifiers must comply with PEP `PEP 440`_.
```
webknjaz:
```suggestion
    versions. Local version identifiers must comply with `PEP 440`_.
```

webknjaz:
> it's a intended and well chooosen limit

It turned out to be a typo

RonnyPfannschmidt (approved):
Thank you for adding those details

webknjaz:
@RonnyPfannschmidt easy merge or did you want to rephrase something?

RonnyPfannschmidt:
I'm traveling and off the computer

RonnyPfannschmidt:
Thanks!

PR #859, opened 2023-06-01
add command support for the libgit2 cli

xloem:
This pull requests makes small changes so the codebase has enough functionality with libgit2 to install itself.

The libgit2 example cli is at https://github.com/libgit2/libgit2/blob/main/examples/lg2.c . This git replacement is used by a-shell at https://github.com/holzschu/a-shell .

- specified the dirty text in describe, lg2 requires
- the default lg2 code does not implement —abbrev-ref, so a whole ref is used if that fails
- most log flags are missing. -n is present as —max-count, which i changed, but it falls back to retrieving unformatted output and parsing it, if dateutil is present to do so. the code calling the function defaults to the system time if None is returned.
- GitWorkdir.from_potential_worktree calls an unsupported command line and i didn’t discern the intended logic of the function to patch it, it looks like it ends up returning data that was passed to it (the root dir parenting the git dir), so i made an alternate condition that defaults to this. is it to handle git dirs that are outside work trees?
- when detecting a command, lg2 returns a failure code for help, but both lg2 and git return a success code for version, so i changed the probe to check version

### DIFF
--- a/src/setuptools_scm/git.py
+++ b/src/setuptools_scm/git.py
@@ -36,7 +36,7 @@ DESCRIBE_UNSUPPORTED = "%(describe"
 DEFAULT_DESCRIBE = [
     "git",
     "describe",
-    "--dirty",
+    "--dirty=-dirty",
     "--tags",
     "--long",
     "--match",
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (commented):
I'm unfamiliar with the details here

If this is about a shell that intentionally replaces the git command with something incompatible,then I have to reject this unless a sensible automated testing approach exists

xloem:
I was interested in doing a little work to support this environment, so I’m interested in what solution would be most welcomed if I work more.

A-shell likely uses lg2 because it is easier to build it in webassembly, although I don’t know for sure.

lg2 implements a subset of the git commands and flags and appears quite compatible but only within that implemented subset.

With the exception of my choice to add optional parsing of the raw logs, it seemed to me like the compatibility changes are quite minor and meshed right in; however, I could instead approach A-shell or lg2 to make changes for support here.

Automated testing for this would involve building the libgit2 tree and running the example binary as git. Maybe I’ll learn your testing environment a little, or maybe this contribution would be better at a later time.

RonnyPfannschmidt:
Basically unless we can have automated tests for this we can't merge in good conscience as we wouldn't have any idea if a new feature or change breaks it

RonnyPfannschmidt:
If there's a easy way to make a docker Image with the right env to run the tests it can be added to the ci

xloem:
I added tests at https://github.com/xloem/setuptools_scm/tree/tests but it looks like there’s more to do to get them all passing than I am up for right now, so closing this for now.

RonnyPfannschmidt:
Thanks for starting this experiment, i hope we can reduce the friction in future

PR #860, opened 2023-06-06
Add note for conda users

d-chambers:
By default, when setuptools-scm projects are built with conda the version returned by `importlib.metadata.version` is always  0.0.0. It took me a few hours to track down why and find a solution, so I have included a small note in the readme and a link to an example on how to fix this to save someone else time in the future.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -677,7 +677,7 @@ Note for Conda Users
 Setuptools-scm will not set the correct version by default when built with conda. In order
 for it to work, the conda recipe needs to be modified to use the
 SETUPTOOLS_SCM_PRETEND_VERSION env variable.
-[Here](https://github.com/conda-forge/pint-feedstock/commit/b107bcafd4a8571b416bd06b7c292eccc62ef3fc)
+`Here <lhttps://github.com/conda-forge/pint-feedstock/commit/b107bcafd4a8571b416bd06b7c292eccc62ef3fc>`_
 is an example of how to do this.
 
 
--- a/README.rst
+++ b/README.rst
@@ -675,8 +675,8 @@ Note for Conda Users
 ~~~~~~~~~~~~~~~~~~~~
 
 Setuptools-scm will not set the correct version by default when built with conda. In order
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
i was not aware that conda anhillates metadata access, this note is incorrect and points to the wrong issue

d-chambers:
> this note is incorrect and points to the wrong issue

Feel free to reword, but it is indeed a real issue and the modification to the conda recipe does fix the problem. We could put this information somewhere else if this isn't the right place, but it would have saved me a few hours if this was mentioned somewhere in the docs 🤷🏻

RonnyPfannschmidt:
Can you link the recipe in question?

d-chambers:
From [Pint](https://github.com/hgrecco/pint):
- [The issue](https://github.com/conda-forge/pint-feedstock/issues/32)
- [The Fix](https://github.com/conda-forge/pint-feedstock/commit/b107bcafd4a8571b416bd06b7c292eccc62ef3fc)

From [DASCore](https://github.com/dasdae/dascore) (I have verified both the bug and fix in this project)
- [The issue](https://github.com/DASDAE/dascore/issues/128)
- [The fix](https://github.com/conda-forge/dascore-feedstock/commit/1b1f17f344d9384a317feafe984e916f22232274)

RonnyPfannschmidt:
At first glance it seems like you have a excessive non standard setuptools manifest grafting setup

Does it even worth for the sdist with normal python

When I get back to the computer I'll check the tarball

RonnyPfannschmidt:
the tarball looks correct, the setuptools and setuptools_scm versions for both seem to be pretty low, its possible that its a setuptools version without the metadata hook and toml support

RonnyPfannschmidt:
closing as incorrect conda setups

backmari:
This issue is mentioned in these docs for conda package maintainers: https://conda-forge.org/docs/maintainer/knowledge_base/#using-setuptools_scm

PR #870, opened 2023-06-25
Remove implementation code from main import + replace write_to with version_file

### LINKED ISSUE #188: write_to should also obey relative_to
Say you have e.g.
```
repo/.git
repo/java/whatever
repo/python
repo/python/setup.py
repo/python/src/yourmagicpackage
```
ie. a typical case where you do need to specify root and relative_to.
In this case it seems preferable for write_to to also obey relative_to, i.e. you (I)'d like to write `relative_to=__file__, root="..", write_to="src/yourmagicpackage/_version.py"` rather than `..., write_to="python/src/yourmagicpackage/_version.py"`.

Obviously this would be a backcompat break, but I think the proposed behavior is more sensible?

### LINKED ISSUE #875: path to write_to relative to what?
I have a "non standard" setup (long legacy) where my package is not at the root of the git repo.

So I have to do:

```
[tool.setuptools_scm]
root = ".."
```
which works for getting the version, and for the file-finding, which is great.

However, when I try to set it up to write the version to a file, I need to specify the path relative to the root, rather than relative to pyproject.toml.

```
[tool.setuptools_scm]
root = ".."
write_to = "py_gnome/gnome/_version.py"
```
(gnome is the package dir)

However, this then finds and writes the file properly when building the sdist. but when build tries to build the sdit, it is again looking for `../py_gnome/gnome/_version.py`

```
  File "/Users/chris.barker/miniconda3/envs/gnome_build/lib/python3.10/site-packages/setuptools_scm/__init__.py", line 77, in dump_version
    with open(target, "w") as fp:
FileNotFoundError: [Errno 2] No such file or directory: '../py_gnome/gnome/_version.py'
```

Which does not exist in the sdist, of course. :-(

Possible Solutions:

1) The write_to path should be relative to where the pyproject.toml file is, rather than the SCM root.

2) When building an sdist (outside of SCM), maybe this code should not be running at all? I *thought* the idea was that the _version.py file would get added to the sdist, to setuptools_scm wouldn't need to try to figure it out again. What does it do? it can't use the SCM from an sdist, so ???

RonnyPfannschmidt:
- [x] finish renaming the internal implementation functions
- [x] add doc-strings as needed

closes #188
closes #875


README.rst
```diff
@@ -488,28 +410,20 @@ function:
...
 
 :SETUPTOOLS_SCM_PRETEND_VERSION:
     when defined and not empty,
     its used as the primary source for the version number
-    in which case it will be an unparsed string
+    in which case it will be an string
+
+    .. warning::
+
+      its strongly  recommended to use use distribution name specific pretend versions
```
antazoey, line 424: There is a random extra space here.

antazoey, line 424: It should be either `it's` or `it is`. `its` means something else.

src/setuptools_scm/__init__.py
```diff
@@ -138,62 +52,23 @@ def get_version(
...
-    parsed_version = _do_parse(config)
-    if parsed_version is None:
-        return None
-    version_string = _format_version(
-        parsed_version,
-        version_scheme=config.version_scheme,
-        local_scheme=config.local_scheme,
-    )
-    if config.write_to is not None:
-        dump_version(
```
antazoey, line 176: my version file is no longer being dumped anywhere and I am trying to figure out why.

PR #886, opened 2023-08-21
ignore jupyter notebooks by default

morotti:
hello,

could you consider excluding jupyter notebook files by default?

notebooks are stored in source control and they end up in the python package when using setuptools-scm.

notebooks are quite problematic for both the git repo and the python package.
they cache the content of each jupyter cell on save. it's very easy for users to accidentally make notebooks that are many many MB, just for loading a bit of data to play with. 
I've seen quite a few accidents where (very) large jupyter notebooks got included in a wheel, making the wheel much larger than necessary.

it would be nice to ignore them by default.
I don't think there is any use case where jupyter notebooks are expected to be included in source dist or wheels, but I could be wrong about that. 

thank you

### DIFF
--- a/src/setuptools_scm/_file_finders/__init__.py
+++ b/src/setuptools_scm/_file_finders/__init__.py
@@ -70,6 +70,16 @@ def scm_find_files(
         for filename in filenames:
             if not force_all_files and _link_not_in_scm(filename):
                 continue
+            if not force_all_files and filename.lower().endswith(".ipynb"):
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt (changes requested):
We go through the hooks of running export for git at least, unless there is a compelling reason not to use Export filters id strongly like to avoid this

RonnyPfannschmidt:
closing due to lack of feedback for the change request, happy to reopen with compelling feedback

PR #900, opened 2023-09-13
Use an UTF8 enabled locale for the run() environment.

arvidma:
I have not managed to make a unit test that recreates the problem. While failing with that, I think I have narrowed down the true origin of the problem: 

It is not mercurial having trouble with LC_ALL=C. It is Python itself that fails during start-up, when it encounters a pth-file in site-packages, that has a non-ascii character inside of it. This in turn only happens when I initiate the `pip wheel` command while my virtualenv (with an editable install of the project) is active.

In order to make a test for this, one would need to prep the test by installing an (in)appropriate package in the test environment. I do not want to take on that type of effort, so I'll just leave this one-liner here for you to accept or reject!

With my findning on the core of the problem, I have a solid work-around available (just don't run pip wheel with an active virtualenv), so I don't feel very strongly about the issue any more.

Kind regards,
Arvid

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -44,6 +44,7 @@ bugfixes
 
 * fix #883: use HeadersParser to ensure only mime metadata in headers is used
 * fix #884: parse calver dates from versions with the v prefix
+* don't use a C locale without UTF-8 support, when running commands.
 
 v7.1.0
 ======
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (approved):
thanks for adding this,
lets add a comment to mention python using ascii without the utf-8 + add a changelog entry on the bugfix to complete it

test in this case is nice to have but not critical

RonnyPfannschmidt:
thanks :love_you_gesture:

PR #911, opened 2023-09-20
Add timeout to tests

jaap3:
Just noticed that the current release workflow is hanging on the pytest run on Windows (msys2).

Looking at previous runs the tests usually don't take much longer than 5/6 minutes (10 minutes tops). Imposing a 15 minute timeout seems reasonable.


RonnyPfannschmidt (approved):

RonnyPfannschmidt:
thanks this is a good add :)

PR #917, opened 2023-09-20
docs: correct word and space in `SETUPTOOLS_SCM_PRETEND_VERSION` warning

antazoey:
* Replaces `its` with `it is`. The word `its` means something else; `it's` was definitely intended.
* Removes an extra space.

NOTE: I am debugging my project's version files are no longer being generated, both locally and in CI but I came across this doc issue so I thought I'd open the PR... But really, my priority here is to figure out why my version file doesn't get created anymore. I tried switching `write_to` to `version_file` but that did not seem to work.

I have been looking at https://github.com/pypa/setuptools_scm/pull/870 to try and find answers.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -108,7 +108,7 @@ Callables or other Python objects have to be passed in `setup.py` (via the `use_
 :   used as the primary source for the version number
     in which case it will be an unparsed string
 
-    !!! warning "its strongly  recommended to use use distribution name specific pretend versions"
+    !!! warning "it is strongly recommended to use use distribution name specific pretend versions"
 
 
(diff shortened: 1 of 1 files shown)


antazoey:
Perhaps I should open a separate issue, but I am having to pin <8 on my project because the version file refuses to generate. I am guessing I am doing something wrong but I can't figure out what. I am going to keep debugging.

RonnyPfannschmidt:
pleae create a own issue and addthe output when using SETUPTOOLS_SCM_DEBUG while building

antazoey:
It sounds like my issue may be this one: https://github.com/pypa/setuptools_scm/issues/919

RonnyPfannschmidt:
thanks for those fixes :+1:

PR #927, opened 2023-09-21
Share the stderr with the user, when non-zero exit code

arvidma:
When a has_command() returns False, we eat all the output and just say that it was not found. In case the False return is due to a non-zero exit status from a command we did actually find, this makes it quite messy to figure out what went wrong.

This is an attempt to let output from the failing command bubble up the stack, with a minimal amount of disruption to surrounding code.

### DIFF
--- a/src/setuptools_scm/_run_cmd.py
+++ b/src/setuptools_scm/_run_cmd.py
@@ -14,7 +14,6 @@ from typing import TypeVar
 
 from . import _log
 from . import _types as _t
-from ._types import Result
 
 if TYPE_CHECKING:
     BaseCompletedProcess = subprocess.CompletedProcess[str]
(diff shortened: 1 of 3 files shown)


src/setuptools_scm/_run_cmd.py
```diff
@@ -172,27 +173,38 @@ def _unsafe_quote_for_display(item: _t.PathT) -> str:
...
     else:
         res = not p.returncode
     if not res and warn:
         warnings.warn("%r was not found" % name, category=RuntimeWarning)
-    return res
+
+    if not res and not message:
+        message = p.stderr
+
+    return Result(status=res, message=message)
```
RonnyPfannschmidt: lets define the Result class in the same module as its only used for this particular function

it mgiht make sense to use the logging based output in any case (when has command initially was created, warnings was used as trace was only opt in by env var

now with the logging setup that chooses level, we can just log a error directly

arvidma: You are very fast to reply! Thank you. :smile: 

Just logging directly would indeed more straight forward. That will remove the need to return a complex type from has_command, so I can just skip the result wrapper entirely. I'll rework the PR.

arvidma:
The diff is a bit smaller now, and there is a test. :smile:

RonnyPfannschmidt (approved):
Lovely

RonnyPfannschmidt:
Want to squash?

arvidma:
> Want to squash?

Yes please!

PR #936, opened 2023-09-22
introduce ruff as a linter

RonnyPfannschmidt:


pyproject.toml
```diff
@@ -43,7 +43,7 @@ dependencies = [
   "packaging>=20",
   "setuptools",
   'tomli>=1; python_version < "3.11"',
-  'typing-extensions; python_version < "3.11"',
+  "typing-extensions",
```
eli-schwartz, line 45: "typing extensions always needed"

Can you clarify what this means? I still do not see anywhere that it is needed at runtime, and you appear to have reverted the work from #850 as this is now once again needed at runtime

eli-schwartz: I do agree, by the way, that it's not clear the dependencies were correct -- it seems to be unconditionally used, but only at mypy/lint time rather than at runtime. So instead of a versioned dependency, it should have been a test extra.

RonnyPfannschmidt: im happy to make it optional again once i have the time to make sure it works as intended

PR #940, opened 2023-09-26
Drop a redundant virtualenv test dependency

hroncok:
It was needed for test_distlib_setuptools_works only, which was removed in d78d9ccf062ad114e727718edd7a0d386bed3f85.


RonnyPfannschmidt (approved):
Thanks

hroncok:
Actually, this is also in pyproject.toml.

PR #942, opened 2023-09-26
Avoid rich as a build backend dependency

### LINKED ISSUE #941: Is it neccessary to pull in rich as build-backend depndency?
Hello,
I've noticed version 8 now declares a build-backend dependency on rich:

https://github.com/pypa/setuptools_scm/blob/6ab652e2beef00947d9a14cd2834a037ed4795db/pyproject.toml#L3-L6

However, the build works without rich. I assume that's because the import is try-expected:

https://github.com/pypa/setuptools_scm/blob/6ab652e2beef00947d9a14cd2834a037ed4795db/src/setuptools_scm/_log.py#L32-L42

In Fedora Linux, I plan to patch this out to simplify the bootstrap of our Python packages for the next Python version. However, I was wondering if this should actually be a build dependency. Why is it?

hroncok:
It was only added for debugging and it is not actually needed.

Fixes https://github.com/pypa/setuptools_scm/issues/941

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,6 @@
 [build-system]
 build-backend = "_own_version_helper:build_meta"
 requires = [
-  "rich",
   "setuptools>=61",
   'tomli; python_version < "3.11"',
 ]


RonnyPfannschmidt (approved):

RonnyPfannschmidt:
Thanks

PR #954, opened 2023-10-09
Reword docker usage hint

WhyNotHugo:
By default docker copies `.git`, so reworded that a bit. Note that, it **is** good practice to avoid copying `.git` into container images.

The provided snippet doesn't infer the version in the host; it merely mounts the git directory so that the version is inferred inside the container build environment. Corrected this clarification.


RonnyPfannschmidt:
Thanks

PR #958, opened 2023-10-13
Fix rst remnants

valberg:
I was reading the usage documentation and came across these small rst remnants which are not rendered correctly.


RonnyPfannschmidt (approved):
Good catch, thanks

PR #959, opened 2023-10-14
add 'info' command to cli

con-f-use:
Adds an `info` command to the `setuptools_scm` command line. This info command can be used to print settings, like so:

```shell
$ python -m setuptools_scm info dist_name parent root relative_to version_scheme local_scheme tag_regex
dist_name = setuptools-scm
parent = /home/jan/devel/python/setuptools_scm
root = .
relative_to = /home/jan/devel/python/setuptools_scm/pyproject.toml
version_scheme = guess-next-dev
local_scheme = node-and-date
tag_regex = re.compile('^(?:[\\w-]+-)?(?P<version>[vV]?\\d+(?:\\.\\d+){0,2}[^\\+]*)(?:\\+.*)?$')
```

This is only a proof of concept implementation, I could work on tests, stabilization and a proper implementation if that was desired.

### More details and examples:

```shell
$ python -m setuptools_scm --help
usage: python -m setuptools_scm [-h] [-r ROOT] [-c PATH] [--strip-dev]  ...

Print project version according to SCM metadata

options:
  -h, --help            show this help message and exit
  -r ROOT, --root ROOT  directory managed by the SCM, default: inferred from config file, or "."
  -c PATH, --config PATH
                        path to 'pyproject.toml' with setuptools_scm config, default: looked up in the current or parent directories
  --strip-dev           remove the dev/local parts of the version before printing the version

extra commands:
  
    ls                  list files managed by the SCM
    info                display setuptools_scm settings according to query e.g. dist_name

$ python -m setuptools_scm info --help
usage: python -m setuptools_scm info [-h] [query ...]

Display setuptools_scm settings according to query e.g. dist_name

positional arguments:
  query       query string(s) for the desired information, omit for list of valid query strings

options:
  -h, --help  show this help message and exit

$ python -m setuptools_scm info
relative_to
root
version_scheme
local_scheme
tag_regex
parentdir_prefix_version
fallback_version
fallback_root
write_to
write_to_template
version_file
version_file_template
parse
git_describe_command
dist_name
version_cls
search_parent_directories
parent
```

### DIFF
--- a/src/setuptools_scm/_cli.py
+++ b/src/setuptools_scm/_cli.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
 import argparse
+import json
 import os
 import sys
+from typing import Any
 
 from setuptools_scm import Configuration
 from setuptools_scm._file_finders import find_files
@@ -10,7 +12,7 @@ from setuptools_scm._get_version_impl import _get_version
 from setuptools_scm.discover import walk_potential_roots
 
 
-def main(args: list[str] | None = None) -> None:
+def main(args: list[str] | None = None) -> int:
     opts = _get_cli_opts(args)
     inferred_root: str = opts.root or "."
 
@@ -36,11 +38,8 @@ def main(args: list[str] | None = None) -> None:
         raise SystemExit("ERROR: no version found for", opts)
     if opts.strip_dev:
         version = version.partition(".dev")[0]
-    print(version)
 
-    if opts.command == "ls":
-        for fname in find_files(config.root):
-            print(fname)
+    return command(opts, version, config)
 
 
 def _get_cli_opts(args: list[str] | None) -> argparse.Namespace:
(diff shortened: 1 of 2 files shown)


con-f-use:
Note: The current implementation does not print the version anymore, when the `ls` command is used. I'm not sure if the old behavior was intentional or a bug.

I also noticed, that invoking `python -m setuptools_scm ...` does not honor the projects build-system requirements (how would it?). I.e. if a project uses a plugin that provides, e.g., a custom version scheme, `python -m setuptools_scm` will error because it is not aware of that scheme until that plugin is installed in the environment the invocation takes place in.

One might also want to restrict the valid queries, filtering out the rather unuseful keywords.

RonnyPfannschmidt (commented):
The move up and early return skip's the version printing

I recommend reordering the logic


I'm currently not sure on the details and won't be able to try it's on a computer soon due to traveling

con-f-use:
No worries, take your time. I'm glad you even look at it at all.

The move is intentional, because I thought that behavior for `ls` might be a bug. If one were to pipe or parse the output of `ls`, one would have to awkwardly filter out the first line. I would expect `ls` to only print files.

In short, the code in this PR has the following intended behavior:

- `python -m setuptools_scm` --> prints the version
- `python -m setuptools_scm ls` --> does NOT print the version, only the files found
- `python -m setuptools_scm info ...` --> does NOT print the version, only the desired settings fields

But if you think it would break that much legacy, I'm not married to that.

RonnyPfannschmidt:
its not that this breaks legacy, its just that the change would require a major release

so i'm under the impression a little extra complexity to have a ls-only would help that

con-f-use:
Sorry, not sure what you mean. Do you want me to change `ls` so it still prints the version and then have something we could call e.g. `list` or `files` that does the same without version printing first?

RonnyPfannschmidt:
correct

i'm also wondering about whether listing files, versions and info should be flags instead of commands

(i'm imagining something like `python -m setuptools_scm --list-files --show-version --show-info  -o json`)

con-f-use:
I'm not averse to doing that. One could also think of re-writing the cli in something like click. Would you be averse to that and the extra dependency?

RonnyPfannschmidt:
unfortunately not - it can only be a extra dependency , never a mandatory one

the argparse setup should get simpler if the new options are enable/disable flags
i propose leaving the "ls" command as is, and adding the opt in/out for the info/version as a starting point thats quicky merged

then in a followup a json output style where each parameter can be asked for can be added

src/setuptools_scm/_cli.py
```diff
@@ -67,13 +66,106 @@ def _get_cli_opts(args: list[str] | None) -> argparse.Namespace:
...
+    if opts.format == "plain":
+        _print_plain(data)
+
+    if opts.format == "key-value":
+        _print_key_value(data)
+
+    return 0
+
+
+def _print_plain(data: dict[Any, Any]) -> None:
```
RonnyPfannschmidt:
```suggestion
def _print_plain(data: dict[str, Any]) -> None:
```

src/setuptools_scm/_cli.py
```diff
@@ -67,13 +66,106 @@ def _get_cli_opts(args: list[str] | None) -> argparse.Namespace:
...
+    if opts.query == []:
+        opts.no_version = True
+        sys.stderr.write("Available queries:\n\n")
+        opts.query = ["queries"]
+        data["queries"] = ["files"] + list(config.__dataclass_fields__.keys())
+
+    if opts.query is None:
+        opts.query = []
+
+    if opts.no_version is False:
```
RonnyPfannschmidt:
```suggestion
    if not opts.no_version:
```

src/setuptools_scm/_cli.py
```diff
@@ -67,13 +66,106 @@ def _get_cli_opts(args: list[str] | None) -> argparse.Namespace:
...
+    for file_ in files:
+        print(file_)
+    queries = data.pop("queries", [])
+    for query in queries:
+        print(query)
+    if data:
+        print("\n".join(data.values()))
+
+
+def _print_key_value(data: dict[Any, Any]) -> None:
```
RonnyPfannschmidt: dito

RonnyPfannschmidt (approved):
sorry this was hanging for a while

looks good :+1: 

please take a look at the type nitpicks and add a scriv fragment to the changelog

src/setuptools_scm/_cli.py
RonnyPfannschmidt:
```suggestion
def _print_key_value(data: dict[str, Any]) -> None:
```

src/setuptools_scm/_cli.py
```diff
@@ -77,11 +101,80 @@ def _get_cli_opts(args: list[str] | None) -> argparse.Namespace:
...
+    data: dict[str, Any] = {}
+
+    if opts.command == "ls":
+        opts.query = ["files"]
+
+    if opts.query == []:
+        opts.no_version = True
+        sys.stderr.write("Available queries:\n\n")
+        opts.query = ["queries"]
+        data["queries"] = ["files"] + list(config.__dataclass_fields__.keys())
```
RonnyPfannschmidt:
```suggestion
        data["queries"] = ["files", *config.__dataclass_fields__]
```

PR #973, opened 2023-11-07
Make Sphinx usage example clearer

### LINKED ISSUE #972: Documentation for Sphinx usage a bit unclear
The docs suggest including the following snippet in a Sphinx `conf.py` file

https://github.com/pypa/setuptools_scm/blob/c39489f9cfced1b584b054564a6579c821db494f/docs/usage.md?plain=1#L123-L128

to set the Sphinx documentation version from the installed package metadata. While in retrospect this probably should have been obvious, I originally copied this snippet verbatim without changing the argument of `get_version` to the relevant package name instead of `'setuptools_scm'`, causing the built documentation to instead display the installed version of `setuptools_scm`. Perhaps to make it more obvious the argument needs to be adjusted to the relevant package name this should match the previous example

https://github.com/pypa/setuptools_scm/blob/c39489f9cfced1b584b054564a6579c821db494f/docs/usage.md?plain=1#L113

and use `"package-name"` as the argument to `get_version`?

matt-graham:
Fixes #972 

Follows previous _version at runtime_ usage example in using `"package-name"` as placeholder argument to `importlib.metadata.version` rather than `'setuptools_scm'` to make it clearer that this needs to be changed to relevant package name.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -122,7 +122,7 @@ except PackageNotFoundError:
 
 ``` {.python file=docs/.entangled/sphinx_conf.py}
 from importlib.metadata import version as get_version
-release: str = get_version('setuptools_scm')
+release: str = get_version("package-name")
 # for example take major/minor
 version: str = ".".join(release.split('.')[:2])
 ```


RonnyPfannschmidt (approved):
Thanks 👍

PR #981, opened 2023-11-20
Documenting `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_${NORMALIZED_DIST_NAME}`

### LINKED ISSUE #978: Docs request: `SETUPTOOLS_SCM_PRETEND_VERSION` recommendation
From https://setuptools-scm.readthedocs.io/en/latest/config/#environment-variables as of `setuptools_scm==8.0.4`:

> `SETUPTOOLS_SCM_PRETEND_VERSION`
> 
> used as the primary source for the version number in which case it will be an unparsed string
> 
> it is strongly recommended to use use distribution name specific pretend versions

Can we add more details to this recommendation?  Why is it strong recommended to use distribution-specific pretend versions?

To me, adding distribution-specifics adds both more work and complexity, so it seems like a downside

jamesbraza:
Resolves https://github.com/pypa/setuptools_scm/issues/978

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -108,17 +108,22 @@ Callables or other Python objects have to be passed in `setup.py` (via the `use_
 :   used as the primary source for the version number
     in which case it will be an unparsed string
 
-    !!! warning "it is strongly recommended to use use distribution name specific pretend versions"
+    !!! warning ""
 
+        it is strongly recommended to use distribution-specific pretend versions
+        (see below).
 
 `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_${NORMALIZED_DIST_NAME}`
-:   used as the primary source for the version number
-    in which case it will be an unparsed string
+:   used as the primary source for the version number,
+    in which case it will be an unparsed string.
+    Specifying distribution-specific pretend versions will
+    avoid possible collisions with third party distributions
+    also using ``setuptools_scm``
 
     the dist name normalization follows adapted PEP 503 semantics, with one or
     more of ".-_" being replaced by a single "_", and the name being upper-cased
 
-    it takes precedence over ``SETUPTOOLS_SCM_PRETEND_VERSION``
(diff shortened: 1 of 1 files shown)


docs/config.md
```diff
@@ -108,17 +108,22 @@ Callables or other Python objects have to be passed in `setup.py` (via the `use_
...
 
+        it is strongly recommended to use distribution-specific pretend versions
+        (see below).
 
 `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_${NORMALIZED_DIST_NAME}`
-:   used as the primary source for the version number
-    in which case it will be an unparsed string
+:   used as the primary source for the version number,
+    in which case it will be an unparsed string.
+    Specifying distribution-specific pretend versions will
```
jamesbraza, line 119: Can I say something like "distribution-specific (aka package-specific)"

RonnyPfannschmidt: I'm wondering if it perhaps should change to python package name to disambiguate

jamesbraza: Do you mean in the docs, or in the docs and source code?

For convenience, https://github.com/pypa/setuptools_scm/pull/500 and https://github.com/pypa/setuptools_scm/pull/509 are the relevant PRs

RonnyPfannschmidt: for parity, as a followup

jamesbraza: I finally got around to this today, and concluded:

1. The term "distribution" is all over the Python docs (e.g. [`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html), [legacy install docs](https://docs.python.org/3.10/install/index.html), [`setuptools` docs](https://setuptools.pypa.io/en/latest/search.html?q=distribution&check_keywords=yes&area=default))
2. "dist" is also all over the place in the source code here

So I think it's fine to stick with the terminology "distribution" here after all. I am going to skip on opening a PR to change the wording from "dist" to "pkg"

PR #1004, opened 2024-01-18
Be nicer about possible misconfigurations

### LINKED ISSUE #1003: Be a smidge more error tolerant?
While playing around with using `setuptools_scm` I hit more than a few `AssertionError`s with the only context being the code and traceback.

Perhaps these could be improved with an assertion message, or even a `ValueError` instead :smile:

thejcannon:
Fixes https://github.com/pypa/setuptools_scm/issues/1003 by turning situations where the user might get an `AssertionError` and turn them into descriptive `ValueError`s.

Additionally, tests have been added to test the behavior.

The situations fixed are (with the real-world facepalm situations I ended up in):
- The entrypoint for `version_scheme` and `local_scheme` don't exist
  - :facepalm: I accidentally used the function name once, instead of the entrypoint name
- The entrypoint(s) for `version_scheme` and `local_scheme` don't return anything
  - :facepalm: I forgot to add a `return` to my custom scheme
- The `tag_regex` is bad, and was loaded from the config
  - :facepalm: This one should be self-explanatory
- The `tag_regex` matched, but the version was blank
  - :facepalm: My regex was borked


src/setuptools_scm/_config.py
```diff
@@ -41,9 +41,9 @@ def _check_tag_regex(value: str | Pattern[str] | None) -> Pattern[str]:
 
     group_names = regex.groupindex.keys()
     if regex.groups == 0 or (regex.groups > 1 and "version" not in group_names):
-        warnings.warn(
-            "Expected tag_regex to contain a single match group or a group named"
-            " 'version' to identify the version part of any tag."
+        raise ValueError(
```
RonnyPfannschmidt, line 46: this one is possibly a breaking change im not sure we can integrate it off hand

src/setuptools_scm/_config.py
```diff
@@ -105,6 +105,9 @@ class Configuration:
 
     parent: _t.PathT | None = None
 
+    def __post_init__(self) -> None:
+        self.tag_regex = _check_tag_regex(self.tag_regex)
```
RonnyPfannschmidt, line 110: taking this here without leaving it in place creates potental trouble for valies valid in config but not in the config objects

src/setuptools_scm/_config.py
```diff
@@ -139,13 +142,11 @@ def from_data(
         given configuration data
         create a config instance after validating tag regex/version class
         """
-        tag_regex = _check_tag_regex(data.pop("tag_regex", None))
```
RonnyPfannschmidt, line 142: this check needs to say even if the validation gets more struct

RonnyPfannschmidt (commented):
i like the idea behind how the details change, but i aslo beleive a larger change is necessary to do the path/version scheme handling in a better way

due to personal time constraints it may take a be a while to propperly reiterate this

thank you for getting the ball rolling and preparing this as a initial vesion

RonnyPfannschmidt:
i intend to merge this after hte next minor release as its a "breaking change" that should go in nonetheless

RonnyPfannschmidt:
i pushed a merge to validate with current code - i have to do a release before merging this

PR #1005, opened 2024-01-18
Add a "only-version" version scheme

### LINKED ISSUE #1002: Add a `just-the-tag` version scheme
Howdy! :wave: 

Love the tool/library. Hoping to use it for https://github.com/pantsbuild/pants versioning.

One thing is that we use `.devN` versions for the development series, so (most) of the version schemes that exist are incompatible if there's any kind of distance from the tag (the only ones that don't assert either create an illegal version like `.devN.postM` or blast the dev number).

Really, the distance is nice to know, but also isn't crucial to us, so meh.

thejcannon:
Fixes https://github.com/pypa/setuptools_scm/issues/1002 by adding a "only-version" version scheme.

### DIFF
--- a/testing/test_version.py
+++ b/testing/test_version.py
@@ -13,12 +13,12 @@ from setuptools_scm.version import calver_by_date
 from setuptools_scm.version import format_version
 from setuptools_scm.version import guess_next_date_ver
 from setuptools_scm.version import guess_next_version
+from setuptools_scm.version import just_version
 from setuptools_scm.version import meta
 from setuptools_scm.version import no_guess_dev_version
 from setuptools_scm.version import release_branch_semver_version
 from setuptools_scm.version import ScmVersion
 from setuptools_scm.version import simplified_semver_version
-from setuptools_scm.version import just_version
 
 
 c = Configuration()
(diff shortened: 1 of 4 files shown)


testing/test_version.py
```diff
@@ -169,6 +170,32 @@ def test_bump_dev_version_nonzero_raises() -> None:
     with pytest.raises(ValueError, match=match):
         guess_next_version(m("1.0.dev1"))
 
+@pytest.mark.parametrize(
+    "version",
+    [
+        "1.dev0",
```
thejcannon, line 177: Ripped from https://peps.python.org/pep-0440/

RonnyPfannschmidt (approved):
I believe the name should be only-tag and the docs need a warning that this means version is no longer pseudo unique per commit

Still adding a blanket approval

I'll try to get this in this morning or by the weekend

thejcannon:
Done and done.

RonnyPfannschmidt:
thanks, would you like to see this squashed or should i merge as is?

thejcannon:
Oh I'm definitely camp squash-merge. "Rename" isn't much of an inspiring commit message :joy:

thejcannon:
@RonnyPfannschmidt when might there be a new version of `setuptools_scm` to use?
Also, if there is going to be one soon, might it include https://github.com/pypa/setuptools_scm/pull/1004?

Thanks! 🙏

RonnyPfannschmidt:
I'll try to merge the clean up mr soon and releas this week

jelacious:
Here's a bump on getting that release up and going as I'd like to use `only-version`. <3

PR #1017, opened 2024-02-21
Fix non-escaped markdown in docs

JCGoran:
The docs presently render a bit strangely due to `_` not being escaped:

Screenshot 2024-02-21 at 08 11 44


RonnyPfannschmidt (approved):
good catch, thanks,

PR #1032, opened 2024-04-15
Use `typing.TYPE_CHECKING` in _own_version_helper.py

johnslavik:
Separated typing-only imports from runtime imports as part of my open-source mission to improve typing across Python packages.

### DIFF
--- a/_own_version_helper.py
+++ b/_own_version_helper.py
@@ -22,9 +22,9 @@ from setuptools_scm.version import guess_next_dev_version
 
 if TYPE_CHECKING:
     from collections.abc import Callable
-    
-    from setuptools_scm import _types as _t
+
     from setuptools_scm import Configuration
+    from setuptools_scm import _types as _t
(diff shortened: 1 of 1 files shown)


johnslavik:
If we get this merged, I'll continue to make similar changes to the actual codebase.

RonnyPfannschmidt (changes requested):
I don't consider this a improvement 

Extra conditions plus breakage of runtime inspection is not a win for me 

As far as I'm concerned this is both more complicated and less useful

johnslavik:
> Extra conditions plus breakage of runtime inspection is not a win for me

Is runtime inspection needed?

> As far as I'm concerned this is both more complicated and less useful

I see your point. Advantages of the practice are listed here: https://github.com/snok/flake8-type-checking#flake8-type-checking. If you believe it's unnecessary, I'll go ahead and close this.

RonnyPfannschmidt:
This one looks fundamentally misguided

PR #1033, opened 2024-04-16
Make .git_archival.txt more reproducible

### LINKED ISSUE #806: The documented `.git_archival.txt` causes git archives to change hash after first post-release commit is made
The documentation suggests using the following content for the `.git_archival.txt` file:

```
node: $Format:%H$
node-date: $Format:%cI$
describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
ref-names: $Format:%D$
```

This means that if a tag is created on top of the main branch, the file initially contains e.g.:

```
ref-names: HEAD -> main, tag: v1
```

but once another commit is added and `main` no longer corresponds to the tag, it changes to:

```
ref-names: tag: v1
```

This causes the git archive hash to change. This is a problem for distributions like Gentoo that are using git archives generated by GitHub. The initial archive that we fetch once the release is made changes once upstream makes new commits.

See e.g.: https://bugs.gentoo.org/895910, https://bugs.gentoo.org/895712

LecrisUT:
I believe the harm of `ref-names` outweigh the benefits of it, especially since this is mostly dependent on the git host provider that would produce the git archive.

Closes #806

### DIFF
--- a/.git_archival.txt
+++ b/.git_archival.txt
@@ -1,4 +1,3 @@
 node: $Format:%H$
 node-date: $Format:%cI$
 describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
-ref-names: $Format:%D$
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -235,8 +235,18 @@ Ensure the content of the following files:
 node: $Format:%H$
 node-date: $Format:%cI$
 describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
-ref-names: $Format:%D$
 ```
+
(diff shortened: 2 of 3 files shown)


RonnyPfannschmidt (approved):

RonnyPfannschmidt:
thanks, its unfortunate as this also looses the branch name (which is sometimes needed)
but the branch name was only ever included in the branch head to begin with

archuletacarla6-netizen (commented):
LecrisUT:fix/git_archival

RonnyPfannschmidt:
So open a pr?

PR #1034, opened 2024-04-17
Remove Warning/Error messages when using `build`

### LINKED ISSUE #987: Provide more guidance on how to use `.git_archival.txt`
I was working on switching a package from Versioneer to setuptools_scm, and wound up scratching my head over warnings about “unprocessed git archival found” when running `python -m build .`. It took me a while to understand what was going on, and I think some more guidance in the docs about how one should build a package when `.git_archival.txt` is present would have been helpful.

Specifically, if you add a `.git_archival.txt` file to a repository [as suggested in `docs/usage.md`](https://github.com/pypa/setuptools_scm/blob/main/docs/usage.md#git-archives) and then run `python -m build .`, you’ll see output like:

```
* Creating venv isolated environment...
* Installing packages in isolated environment... (setuptools>=64, setuptools_scm>=8)
* Getting build dependencies for sdist...
* Building sdist...
* Building wheel from sdist
* Creating venv isolated environment...
* Installing packages in isolated environment... (setuptools>=64, setuptools_scm>=8)
* Getting build dependencies for wheel...
/private/var/folders/jk/1hv06w454vj4q4rk2gl0zg800000gn/T/build-env-_6jt3k01/lib/python3.12/site-packages/setuptools_scm/git.py:308: UserWarning: git archive did not support describe output
  warnings.warn("git archive did not support describe output")
/private/var/folders/jk/1hv06w454vj4q4rk2gl0zg800000gn/T/build-env-_6jt3k01/lib/python3.12/site-packages/setuptools_scm/git.py:327: UserWarning: unprocessed git archival found (no export subst applied)
  warnings.warn("unprocessed git archival found (no export subst applied)")
ERROR setuptools_scm._file_finders.git listing git files failed - pretending there aren't any
* Installing packages in isolated environment... (wheel)
* Building wheel...
/private/var/folders/jk/1hv06w454vj4q4rk2gl0zg800000gn/T/build-env-_6jt3k01/lib/python3.12/site-packages/setuptools_scm/git.py:308: UserWarning: git archive did not support describe output
  warnings.warn("git archive did not support describe output")
/private/var/folders/jk/1hv06w454vj4q4rk2gl0zg800000gn/T/build-env-_6jt3k01/lib/python3.12/site-packages/setuptools_scm/git.py:327: UserWarning: unprocessed git archival found (no export subst applied)
  warnings.warn("unprocessed git archival found (no export subst applied)")
ERROR setuptools_scm._file_finders.git listing git files failed - pretending there aren't any
Successfully built mypkg-0.1.2.dev0+g8e7dc46.d20231204.tar.gz and mypkg-0.1.2.dev0+g8e7dc46.d20231204-py3-none-any.whl
```

I’ve cut some of the extra log lines to make the two warnings and errors clearer.

I’m pretty certain what’s happening here is that `build` creates an sdist, which works fine, but then unpacks that sdist in a temp directory to build the wheel from it, which is not fine. The unpacked sdist is not in a git repo, so setuptools-scm can’t find version info that way, and the sdist, which was build from my repo’s working directory, has a `.git_archival.txt` file that is just a template/unprocessed, so setuptools-scm is warning about that and then failing to get any useful versioning info.

That took me a little while to understand, and suggested to me that the *right* way to build a package when you have a `.git_archival.txt` file present is to do one of two things:
1. Ensure the file is not included in the build via `MANIFEST.in` or some config setting for whatever build backend the project is using, or
2. Always build from an actual archive. That is, a build process should do something like:

    ```sh
    git archive --output ../build_dir/source_archive.tar <commit-or-tag-to-build>
    cd ../build_dir
    tar -xf source_archive.tar
    rm source_archive.tar
    python -m build .
    ```

Do I have that right? If so, I think it would be helpful to add to the docs. I’m happy to make those changes in a PR if my understanding is correct.

(Obviously these specific results are caused by how `build` does its work, and other font-ends might or might not cause this warning. But I think the problem and solutions above are fairly generic and helpful to clarify. A lot of front-ends or build techniques would still wind up with the unprocessed `.git_archival.txt` file in an sdist if you don’t take one of the two above steps, and that doesn’t seem good.)

---

As a side note, I also found the dual warnings confusing:

```
/private/var/folders/jk/1hv06w454vj4q4rk2gl0zg800000gn/T/build-env-_6jt3k01/lib/python3.12/site-packages/setuptools_scm/git.py:308: UserWarning: git archive did not support describe output
  warnings.warn("git archive did not support describe output")
/private/var/folders/jk/1hv06w454vj4q4rk2gl0zg800000gn/T/build-env-_6jt3k01/lib/python3.12/site-packages/setuptools_scm/git.py:327: UserWarning: unprocessed git archival found (no export subst applied)
  warnings.warn("unprocessed git archival found (no export subst applied)")
```

And it wasn’t until I started searching the source here that I realized the first warning is not really correct and is kind of redundant. It would be clearer if only the second warning (“unprocessed git archival found”) was logged and the first (“git archive did not support describe output”) was suppressed in this case.

LecrisUT:
Basically when running `python3 -m build` on a perfectly valid project, we find ourselves surprised by unnecessary warning and error messages, namely:
```
UserWarning: git archive did not support describe output
UserWarning: unprocessed git archival found (no export subst applied)
ERROR setuptools_scm._file_finders.git listing git files failed - pretending there aren't any
```

See the linked issue for more context on why these messages occur. This PR is trying to eliminate those messages, but first I am writing a test to reproduce a common environment. Thanks to @Mr0grog for the help debugging this issue

TODO:
- [x] Add test to check `build` messages
- [ ] Resolve the git_archival.txt warning issue
- [ ] Resolve the setuptools_scm._file_finders.git error issue

Closes #987

### DIFF
--- a/testing/test_regressions.py
+++ b/testing/test_regressions.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import pprint
 import subprocess
 import sys
+import textwrap
 
 from dataclasses import replace
 from importlib.metadata import EntryPoint
@@ -16,6 +17,8 @@ from setuptools_scm._run_cmd import run
 from setuptools_scm.git import parse
 from setuptools_scm.integration import data_from_mime
 from setuptools_scm.version import meta
+
+from .wd_wrapper import WorkDir
 
 
 def test_data_from_mime_ignores_body() -> None:
@@ -95,6 +98,52 @@ def test_case_mismatch_on_windows_git(tmp_path: Path) -> None:
     run("git init", camel_case_path)
     res = parse(str(camel_case_path).lower(), Configuration())
     assert res is not None
+
+
+def test_fallback_message(wd: WorkDir, monkeypatch: pytest.MonkeyPatch) -> None:
+    """Check that no warning/error messages are normally outputted"""
+    # setup a basic project that would not trigger any (other) warning/errors
+    pyproject = """\
+        [build-system]
+        requires = ["setuptools", "setuptools_scm"]
+        build-backend = "setuptools.build_meta"
+
+        [project]
+        name = "example"
+        dynamic = ["version"]
+
+        [tool.setuptools_scm]
+        """
+    git_archival = """\
+        node: $Format:%H$
+        node-date: $Format:%cI$
+        describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
+        """
+    wd.write("pyproject.toml", textwrap.dedent(pyproject))
+    wd.write(".git_archival.txt", textwrap.dedent(git_archival))
+    wd.write(".gitattributes", ".git_archival.txt  export-subst")
+    wd.write("README.md", "")
+    wd("git init")
+    wd("git config user.email user@host")
+    wd("git config user.name user")
+    wd.add_command = "git add ."
+    wd.commit_command = "git commit -m test-{reason}"
+    wd.add_and_commit()
+    wd("git tag v0.1.0")
+    version = wd.get_version()
+    assert version == "0.1.0"
+
+    # Run build and check messages
+    monkeypatch.delenv("SETUPTOOLS_SCM_DEBUG")
+    # with caplog.at_level(logging.WARN):
+    res_build = subprocess.run(
+        [sys.executable, "-m", "build", "-nx"],
+        cwd=wd.cwd,
+        capture_output=True,
+    )
+    assert res_build.returncode == 0
+    # Maybe there is a cleaner way to use caplog?
+    assert not res_build.stderr
 
 
 def test_entrypoints_load() -> None:


testing/test_regressions.py
```diff
@@ -97,6 +100,52 @@ def test_case_mismatch_on_windows_git(tmp_path: Path) -> None:
...
+    monkeypatch.delenv("SETUPTOOLS_SCM_DEBUG")
+    # with caplog.at_level(logging.WARN):
+    res_build = subprocess.run(
+        [sys.executable, "-m", "build", "-nx"],
+        cwd=wd.cwd,
+        capture_output=True,
+    )
+    assert res_build.returncode == 0
+    # Maybe there is a cleaner way to use caplog?
+    assert not res_build.stderr
```
LecrisUT, line 146: Is there a better way to run and check this?

testing/test_regressions.py
```diff
@@ -97,6 +100,52 @@ def test_case_mismatch_on_windows_git(tmp_path: Path) -> None:
     assert res is not None
 
 
+def test_fallback_message(wd: WorkDir, monkeypatch: pytest.MonkeyPatch) -> None:
+    """Check that no warning/error messages are normally outputted"""
+    # setup a basic project that would not trigger any (other) warning/errors
+    pyproject = """\
```
LecrisUT, line 106: Is there a fixture to more easily set all of this up?

RonnyPfannschmidt: not yet , there has not been much work on e2e testing the build backend integrations

testing/test_regressions.py
RonnyPfannschmidt: A hack that might be reasonable would be a fake object directly passed to the integration function

LecrisUT: You mean instead of `python3 -m build`? I don't think that will end up well because what we want to test here is how `build` creates the sdist and from the sdist creates the wheel, and in both steps `setuptools-scm` is called with different "sources".

Another option I was thinking is calling [`build.__main__.main`](https://github.com/pypa/build/blob/0665ea731db6f431fb4f800b7f6849597c28677c/src/build/__main__.py#L391C5-L391C9) directly? Then all of the context should be preserved and the `caplog` should in principle work?

RonnyPfannschmidt: as far as i remember, build has to start sub-processes for a build
so currently  have no idea for a clean solution

RonnyPfannschmidt:
closing this to clean out my pipeline - happy to revive this if someone can work on it

PR #1044, opened 2024-05-06
Apply ruff/flake8-implicit-str-concat rule ISC001

DimitriPapadopoulos:
https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/

### DIFF
--- a/testing/test_regressions.py
+++ b/testing/test_regressions.py
@@ -35,7 +35,10 @@ def test_pkginfo_noscmroot(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> N
 
     tmp_path.joinpath(".git").mkdir()
     p.joinpath("setup.py").write_text(
-        "from setuptools import setup;" 'setup(use_scm_version={"root": ".."})',
+        """\
+from setuptools import setup
+setup(use_scm_version={"root": ".."})
+""",
         encoding="utf-8",
     )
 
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt:
At first glance all occurences where intentionally line split and had gotten cobbled to a single line, i  think multi line Strings are needed

DimitriPapadopoulos:
This is the result of applying black in #255, and `ruff format` will again fold the multiple lines into a single line — unless the code is enclosed in `# fmt: off` and `# fmt: on`.

Would you rather add `# fmt:` directives?

RonnyPfannschmidt:
I think real multi line Strings should be used

DimitriPapadopoulos:
I'll try to do that with `# fmt`.

RonnyPfannschmidt:
what i meant was using tripple quote string, not disabling fmt

testing/test_regressions.py
```diff
@@ -35,7 +35,8 @@ def test_pkginfo_noscmroot(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> N
 
     tmp_path.joinpath(".git").mkdir()
     p.joinpath("setup.py").write_text(
-        "from setuptools import setup;" 'setup(use_scm_version={"root": ".."})',
+        '''from setuptools import setup;
```
DimitriPapadopoulos: Do we want to keep `setup.py` as an one-liner?
```python
from setuptools import setup;setup(use_scm_version={"root": ".."})
```
Or perhaps change as follows?
```python
from setuptools import setup
setup(use_scm_version={"root": ".."})
```

RonnyPfannschmidt: Good idea

RonnyPfannschmidt:
thanks!

PR #1046, opened 2024-05-08
Exclude bots from generated release notes

hugovk:
The generated release notes at https://github.com/pypa/setuptools_scm/releases/tag/v8.1.0 are quite hard to read, they're nearly all dependency updates, for things that don't affect end user changes, making it hard to find the real end-user changes.

image

The good news is GitHub allows us to easily filter those out via `changelog.exclude.authors` in a config file:

https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuration-options

Let's use automation to improve the automation :) :robot: :rocket:

See https://github.com/tox-dev/tox-ini-fmt/pull/120 for another example.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,5 @@
+changelog:
+  exclude:
+    authors:
+      - dependabot
+      - pre-commit-ci


RonnyPfannschmidt (approved):

RonnyPfannschmidt:
Thanks!

PR #1055, opened 2024-07-09
Update README.md to correct CI badge status

shenxianpeng:
It seems that the current GitHub CI badge does not correctly reflect the true status.


RonnyPfannschmidt (approved):
thanks, i totally missed that one

PR #1069, opened 2024-08-23
docs: clarify usage of version files

ianw:
I found the current discussion around version files somewhat confusing.

When you're reading the usage guide, the second point is telling you to make a version file; strongly implying you should.  But in reality, you most likely do not need to bother with keeping a specific version file.  Explain more clearly about the implications of this.

Under this is a section on using `python -m setuptools_scm --help` which seems like an orphan from the "as a cli" section; move it down there.

The runtime section starts by telling you what is strongly discouraged without explaination why.  Make it clearer by giving the first example as the most supported thing, which is using standard importlib; then briefly discuss the version file from above, explaining what's in it for a standard Python template, then the strongly discouraged direct import method.

In the config doc, update to discuss `importlib` for consistency.

Correct some capitalization and other minor formatting bits while we're here.


RonnyPfannschmidt (approved):
Thank you for starting this 

The long term plan is to completely deprecate the exposed get-version functionality as people keep missusing and misunderstanding it

PR #1076, opened 2024-10-01
`version.py` including `__all__`

### LINKED ISSUE #1075: Include `__all__` in autogenerated `version_file `
With an autogenerated `version_file ` from `setuptools_scm==8.1.0`, it does not include an `__all__`.

Thus, my IDE imported `TYPE_CHECKING` from the `version.py` instead of from `typing`.

Can we include an `__all__` in the autogenerated `version.py` file so code inspection won't try to import `TYPE_CHECKING`

jamesbraza:
Closes https://github.com/pypa/setuptools-scm/issues/1075


src/setuptools_scm/_integration/dump_version.py
```diff
@@ -15,9 +15,13 @@
     ".py": """\
 # file generated by setuptools-scm
 # don't change, don't track in version control
+
+__all__ = ["__version__", "__version_tuple__", "version", "version_tuple"]
+
 TYPE_CHECKING = False
 if TYPE_CHECKING:
     from typing import Tuple, Union
+
```
jamesbraza, line 25: This would be done by `ruff` format if you ran it on the file, so I included it

RonnyPfannschmidt (changes requested):
lovely, thanks
lets add a changelog fragment as well

jamesbraza:
Nice, I just did it 👌 please let me know if I did it incorrectly

src/setuptools_scm/_integration/dump_version.py
RonnyPfannschmidt: https://github.com/pypa/setuptools-scm/blob/d665621e25f091c0de5cd35b3d57cc00bd26e260/testing/test_functions.py#L136-L153

needs to add ruff as followup

RonnyPfannschmidt (approved):

jamesbraza:
Thanks for the approval. I don't have the ability to merge this fyi

src/setuptools_scm/_integration/dump_version.py
jamesbraza: Thanks for the suggestion, I added a `ruff` test just now 👌 

Resolving this one

RonnyPfannschmidt:
the change log fragment was not matching the tooling, but its my falt it wasn't documented well enough - thanks for making it nice and doing the extra round-trip, to ensure its awesome :+1:

PR #1077, opened 2024-10-02
Disable deprecated/problematic ruff rules

DimitriPapadopoulos:
UP027: deprecated
UP038: https://github.com/astral-sh/ruff/issues/7871


RonnyPfannschmidt:
if we don't fix/revert the changes those might have done, its less maintenance to just wait for ruff to fix the stuff with own rules later

as far as i can tell we are not affected by the performance impact, and this pr doesn't change anything

instead of manually searching for fixes, i'd prefer to just eventually have a ruff update fix it automatically in future

DimitriPapadopoulos:
It's just that I fell bad for importing these bad rules here :smile:

RonnyPfannschmidt:
i understand the sentiment, luckily in the grand scheme of things the impact doesn't matter as much as it feels and we can just wait it out ;)

PR #1080, opened 2024-10-23
Better cleanup PYTHONPATH for Mercurial calls (support for PDM)

paugier:
PDM uses the same strategy as pip for the build isolation except that the path of the temporary directory starts with "pdm-build-env-" and not "pip-build-env-".

Therefore, Mercurial (a Python application) installations can be broken and we need to clean up `PYTHONPATH` from such paths.

### DIFF
--- a/src/setuptools_scm/_run_cmd.py
+++ b/src/setuptools_scm/_run_cmd.py
@@ -117,7 +117,7 @@ def avoid_pip_isolation(env: Mapping[str, str]) -> dict[str, str]:
         [
             path
             for path in new_env["PYTHONPATH"].split(os.pathsep)
-            if "pip-build-env-" not in path
+            if "-build-env-" not in path
         ]
     )
     return new_env


RonnyPfannschmidt (approved):
i wonder if there is a good integration test for this

paugier:
I don't think there is a specific test about this mechanism. However, `setuptools_scm._run_cmd.run` is executed during the tests. I guess that setuptools_scm is tested with an installation with pip. If it is the case, this line is covered.

A problem is that pep 517 does not specify how the frontends implement the isolated build (https://peps.python.org/pep-0517/#build-environment), so we can just try to avoid known issues.

paugier:
Thanks @RonnyPfannschmidt. Do you think a new version could be released on PyPI? I see that the last version was released on May 2024.

RonnyPfannschmidt:
I'll try, but no promises, I'm suffering a stomach infection

paugier:
Oh I hope you will recover soon. Be brave.

I'm working on modernizing Mercurial setup.py and I'd like to try to use setuptools-scm. But of course your health is more important than open-source!

Beside, do not hesitate to tell me if I can help on anything for setuptools-scm.

PR #1081, opened 2024-10-25
pipx.run entry-points should be setuptools-scm

paugier:
The package is now called setuptools-scm.

`-` are more common than `_` for command names (at least on Linux).

I guess it make more sense to write `pipx run setuptools-scm`.

Note: currently (before next release), this works

```
pipx run --spec setuptools-scm@https://github.com/pypa/setuptools-scm/archive/refs/heads/main.zip setuptools_scm
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,7 @@ repository = "https://github.com/pypa/setuptools-scm/"
 [project.entry-points."distutils.setup_keywords"]
 use_scm_version = "setuptools_scm._integration.setuptools:version_keyword"
 [project.entry-points."pipx.run"]
+setuptools-scm = "setuptools_scm._cli:main"
 setuptools_scm = "setuptools_scm._cli:main"
 [project.entry-points."setuptools.file_finders"]
 setuptools_scm = "setuptools_scm._file_finders:find_files"


paugier:
Note for myself: this works

```
pipx run -v --spec setuptools-scm@git+https://github.com/paugier/setuptools_scm.git@fix-pipx-command-name setuptools-scm
```

RonnyPfannschmidt:
Based in normalization I suspect we will need the old name

I think pipx needs to support normalization flips before we can drop it

pyproject.toml
```diff
@@ -72,7 +72,7 @@ repository = "https://github.com/pypa/setuptools-scm/"
 [project.entry-points."distutils.setup_keywords"]
 use_scm_version = "setuptools_scm._integration.setuptools:version_keyword"
 [project.entry-points."pipx.run"]
-setuptools_scm = "setuptools_scm._cli:main"
+setuptools-scm = "setuptools_scm._cli:main"
```
RonnyPfannschmidt, line 75:
```suggestion
setuptools-scm = "setuptools_scm._cli:main"
setuptools_scm = "setuptools_scm._cli:main"
```

RonnyPfannschmidt (changes requested):
my understanding is that we need both-until normalization is supported

pyproject.toml
paugier: Ah ok thanks. I rebased and amended the commit with this change.

RonnyPfannschmidt (approved):

PR #1087, opened 2024-11-18
Avoid non-monotonic versions

### LINKED ISSUE #1025: Non-monotonic development versions with release branch scheme
I use `setuptools_scm` for versioning hg-git, with the `release-branch-semver` scheme, where I have release branches named e.g. `1.0.x` and `1.1.x`. When doing a stable release, I do it on the release branch and merge it into `default`. Unfortunately, the `devN` is merely calculated as the length to the most recent tag, which then makes it _decrease_ after a minor release is merged into the main development branch.

I checked, and I can reproduce this using both [Mercurial](http://foss.heptapod.net/mercurial/hg-git/-/pipelines) and [Git](https://github.com/danchr/hg-git) sources. For example, the version determined relative to `1.1.0` is `1.2.0.dev26+g7069705` whereas after the merge, it.s `1.2.0.dev17+g5b33fad`.

As best I can tell, you cannot easily customise the `devN` part. For `release-branch-semver`, I believe it should either:

1) Use some monotonically increasing value, like the date or a timestamp.
2) Somehow take the minor into account, e.g. by multiplying the distance by 100 or 1000 times the minor release number.

danchr:
Using the `release-branch-semver` scheme, a commit five nodes off e.g. `1.0.0` will get `1.1.0dev5`, but a commit four nodes off `1.0.1` will get `1.1.0dev4`, despite it being more commits from `1.0.0`. To fix this, we add a factor that we use for bumping the distance based on the micro release we're discarding. The default is `1000`; for most projects, that should be _plenty_ of commits per bug-fix release.

Fixes #1025.

### DIFF
--- a/testing/test_mercurial.py
+++ b/testing/test_mercurial.py
@@ -32,7 +32,12 @@ def wd(wd: WorkDir) -> WorkDir:
 archival_mapping = {
     "1.0": {"tag": "1.0"},
     "1.1.0.dev3+h000000000000": {
-        "latesttag": "1.0",
+        "latesttag": "1.0.0",
+        "latesttagdistance": "3",
+        "node": "0" * 20,
+    },
+    "1.1.0.dev103+h000000000000": {
+        "latesttag": "1.0.1",
         "latesttagdistance": "3",
         "node": "0" * 20,
     },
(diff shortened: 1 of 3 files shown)


testing/test_mercurial.py
```diff
@@ -32,7 +32,12 @@ def wd(wd: WorkDir) -> WorkDir:
 archival_mapping = {
     "1.0": {"tag": "1.0"},
     "1.1.0.dev3+h000000000000": {
-        "latesttag": "1.0",
+        "latesttag": "1.0.0",
+        "latesttagdistance": "3",
```
RonnyPfannschmidt, line 36: i need some more examples, i cant quite wrap my head around it

this is a orkaround to ease the pain, but its also unable to completely prevent it

Issue #921, opened 2023-09-20
Consistency of documentation of what version of setuptools is required


ksunden:
Currently there are references to the required version being:

- `>=60`
  - [README.md](https://github.com/pypa/setuptools_scm/blob/ceca2352f09ed22f825f45a392cca5aff83c6aff/README.md?plain=1#L29)
- `>=61`
  - [runtime warning](https://github.com/pypa/setuptools_scm/blob/ceca2352f09ed22f825f45a392cca5aff83c6aff/src/setuptools_scm/_integration/setuptools.py#L37)
  - [your own pyproject.toml buld requirements](https://github.com/pypa/setuptools_scm/blob/ceca2352f09ed22f825f45a392cca5aff83c6aff/pyproject.toml#L9)
  - [your test workflow](https://github.com/pypa/setuptools_scm/blob/ceca2352f09ed22f825f45a392cca5aff83c6aff/.github/workflows/python-tests.yml#L82)
- `>=62`
  - [docs/usage.md](https://github.com/pypa/setuptools_scm/blob/ceca2352f09ed22f825f45a392cca5aff83c6aff/docs/usage.md?plain=1#L13)
- `>=64`
  - [docs/index.md](https://github.com/pypa/setuptools_scm/blob/ceca2352f09ed22f825f45a392cca5aff83c6aff/docs/index.md?plain=1#L24)
  - [docs/customizing.md](https://github.com/pypa/setuptools_scm/blob/ceca2352f09ed22f825f45a392cca5aff83c6aff/docs/index.md?plain=1#L24)
  - [docs/examples/.../pyproject.toml](https://github.com/pypa/setuptools_scm/blob/ceca2352f09ed22f825f45a392cca5aff83c6aff/docs/examples/version_scheme_code/pyproject.toml#L4)

At various places in the docs/code.

Additionally, there is a comment that states:

https://github.com/pypa/setuptools_scm/blob/ceca2352f09ed22f825f45a392cca5aff83c6aff/docs/index.md?plain=1#L16-L17

However, the `pyproject.toml` _does_ include `setuptools` as a runtime dependency, but does so with no pin, allowing broken environments to exist:

https://github.com/pypa/setuptools_scm/blob/ceca2352f09ed22f825f45a392cca5aff83c6aff/pyproject.toml#L48

Making these sources consistent would be appreciated, as it is rather hard to tell what I actually need to do as a user.

[...]

PR #1088, opened 2024-11-24
Add a lower bound pin on setuptools dependency

mayeut:
mitigates #921

The docs and the checks are still inconsistent (setuptools>=61 & setuptools>=64 are used depending on checks or doc).
This aligns the runtime dependency with the check.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dynamic = [
 ]
 dependencies = [
   "packaging>=20",
-  "setuptools",
+  "setuptools>=61",
   'tomli>=1; python_version < "3.11"',
   'typing-extensions; python_version < "3.10"',
 ]


RonnyPfannschmidt (approved):

mayeut:
re-reading #921, the doc says:
https://github.com/pypa/setuptools-scm/blob/3c145dfd784e41cdb774449a421fe5e17a664786/docs/index.md?plain=1#L19-L20

If the intent of leaving a lower bound blank was for this, then this PR should probably be closed while waiting for one that actually removes the dependency and/or update the check to only run when called from setuptools entrypoints ?

RonnyPfannschmidt (approved):

RonnyPfannschmidt:
a new package name will bbe used for non setuptools dependency in future as it tuns out build env configuration is fragile enough already without setuptools_scm adding a extra dimension of error

PR #1094, opened 2025-01-03
Improve the docs about `version_scheme` and `local_scheme`

maresb:
I added cross-references to the implementations. it was confusing to me before that the available options weren't listed in `config.md`. See comments in individual commits for more details.


RonnyPfannschmidt (commented):
Thanks!

PR #1105, opened 2025-02-13
Clean up _git_toplevel()

rossburton:
Instead of asking git what the relative path of the current directory is from the root of the repository and then reverse-engineering the root, simply call git rev-parse --show-toplevel.

This option was added in git 1.7.0 back in 2010, so I think it's fair to assume it exists.


rossburton:
Love it when a test suite works locally but fails in CI.

RonnyPfannschmidt:
I vaguely recall pain and suffering from windows path normalization 

It may be necessary to keep the cutoff of the tailing slashes and to use normalization

RonnyPfannschmidt:
closing the windows hack is unfortunately necessary

PR #1106, opened 2025-02-13
respect GIT_CEILING_DIRECTORIES

Ecordonnier:
Fix for https://github.com/pypa/setuptools-scm/issues/1103

When searching for the root-directory of the git repository e.g. with git rev-parse --show-toplevel, git stops the search when reaching $GIT_CEILING_DIRECTORIES. By ignoring this variable, the function _git_toplevel can go above the real git repository (e.g. when packaging a tarball without .git repository), and then runs "git archive" on an unrelated git repository.


Ecordonnier:
Note that unfortunately the log.info() call I added is invisible while running as part of a yocto build in the use-case I described in https://github.com/pypa/setuptools-scm/issues/1103 . It seems that only log-levels at warning or above are shown in this use-case, however this is not really a warning, so I left it at info level.

RonnyPfannschmidt (approved):

Ecordonnier:
@rossburton FYI

PR #1109, opened 2025-02-22
Apply more ruff rules

DimitriPapadopoulos:


src/setuptools_scm/_log.py
```diff
@@ -18,7 +18,7 @@
 
 
 class AlwaysStdErrHandler(logging.StreamHandler):  # type: ignore[type-arg]
-    def __init___(self) -> None:
```
DimitriPapadopoulos, line 21: This was a real bug, by the way, wasn't it?

RonnyPfannschmidt: I suspect it was completely missed due to working fine outside of pytest and using rich in testing

PR #1113, opened 2025-02-24
Enforce more ruff rules

DimitriPapadopoulos:
Add these rules:
* [flake8-comprehensions (C4)](https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4)
* [flake8-implicit-str-concat (ISC)](https://docs.astral.sh/ruff/rules/#flake8-implicit-str-concat-isc)
* [flake8-logging (LOG)](https://docs.astral.sh/ruff/rules/#flake8-logging-log)
* [flake8-logging-format (G)](https://docs.astral.sh/ruff/rules/#flake8-logging-format-g)
* [flake8-pie (PIE)](https://docs.astral.sh/ruff/rules/#flake8-pie-pie)
* [flynt (FLY)](https://docs.astral.sh/ruff/rules/#flynt-fly)
* [mccabe (C90)](https://docs.astral.sh/ruff/rules/#mccabe-c90)
* [Perflint (PERF)](https://docs.astral.sh/ruff/rules/#perflint-perf)
* [pycodestyle (W) ](https://docs.astral.sh/ruff/rules/#warning-w)
* [pygrep-hooks (PGH)](https://docs.astral.sh/ruff/rules/#pygrep-hooks-pgh)
* [Pylint (PLE)](https://docs.astral.sh/ruff/rules/#error-ple)

Reorder rules, same order as [ruff documentation](https://docs.astral.sh/ruff/rules/).

PR #1114, opened 2025-02-24
Python 3.13 (more)

DimitriPapadopoulos:
~~It's unclear why pre-commit wants to discard Python 3.13 support. I cannot reproduce locally.~~

The problem is that the [pyproject-fmt](https://github.com/tox-dev/pyproject-fmt) hook version is outdated. Either you accept version 2 and its formatting issues, or you get rid of this hook.

Requires 448838e from #1113 to fix the CI error.

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -34,7 +34,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


RonnyPfannschmidt (approved):

RonnyPfannschmidt:
The hook has to go

blg-666 (approved):

RonnyPfannschmidt:
Thanks!

PR #1116, opened 2025-03-13
Update minimum dependencies on setuptools

do3cc:
In another project that has an indirect dependency on setuptools-scm, I am trying to use a feature of uv for CI testing, where it will install the minimum support version of each dependency. This works surprisingly well, with a few minor exceptions.
I have tested the actual minimum versions of setuptools for setuptools-scm for python version 3.9 to 3.13 and added the information.
I would prefer to just up the minimum version for all python versions to 67, however I suspect that this package is sufficient popular that it will be used in linux distribution packages and getting updated setuptools versions are hard get. So for the PR, I created different pinnings for different python versions. But I am more than happy to change this to one version that works for all supported python versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,8 @@
 [build-system]
 build-backend = "_own_version_helper:build_meta"
 requires = [
-  "setuptools>=61",
+  "setuptools>=64",
+  "setuptools>=67;python_version>='3.12'",
   'tomli<=2.0.2; python_version < "3.11"',
 ]
 backend-path = [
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
We can't update the dependency easily 

We may actually have to drop the minimum version due to bugs in Ubuntu lts

do3cc:
Ugh, Linux distributions and their promises for long time support on the back of others :-(

Thank you, I will close this, I don't want to cause large scale breakage for this.

RonnyPfannschmidt:
There will eventually be releases that force breakage on broken Linux

PR #1124, opened 2025-03-24
wip: add test for build behaviour across python/setuptools versions

RonnyPfannschmidt:
- [ ] trim down matrix to significant combinations
- [ ] re-validate order of keyword/finalize hooks 
- [ ] validate versions in envs
- [ ] make it pretty


RonnyPfannschmidt:
discontinuing this effort - i learned its a combination of importlib.metadata in old python

Issue #1131, opened 2025-04-22
8.3.0 introduces dependency loop with importlib_metadata on Python < 3.10


legoktm:
Hi, thanks for working on and maintaining setuptools-scm.

fc2d84179c4e8bc7123ba0b34f692dd029e3d4bd introduced a dependency on importlib_metadata for Python < 3.10, which causes a dependency loop when building from source, because importlib_metadata depends on setuptools-scm. From what I can tell this is the same loop that was identified as problematic in #722 (2022), and then again in #918 (2023). Maybe there needs to be some stronger mechanism to prevent it in the future?

A simple reproducer (some output omitted to make it more readable):
```
$ podman run --rm -it python:3.9 bash -c "pip install --no-binary=:all: --no-cache-dir setuptools-scm==8.3.0"
Collecting setuptools-scm==8.3.0
  Downloading setuptools_scm-8.3.0.tar.gz (78 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 78.0/78.0 kB 4.4 MB/s eta 0:00:00
  Installing build dependencies ... error
  error: subprocess-exited-with-error
  
  × pip subprocess to install build dependencies did not run successfully.
  │ exit code: 1
  ╰─> [99 lines of output]
      ...
      Collecting importlib-metadata>=4.6
        Downloading importlib_metadata-8.6.1.tar.gz (55 kB)
           ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 55.8/55.8 kB 13.0 MB/s eta 0:00:00
        Installing build dependencies: started
        Installing build dependencies: finished with status 'error'
        error: subprocess-exited-with-error
      
        × pip subprocess to install build dependencies did not run successfully.
        │ exit code: 2
        ╰─> [60 lines of output]
            Collecting setuptools_scm[toml]>=3.4.1
              Downloading setuptools_scm-8.3.0.tar.gz (78 kB)
                 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 78.0/78.0 kB 2.8 MB/s eta 0:00:00
            ERROR: Exception:
            Traceback (most recent call last):
              ...
              File "/usr/local/lib/python3.9/site-packages/pip/_internal/operations/build/build_tracker.py", line 92, in add
                raise LookupError(message)
[...]

PR #1133, opened 2025-04-23
Add warning for people adding dependencies

legoktm:
To hopefully prevent regressions like #1131 from happening again.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,8 @@
 
 [build-system]
 build-backend = "_own_version_helper:build_meta"
+# Be careful that dependencies added here do not also depend upon setuptools-scm,
+# otherwise it'll cause problems with builds from source (e.g. #722, #918, #1131)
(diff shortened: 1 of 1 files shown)


legoktm:
Ideally there would be a CI check of some kind that tried building from source to catch these issues, but this is hopefully a first step.

RonnyPfannschmidt:
In near future there will be new dependencies 

The bootstrap issue will get a solution

legoktm:
OK, I'm curious as to what that will look like :) But in the meantime, are you saying you don't want this patch? At least to me it seems like a simple solution to temporarily avoid more regressions.

RonnyPfannschmidt:
im closing this one as its a particular detail of the build pipeline

the solution will be a build backend for build-system deps that hopefully allows sdists that can self-install with only that dependency

and that particular dependencies sdist being able to self-contained self-building

PR #1136, opened 2025-05-04
fix: update customizing.md to fix missing import

### LINKED ISSUE #1135: The `Importing in setup.py` example in `customizing.md` contains a missing import
The code block at https://github.com/pypa/setuptools-scm/blob/main/docs/customizing.md?plain=1#L57
I believe should be:
``` { .python title="setup.py" }
from setuptools import setup
from setuptools_scm.version import get_local_dirty_tag

def clean_scheme(version):
    return get_local_dirty_tag(version) if version.dirty else '+clean'

setup(use_scm_version={'local_scheme': clean_scheme})
```

shenxianpeng:
closes #1135

### DIFF
--- a/docs/customizing.md
+++ b/docs/customizing.md
@@ -55,7 +55,7 @@ dynamic = [
 With the pep 517/518 build backend, setuptools-scm is importable from `setup.py`
 
 ``` { .python title="setup.py" }
-import setuptools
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt (approved):
Thanks

PR #1143, opened 2025-06-11
chore: remove redundant call to require_command()

manueljacob:
In 560a1cb5ff3a3c052e3641034476a3b0554edb86, the call to require_command() was moved from HgWorkdir.from_potential_worktree() to the caller. That caller is the only caller of GitWorkdirHgClient.from_potential_worktree() as well, so it is redundant and can be removed.


RonnyPfannschmidt (approved):
Thanks nice catch

PR #1144, opened 2025-06-11
Make Mercurial command configurable by an environment variable

manueljacob:


manueljacob:
Of course, I’m not quite happy with reloading the modules in the test. But it allows the non-test code to stay simple. Another possibility would be to monkey-patch the module-level constants, but I prefer testing the code like it is used by the user (who sets the environment variable) instead of depending on implementation details.

RonnyPfannschmidt:
Why is it necessary to set a different hg command? 

I need a surrounding context

manueljacob:
See the commit messages.

I’ll usually put motivations and other long explanations in the commit message for long-term record. If the pull request contains more than one commit, GitHub won’t automatically copy the commit message to the pull request description.

RonnyPfannschmidt:
Let's add a changelog entry for this and a example in the docs so users can quickly grasp it

manueljacob:
Okay, I will push a changelog entry. BTW, it was not obvious to me what the format of changelog entries in `changelog.d` is, until I happened to come across the changelog entry “introduce scriv for changelog management”.

What would be a good example for the docs? I think developing Mercurial extensions is a bit too specific, so I guess it should be using `chg` to speed things up?

RonnyPfannschmidt:
In case of doubt ignore scriv i messed up correctly using it the last few releases 

I'll need a release prep bot to get that out of the door correctly

manueljacob:
> I'll need a release prep bot to get that out of the door correctly

You automated package version generation. Shouldn’t be too hard? ;)

RonnyPfannschmidt:
With github actions its a pain in the neck 

Like multiple 3rd party actions due to platform shortcomings 

At least its not gitlab

manueljacob:
> What would be a good example for the docs? I think developing Mercurial extensions is a bit too specific, so I guess it should be using `chg` to speed things up?

@RonnyPfannschmidt Any opinion on the example? Otherwise, I’ll go ahead with the `chg` one.

RonnyPfannschmidt:
Chg is probably the best example 

Hevk it may make sense to use it as default if avaliable (with development caveats)

manueljacob:
I pushed the updated documentation, adding a short motivation.

We can separately discuss using `chg` by default.

RonnyPfannschmidt (approved):
Thanks

src/setuptools_scm/hg_git.py
```diff
@@ -11,6 +11,7 @@
 from ._run_cmd import CompletedProcess as _CompletedProcess
 from ._run_cmd import run as _run
 from .git import GitWorkdir
+from .hg import HG_COMMAND
```
RonnyPfannschmidt, line 14: Let's turn this into a module import to ease test instrumentation

testing/test_hg_git.py
```diff
@@ -81,3 +88,34 @@ def test_base(repositories_hg_git: tuple[WorkDir, WorkDir]) -> None:
...
+
+
+def test_hg_command_from_env(
+    repositories_hg_git: tuple[WorkDir, WorkDir],
+    monkeypatch: pytest.MonkeyPatch,
+    request: pytest.FixtureRequest,
+    hg_exe: str,
+) -> None:
+    wd = repositories_hg_git[0]
+    with monkeypatch.context() as m:
```
RonnyPfannschmidt, line 113: Instead of reloading rhe modules lets turn the logic into a function the test can call

RonnyPfannschmidt:
I plan to merge  Monday as I'm mid travel

PR #1154, opened 2025-07-24
Add commit_id to dump_file template

JulianFP:
For my use case I want to report not only the current version of my application but also the git hash at which it was built. setuptools-scm includes the git hash if the version isn't exact (i.e. the application wasn't built from a commit with a git tag), but not if it is. This behavior on it's own is fine and expected imo. However I still needed an easy way to access the git hash in these cases in addition to the version, so I maid this small adjustment to the dump_file template so that it always includes the current git_hash in addition the the version and version_tuple, regardless of whether the version or version_tuple includes the git hash already or not.
I am currently already using this change for my own packages project-W and project-W-runner, but maybe they are useful to other users as well so I decided to create this PR.


RonnyPfannschmidt (changes requested):
i like the idea - please fix the tests and propose a change-log entry (i'll add it to the change-log)

JulianFP:
> i like the idea - please fix the tests and propose a change-log entry (i'll add it to the change-log)

Fixed the tests. I also renamed the new attributes to `sha` and `__sha__` since these terms are independent from the scm in use (and interestingly the template used inside the tests already included these attributes). I can always change them back to `git_hash` or something else like `revision_hash` if you prefer.

Proposed change-log entry:
Add `sha` and `__sha__` attributes to the dump_file template for access to the revision hash in addition to the version

testing/test_basic_api.py
```diff
@@ -184,9 +184,11 @@ def read(name: str) -> str:
     scm_version = meta("1.0", distance=42, config=c)
     dump_version(tmp_path, version, "first.py", scm_version=scm_version)
     lines = read("first.py").splitlines()
-    assert lines[-2:] == [
+    assert lines[-4:] == [
         "__version__ = version = '1.0.dev42'",
         "__version_tuple__ = version_tuple = (1, 0, 'dev42')",
+        "",
```
RonnyPfannschmidt, line 190: This should get a different test

testing/test_basic_api.py
```diff
@@ -23,7 +23,7 @@
 template = """\
 __version__ = version = {version!r}
 __version_tuple__ = version_tuple = {version_tuple!r}
-__sha__ = {scm_version.node!r}
+__sha__ = sha = {scm_version.node!r}
```
RonnyPfannschmidt: The template is intentionally different from the normal template 

If its the same using it in test no longer shows different behavior

src/setuptools_scm/_integration/dump_version.py
```diff
@@ -31,9 +31,13 @@
 __version__: str
 __version_tuple__: VERSION_TUPLE
 version_tuple: VERSION_TUPLE
+sha: str
```
RonnyPfannschmidt: Lets use the name node_id or commit_id 

Just sha is actually incorrect for scm like jj pujil or darcs

RonnyPfannschmidt (changes requested):
Thanks unfortunately it seems like that one test missed in naming 

Perhaps i should have given the template in the test a more detailed name 

I also took note that there's potential situations where nodeid isn't known 

I believe those where the reason for not including it in the beginning 

But I'm unaware of a easy acceptance test for that so i need to log a follow-up

testing/test_basic_api.py
JulianFP: I agree, I added a second test that doesn't use the test template but where the commit id is not None

testing/test_basic_api.py
JulianFP: Ah ok understood, I changed it back

src/setuptools_scm/_integration/dump_version.py
JulianFP: Alright, I changed it to `commit_id`

JulianFP:
> I also took note that there's potential situations where nodeid isn't known
> 
> I believe those where the reason for not including it in the beginning

I changed the type of the attribute to `Union[str, None]`. This way the application can handle the case where the nodeid isn't known how they see fit. I think that might be the best solution to handle this?

RonnyPfannschmidt (approved):

RonnyPfannschmidt:
well done, i'll merge + add a changelog entry after i finish sorting out my other pr

JulianFP:
> well done, i'll merge + add a changelog entry after i finish sorting out my other pr

Alright, thank you!

PR #1166, opened 2025-07-31
PEP 639 compliance

DimitriPapadopoulos:
See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,6 @@ name = "setuptools-scm"
 description = "the blessed package to manage your versions by scm tags"
 readme = "README.md"
 license = "MIT"
-license-files = ["LICENSE"]
 authors = [
     {name="Ronny Pfannschmidt", email="opensource@ronnypfannschmidt.de"}
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@
 [build-system]
 build-backend = "_own_version_helper:build_meta"
 requires = [
-  "setuptools>=61",
+  "setuptools>=77.0.3",
   'tomli<=2.0.2; python_version < "3.11"',
 ]
 backend-path = [
@@ -15,7 +15,8 @@ backend-path = [
 name = "setuptools-scm"
 description = "the blessed package to manage your versions by scm tags"
 readme = "README.md"
-license.file = "LICENSE"
+license = "MIT"
+license-files = ["LICENSE"]
 authors = [
     {name="Ronny Pfannschmidt", email="opensource@ronnypfannschmidt.de"}
 ]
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt:
windows errors are unrelated and fixed in another pr

RonnyPfannschmidt:
this implicitly makes the project python 3.9 plus - so it will have to wait until after the 8.4 release as i want to ensure bugfixes for build issues with python3.8 and 3.9 deliver to down-streams/enterprise distro users

DimitriPapadopoulos:
Additionally, part of the ecosystem (`pip-licenses`) doesn't support PEP 639 yet.

pyproject.toml
```diff
@@ -15,15 +15,15 @@ backend-path = [
 name = "setuptools-scm"
 description = "the blessed package to manage your versions by scm tags"
 readme = "README.md"
-license.file = "LICENSE"
+license = "MIT"
+license-files = ["LICENSE"]
```
DimitriPapadopoulos: I believe setuptools does not need an explicit `license-files` key according to [Controlling files in the distribution](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html):
> * All files specified by the `license-files` configuration parameter in `pyproject.toml` and/or equivalent in `setup.cfg`/`setup.py`; note that if you don’t explicitly set this parameter, `setuptools` will include any files that match the following glob patterns: `LICEN[CS]E*`, `COPYING*`, `NOTICE*`, `AUTHORS**`;

@RonnyPfannschmidt Do you prefer implicit or explicit?

RonnyPfannschmidt: implicit please

RonnyPfannschmidt (approved):
thanks for preparing this - the failing tests are from python versions to remove

RonnyPfannschmidt:
afte the next release i'll take this into the branch for the next version

pyproject.toml
DimitriPapadopoulos: Perhaps implicit is not a good idea after all. From [Appendix: Rejected Ideas](https://peps.python.org/pep-0639/appendix-rejected-ideas/#use-a-default-value-for-license-files-if-not-specified):
> ### Use a default value for `license-files` if not specified
> 
> A previous draft of the PEP proposed a default value for detecting license files in case the users have not declared any and not marked the key as dynamic. That value was defined as an array of globs:
> ````
> ["LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*"]
> ````
> 
> However, this would create an exception among the existing metadata, as no other key has got implicit defaults defined. Implicit values in `pyproject.toml` keys are delegated to the dynamic field, which is specified as being calculated. Also, the values were chosen arbitrarily, without a strong justification why they should pose a standard.

The implicit default values look like a leftover from an experimental feature that eventually got rejected.

RonnyPfannschmidt:
i merged this in my prep branch, please no longer force push as that would disconnect history

PR #1235, opened 2025-10-31
git: Exit clearly when in a "dubious" directory

### LINKED ISSUE #784: git rev-parse HEAD failure should sometimes be fatal
# Background

On github actions when running in a container, by default (at least in my container) the build directory isn't considered 'safe' by git because it has 'dubious ownership'. As a result, `git rev-parse HEAD` and other commands fail, which resulted in my wheel being created without some of the package_data files that I expected to be there. Naturally, it took a very long time to work out what exactly was happening and why.

Reference: https://github.com/robotpy/robotpy-wpimath/actions/runs/3451821670/jobs/5761310334, see cross-build ` robotpy/build-actions/build-wheel@v2023`:

```
 ----
cmd:
     git rev-parse HEAD
 in: /__w/robotpy-wpimath/robotpy-wpimath
err:
     fatal: detected dubious ownership in repository at '/__w/robotpy-wpimath/robotpy-wpimath'
    To add an exception for this directory, call:

    	git config --global --add safe.directory /__w/robotpy-wpimath/robotpy-wpimath

ret: 128
listing git files failed - pretending there aren't any
```

# Feature request

It seems to me that some of the commands that setuptools-scm does are totally optional -- for example, determining a version in a shallow checkout. Some things shouldn't be optional -- if listing files fails it seems like that should be a fatal error if we are building from a valid git repository -- then that lets the user know something weird is happening so they can address it, instead of failing silently.

If that's not acceptable for whatever reason, then I propose that an environment variable be added to force failure in these cases.

Happy to make a PR for either case.

ianw:
I hit a _very_ hard to diagnose issue today after switching part of a CI environment.  It turns out that due to non-privileged container magic the CI step was not running as the user that cloned the repo. This manifested as the sdist generated during this step just not having a few extra data files; but the build still passed.  So several steps later, suddenly files have just vanished from the sdist, which of course is not something that is checked for and so testing blows up in very strange ways.

After I understood what was going on, there is a _tiny_ little message hidden amongst the logs that gives a hint of what's going on.

```
...
writing requirements to src/proj.egg-info/requires.txt 
writing top-level names to src/proj.egg-info/top_level.txt 
listing git files failed - pretending there aren't any 
reading manifest file 'src/proj.egg-info/SOURCES.txt' 
writing manifest file 'src/proj.egg-info/SOURCES.txt' 
...
```

What is actually happening is that if you run git you get `git status fatal: detected dubious ownership in repository at '/..proj'`.  This is the well-known CVE-2022-24765 issue where trusting a `.git` config dir from another user causes problems.  In 6a3bb967167545a2ca074ed56a8e4c2280946325 all the calls in `setuptools_scm/git.py` were updated to use `--git-dir` directly -- git will not complain if you have told it to explicitly trust the config dir like this.

However, there are calls in `_file_finders/git.py` to find the top-level that are not using this.  It silently (modulo an easily missed log) skips adding files when this occurs.

I can not see that this would ever be the behaviour you would want. If it had of exploded telling me the git call failed, it would have short-cut all of the problem finding.

This adds an explicit match on the static part of this git failure message and raises a `SystemExit` if it hits.  A test case that mocks such a situation is added.

Closes: https://github.com/pypa/setuptools-scm/issues/784

### DIFF
--- a/src/setuptools_scm/_file_finders/git.py
+++ b/src/setuptools_scm/_file_finders/git.py
@@ -22,6 +22,18 @@ def _git_toplevel(path: str) -> str | None:
         cwd = os.path.abspath(path or ".")
         res = _run(["git", "rev-parse", "HEAD"], cwd=cwd)
         if res.returncode:
+            # This catches you being in a git directory, but the
+            # permissions being incorrect.  With modern contanizered
+            # CI environments you can easily end up in a cloned repo
+            # with incorrect permissions and we don't want to silently
+            # ignore files.
+            if "--add safe.directory" in res.stderr and not os.environ.get(
+                "SETUPTOOLS_SCM_IGNORE_DUBIOUS_OWNER"
+            ):
+                log.error(res.stderr)
+                raise SystemExit(
+                    "git introspection failed: {}".format(res.stderr.split("\n")[0])
+                )
             # BAIL if there is no commit
             log.error("listing git files failed - pretending there aren't any")
             return None
--- a/testing/test_git.py
+++ b/testing/test_git.py
@@ -5,6 +5,7 @@ import os
 import shutil
 import subprocess
 import sys
+import textwrap
 
 from datetime import date
 from datetime import datetime
@@ -19,6 +20,7 @@ from unittest.mock import patch
 import pytest
 
 import setuptools_scm._file_finders
+import setuptools_scm._file_finders.git
 
 from setuptools_scm import Configuration
 from setuptools_scm import NonNormalizedVersion
@@ -861,3 +863,35 @@ def test_git_no_commits_uses_fallback_version(wd: WorkDir) -> None:
     assert str(version_no_fallback.tag) == "0.0"
     assert version_no_fallback.distance == 0
     assert version_no_fallback.dirty is True
+
+
+@pytest.mark.issue("https://github.com/pypa/setuptools-scm/issues/784")
+def test_dubious_dir(
+    wd: WorkDir, caplog: pytest.LogCaptureFixture, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """Test that we exit clearly if we are in a unsafe directory"""
+    wd.commit_testfile()
+    git_wd = git.GitWorkdir(wd.cwd)
+
+    def _run(*args, **kwargs) -> CompletedProcess:  # type: ignore[no-untyped-def]
+        """Fake "git rev-parse HEAD" to fail as if you do not own the git repo"""
+        stderr = textwrap.dedent(f"""
+        fatal: detected dubious ownership in repository at '{git_wd}'
+        To add an exception for this directory, call:
+
+            git config --global --add safe.directory /this/is/a/fake/path
+        """)
+        orig_run = run
+        if args[0] == ["git", "rev-parse", "HEAD"]:
+            return CompletedProcess(
+                args=[], stdout="%cI", stderr=stderr, returncode=128
+            )
+        return orig_run(*args, **kwargs)
+
+    monkeypatch.setattr(setuptools_scm._file_finders.git, "_run", _run)
+    with pytest.raises(SystemExit):
+        git_find_files(str(wd.cwd))
+
+    assert "fatal: detected dubious ownership in repository" in " ".join(
+        caplog.messages
+    )
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -622,6 +622,9 @@ would be required when not using `setuptools-scm`.
 - ✅ **Check**: Is setuptools-scm installed in your build environment?
 - ✅ **Check**: Are you in a valid SCM repository?
 
+**Problem: Error about dubious ownership**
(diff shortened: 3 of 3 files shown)


src/setuptools_scm/_file_finders/git.py
```diff
@@ -22,8 +23,18 @@ def _git_toplevel(path: str) -> str | None:
...
+            # with incorrect permissions and we don't want to silently
+            # ignore files.
+            if "--add safe.directory" in res.stderr:
+                log.error(res.stderr)
+                raise SystemExit(
+                    "git introspection failed: {}".format(res.stderr.split("\n")[0])
+                )
             # BAIL if there is no commit
-            log.error("listing git files failed - pretending there aren't any")
+            warnings.warn("listing git files failed - pretending there aren't any")
```
RonnyPfannschmidt: this should stay a log error

testing/test_git.py
```diff
@@ -861,3 +863,37 @@ def test_git_no_commits_uses_fallback_version(wd: WorkDir) -> None:
...
+
+    def _run(*args, **kwargs) -> CompletedProcess:  # type: ignore[no-untyped-def]
+        if args[0] == ["git", "rev-parse", "HEAD"]:
+            return CompletedProcess(
+                args=[], stdout="%cI", stderr=stderr, returncode=128
+            )
+        return orig_run(*args, **kwargs)
+
+    with pytest.raises(SystemExit):
+        with patch.object(setuptools_scm._file_finders.git, "_run", _run):
```
RonnyPfannschmidt: this should use monkeypatch not patch

RonnyPfannschmidt (changes requested):
thank you for bringing this up and preparing the details

this is indeed a problematic oversight

unfortunately this direly needed correction is also something that may possibly need a opt-out, as bringing it about may break pipelines that "seemed to work"



the new behavior you added is correct, unfortunately we have to account for prior accidental miss-use that's unable to correct timely

so lets add a env var to allow the misbehavior to continue for people unable to currectly fix their pipelines in a timely fashion

however if a git config call can fix the issue as well then we may just document that in the error message instead of adding a opt out env var

i just fear that some ci envs are enough of a mess the env var will still be needed

src/setuptools_scm/_file_finders/git.py
ianw: My thinking here was that this seems at least as important as,  perhaps slightly more important than

```
* Getting build dependencies for sdist...
/tmp/proj/.tox/.pkg/lib/python3.13/site-packages/setuptools_scm/git.py:202: UserWarning: "/tmp/proj" is shallow and may cause errors
  warnings.warn(f'"{wd.path}" is shallow and may cause errors')
```

can we consider making them the same level with the `warnings` call?

RonnyPfannschmidt: indeed ,this can be a log call as well - please note that theres significant change in the develop branch (splitting most of the implementations into the vcs-versioning package there=

testing/test_git.py
ianw: Done

ianw:
> unfortunately this direly needed correction is also something that may possibly need a opt-out, as bringing it about may break pipelines that "seemed to work"

Yes, I suspect you are right that people hit this and crafted unnecessary `MANIFEST.in` files.  I've added and documented an env var.

src/setuptools_scm/_file_finders/git.py
ianw: TBH I think these are probably too quiet; I'll leave this for now but perhaps surfacing the warnings better is food for thought on a new branch

ianw:
> however if a git config call can fix the issue as well then we may just document that in the error message instead of adding a opt out env var

The output I get in a failure case is 

```
/tmp/proj/.tox/.pkg/lib/python3.13/site-packages/setuptools_scm/git.py:202: UserWarning: "/tmp/proj" is shallow and may cause errors
  warnings.warn(f'"{wd.path}" is shallow and may cause errors')
/tmp/proj/.tox/.pkg/lib/python3.13/site-packages/setuptools/config/_apply_pyprojecttoml.py:82: SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
!!

        ********************************************************************************
        Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

        By 2026-Feb-18, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************

!!
  corresp(dist, value, root_dir)
running egg_info
writing src/proj.egg-info/PKG-INFO
writing dependency_links to src/proj.egg-info/dependency_links.txt
writing entry points to src/proj.egg-info/entry_points.txt
writing requirements to src/proj.egg-info/requires.txt
writing top-level names to src/proj.egg-info/top_level.txt
fatal: detected dubious ownership in repository at '/tmp/proj'
To add an exception for this directory, call:

	git config --global --add safe.directory /tmp/proj
git introspection failed: fatal: detected dubious ownership in repository at '/tmp/proj'

ERROR Backend subprocess exited when trying to invoke get_requires_for_build_sdist
```

There's like 3 levels of warnings (the setuptools_scm `warnings`, deprecations from `setuptools` which are louder than problems, the `log.warn` level stuff :)  But it stops and gives you a good clue what to do, so that's enough :)

RonnyPfannschmidt (approved):

PR #1250, opened 2025-12-13
Fix list rendering in docs/config.md

adamtheturtle:
Before (https://setuptools-scm.readthedocs.io/en/latest/config/):

image

After:

image

PR #1257, opened 2026-01-23
docs: Fix list & code block formatting on RtD

ferdnyc:
The docs' source formatting was triggering some quirks of RtD's MarkDown parser, things like:

1. It doesn't recognize a new bulleted list following body text (or a parent list item) unless there's a blank line preceding the first list marker. Without the blank, the list items get run into the preceding paragraph/parent list.
2. While it didn't necessarily affect formatting, each MarkDown heading is supposed to be followed by a blank line
3. Fenced code blocks with a `file="name"` attribute weren't showing the "name" string as the filename. Changed to `title="name"`, which is visible.

(**EDIT:** It seems this is a followup of sorts to #1250, which started the process of addressing these.)

### ======== SCREENSHOT: Before (1) ========

2, 3, and 4 are supposed to be a bulleted list child of list item 1. The ordered list should only have two items.

> image

### ======== SCREENSHOT: Before (2) ========

Each of these paragraphs is supposed to be an introductory line of text followed by a 2-item bulleted list.

> image

### DIFF
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -292,7 +292,7 @@ publish-release:
 The environment variable `SETUPTOOLS_SCM_OVERRIDES_FOR_${DIST_NAME}` must be set where:
 
 1. **`${DIST_NAME}`** is your package name normalized according to PEP 503:
-   
+
    - Convert to uppercase
    - Replace hyphens and dots with underscores
    - Examples: `my-package` → `MY_PACKAGE`, `my.package` → `MY_PACKAGE`
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -60,6 +60,7 @@ setuptools-scm provides the `no-local-version` local scheme and environment vari
 #### The Problem
 
 By default, setuptools-scm generates version numbers like:
+
 - `1.2.3.dev4+g1a2b3c4d5` (development version with git hash)
 - `1.2.3+dirty` (dirty working directory)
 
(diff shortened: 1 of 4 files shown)


ferdnyc:
Wow, the Mercurial tests seem _impossibly_ slow, the CI just grinds to a halt when it hits them.

ferdnyc:
Still not _quite_ there, the run-together text is fixed but this is still wrong: 

image


Perhaps the unordered child items are under-indented.

docs/integrations.md
```diff
@@ -290,6 +292,7 @@ publish-release:
 The environment variable `SETUPTOOLS_SCM_OVERRIDES_FOR_${DIST_NAME}` must be set where:
 
 1. **`${DIST_NAME}`** is your package name normalized according to PEP 503:
+
    - Convert to uppercase
    - Replace hyphens and dots with underscores
    - Examples: `my-package` → `MY_PACKAGE`, `my.package` → `MY_PACKAGE`
```
ferdnyc:
```suggestion
    - Convert to uppercase
    - Replace hyphens and dots with underscores
    - Examples: `my-package` → `MY_PACKAGE`, `my.package` → `MY_PACKAGE`
```

ferdnyc: Yup, that did it.

ferdnyc:
<s>These other pages suffer from the same issue, though:

* Usage
* Configuration
* Extending</s>

All now addressed, I believe.

:heavy_check_mark: Everything looks right to me, now, on all four pages affected.

RonnyPfannschmidt (approved):
thanks 👍

PR #1260, opened 2026-02-04
Allow for testing while building a Debian package

mr-c:
For the Debian package of `setuptools-scm` we set  `SETUPTOOLS_SCM_PRETEND_VERSION` and that interferes with the tests


RonnyPfannschmidt:
Thanks

PR #1270, opened 2026-02-24
Fixes #1263, clarifying [tool.setuptools_scm] requirement

### LINKED ISSUE #1263: Wrong statement in Pypi README
In the Pypi README it states:

The `[tool.setuptools_scm]` section is only needed if you want to customize
configuration options.

But this not true to my experience, because if I dont include it, my project gets always the version 0.0.0 regardless of tags.

bljoseph11:


RonnyPfannschmidt:
closing as wrong fix  #1267 will update the docs about all the details and version needs to correct the oversight

PR #1286, opened 2026-03-09
fix: Use missing-argument sentinel for dump_version argument

effigies:
#1183 added a warning that it claimed could be silenced by passing `scm_version=None`, but it uses `None` as its missing value sentinel, meaning the warning can never be silenced. This uses a sentinel object as a default value, so that passing `None` does successfully silence the warning and opt in to default behavior.


RonnyPfannschmidt:
the warning intentionally happens at missing parameter - the parameter is required
allowing it to be missing is a curtsey for downstream users - anyone using it should fix the code

effigies:
The warning says to pass `None` to silence the warning, but `None` is the sentinel used, so the warning cannot be silenced.

RonnyPfannschmidt:
You are right  thanks for calling out
out
Ill have to decide on a transition tineline

RonnyPfannschmidt:
I  can merge/ release after a changelog fragment is added

PR #1303, opened 2026-03-22
Filter out blobs when we unshallow a Git repo

clayote:
setuptools-scm only needs tags and parents; it doesn't need the actual content of any commit in a Git repository. This PR makes it only fetch the tags and parents, not the blobs, when `pre_parse = "fetch_on_shallow"`. This results in far less bandwidth usage, which can save lots of time -- potentially avoiding the timeout -- on repos with long histories, hosted cheaply.

### DIFF
--- a/vcs-versioning/changelog.d/1303.feature.md
+++ b/vcs-versioning/changelog.d/1303.feature.md
@@ -1,0 +1,3 @@
+The "fetch_on_shallow" option for Git worktrees no longer fetches the contents of Git commits--
+only their tags and refs.
+It takes hardly any bandwidth as a result, and is quite fast.
--- a/vcs-versioning/src/vcs_versioning/_backends/_git.py
+++ b/vcs-versioning/src/vcs_versioning/_backends/_git.py
@@ -12,6 +12,7 @@ from datetime import date, datetime, timezone
 from enum import Enum
 from os.path import samefile
 from pathlib import Path
+from subprocess import CalledProcessError
 from typing import TYPE_CHECKING
 
 from .. import _discover as discover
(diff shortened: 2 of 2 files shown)


RonnyPfannschmidt (changes requested):
Thanks for tackling this

Please add a changelog fragment 

Btw what's the minimal git version supporting this?

clayote:
> Btw what's the minimal git version supporting this?

I think version 2.32:

* ["git rev-list" learns the "--filter=object:type=<type>" option, which can be used to exclude objects of the given kind from the packfile generated by pack-objects.](https://github.com/git/git/blob/6e8d538aab8fe4dd07ba9fb87b5c7edcfa5706ad/Documentation/RelNotes/2.32.0.adoc#L98)

clayote:
Should I add a fallback to the old behavior when the user's on an old version of git?

Or I could make it use `--deepen=100` in a loop until it gets a usable output from `git describe`, in that case...

RonnyPfannschmidt:
its only relevant for rhel9 as far as i can tell, so fallback based on git version is the most sane

i'd avoid a deepen loop - it could backfire for pathological setups (like fork clone that lacks tags for a few years (yes i had one of those once, and nobody but me to blame))

clayote:
Instead of checking the version number, I've made it check the return code when it calls git with `--filter=blob:none`. The documentation is unclear on what subcommands accept `--filter` even today, so being flexible like this seems wise.

RonnyPfannschmidt (approved):
For now im going to accept this without integration tests 

I hope i never have to add a pipeline for old rhel

PR #1313, opened 2026-03-25
fix(ci): add griffecli for API stability check

RonnyPfannschmidt:
## Summary

Newer Griffe releases moved the CLI into a separate `griffecli` package. The `griffe` entry point then fails with:

```
ImportError: Please install `griffecli` to use 'griffe.main'
```

This broke the **API Stability Check** workflow step that runs `setuptools-scm/check_api.py` (see failing job on dependabot PR #1310).

## Changes

- Add `griffecli` to the setuptools-scm `test` dependency group (installed by `uv sync --all-packages --all-groups` in CI).
- Towncrier fragment `1310.misc.md`.

## Related

- Unblocks / complements #1310 (uv dependency bumps).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/setuptools-scm/changelog.d/1310.misc.md
+++ b/setuptools-scm/changelog.d/1310.misc.md
@@ -1,0 +1,1 @@
+Add `griffecli` to test dependencies so the API stability check keeps working after the Griffe CLI was split into a separate package.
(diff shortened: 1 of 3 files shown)


pawamoy:
Note that `griffe` depends on `griffecli`, so maybe it's just a matter of updating your lock file.

PR #1352, opened 2026-03-31
Add a timeout to a sudo-running tests

hroncok:
Or else it hangs forever when it asks for password/fingerprint.


hroncok:
I have a fingerprint configured for sudo, so maybe this is the reason I was seeing the hang? Is the devnull stdin normally making the command fail immediately?

RonnyPfannschmidt (approved):

PR #1386, opened 2026-06-14
add recommended_tag_format option

### LINKED ISSUE #1385: make it easier to skip over non-version tags
It's currently pretty easy to trigger the `packaging.version.InvalidVersion` error if you use tags for multiple purposes.

Consider [a demo repo](https://github.com/itcarroll/describe_pattern) that has this log:

```shell
2d73ca3 (HEAD -> main, tag: 2026-event, origin/main) setuptools
98c6a94 (tag: v0.1.0) initial commit
```

Calling `uvx setuptools-scm` raises the `InvalidVersion` error because `git describe --tags --match '*[0-9]*'` picks up a tag never meant to mark a version.

The error can be resolved by adding this configuration:

```
[tool.setuptools_scm.scm.git]
describe_command = "git describe --tags --match '[vV]*'"
```

Understanding why this fix is relevant goes beyond the casual git or setuptools-scm user's experience.

I'd like to suggest (and offer to contribute), an enhancement that either (or both):
- introduces a configuration option to toggle a match pattern that assumes the recommended tag scheme (starts with "v" or "V")
- makes the `setuptools-scm[simple]` approach default to the recommended tag scheme

itcarroll:
Closes #1385

Introduce a boolean `scm.git.recommended_tag_format` config option, with a default value of `False`.

When `True`, the `make_describe_command` is given `"*[vV][0-9]*"` and the result overrides the current describe command which is generated with `make_describe_command("*[0-9]*")`.

I'll mark as ready for review when the following are complete, but welcome comments anytime.

- [ ] consider FutureWarning to change default to True when `setuptools-scm[simple]` is in the build requirements
- [ ] include in docs
- [ ] changelog entry

### DIFF
--- a/vcs-versioning/src/vcs_versioning/_backends/_git.py
+++ b/vcs-versioning/src/vcs_versioning/_backends/_git.py
@@ -354,6 +354,8 @@ def version_from_describe(
 ) -> ScmVersion | None:
     if config.scm.git.describe_command is not None:
         describe_command = config.scm.git.describe_command
+    elif config.scm.git.recommended_tag_format:
+        describe_command = make_describe_command("*[vV][0-9]*")
 
     if describe_command is not None:
         if isinstance(describe_command, str):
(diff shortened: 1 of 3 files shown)


RonnyPfannschmidt:
This PR is superseded by the broader design in #1411, which introduces both `tag.prefix` and `tag.strict` as orthogonal, composable options. The `tag.strict = true` option is the generalized version of `recommended_tag_format = true` proposed here, and composes with `tag.prefix` for monorepo use cases.

PR #1392, opened 2026-06-17
Add `py.typed` marker file to `vcs-versioning`

eerovaher:
Without the marker file type checkers [are not allowed](https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information) to use type annotations in `vcs-versioning` when type-checking downstream packages. This is a bug because the annotations that now reside in `vcs-versioning` could be used for downstream type-checking with e.g. `setuptools-scm==9.2.2` which was released before the codebase was split.


RonnyPfannschmidt:
Thanks i missed that one

PR #1405, opened 2026-06-21
fix: replace git archive with git ls-files for file discovery (#662)

### LINKED ISSUE #662: git list_files hides freshly added files due to archive not knowing them yet

RonnyPfannschmidt:
## Summary

- Replace `git archive HEAD | tarfile` with `git ls-files -z --recurse-submodules -- . ':(exclude,attr:export-ignore)'` for file discovery
- Fixes freshly `git add`-ed files being invisible to setuptools file discovery (#662)
- Adds native submodule support with correct prefixed paths
- Preserves `export-ignore` honoring via pathspec attr magic

## Background

The archive approach was originally added due to strange git parsing bugs and to honor `export-ignore`. The new approach:

- **`-z`** eliminates parsing issues (NUL-separated output, no quoting)
- **`:(exclude,attr:export-ignore)`** preserves export-ignore filtering without tar parsing
- **`--recurse-submodules`** lists files inside submodules (which `git archive` never did)
- **Index state** means staged files are visible immediately (fixes #662)

## Test plan

- [x] Existing `test_git_archive_export_ignore` passes (export-ignore still honored)
- [x] New `test_git_added_but_uncommitted_files_visible` verifies #662 fix
- [x] New `test_git_export_ignore_with_staged_files` verifies export-ignore with staged files
- [x] New `test_git_submodule_files_listed` verifies submodule paths and export-ignore inside submodules
- [x] All 486 vcs-versioning tests pass
- [x] All 167 setuptools-scm integration tests pass
- [x] Pre-commit (ruff, mypy, codespell) passes

Made with [Cursor](https://cursor.com)

### DIFF
--- a/vcs-versioning/changelog.d/662.bugfix.md
+++ b/vcs-versioning/changelog.d/662.bugfix.md
@@ -1,0 +1,1 @@
+Replace git archive with git ls-files for file discovery, fixing freshly added files being hidden while still honoring export-ignore via gitattributes
--- a/vcs-versioning/src/vcs_versioning/_file_finders/_git.py
+++ b/vcs-versioning/src/vcs_versioning/_file_finders/_git.py
@@ -3,8 +3,10 @@ from __future__ import annotations
 import logging
 import os
 import subprocess
+from pathlib import Path
 
 from .. import _types as _t
+from .._backends._git import run_git
 from .._compat import norm_real, strip_path_suffix
 from .._integration import data_from_mime
 from .._run_cmd import run as _run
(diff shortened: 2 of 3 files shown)


MAProsper:
Hi, not sure if this is the correct to comment this, feel free to move it.

I used `git archive` as a reference on how to configure my repository for proper export.
The new `ls-files` method tries to approximate its behavior, but unexpectedly adds `--recurse-submodules`, which was never how `archive` worked.

This caused the source distribution of my package, which i carefully managed what to include, to suddenly explode to x45 its size, and made CI pipelines fail.

If including submodules is a desired divergence from `archive`, an option to disabled it and have it behave as previously (`ls-files` but without `recurse`), would be appreciated.

This behavior was specially surprising as the i have all submodules under a shared `vendor` directory, which is specifically tag with `export-ignore`.

Perhaps a better fix would be to do separate `ls-files` calls for submodules, but only if they appear in their parent's repository `ls-files -- . ":(exclude,attr:export-ignore)"`. That way instead of no submodules or all submodules, it can be controlled what submodules to include with the `export-ignore` attribute.

RonnyPfannschmidt:
Please open a new issue

PR #1455, opened 2026-06-27
doc: fix simple configuration example in README.md

tydeu:
`README.md` states that the `[tool.setuptools-scm]` section is optional in 8.1+. However, in 9.2+, this is no longer true. The section must either be present or `setuptools-scm` required with the `simple` extra. This PR updates the example to reflect that. It nows follows the example in the usage docs (with the `simple` extra and a `>=9.2` requirement).

This is the example that shows up on PyPI, so it is the first thing new users are likely to try. Thus, it is important `README.md` stays up to date. I took me a while to figure this out as a new user.


RonnyPfannschmidt:
thanks!

PR #1465, opened 2026-06-30
feat: add dynamic-metadata integration

henryiii:
This is a draft implementation of dynamic-metadata (0.3 version), to evaluate if this is something interesting to add. Also see https://discuss.python.org/t/dynamic-metadata-the-package/107953 and <https://dynamic-metadata.readthedocs.io> for more info. Currenly, only the upcoming scikit-build-core 1.0 supports `tool.dynamic-metadata`, but I'm also planning to contribute it to meson-python, and would like to see if other backends are intersted in supporting it.

This replaces the (now deprecated) `scikit_build_core.metadata.setuptools_scm` plugin. There is not a matching `dynamic_metadata.plugins.setuptools_scm` plugin.

I did an initial plan and commit with Claude (Opus 4.8), then removed about half of what it did that wasn't needed in the clean up commit (both commits are present here for inspection). I left the "integrations" section where it put it (first), but I can move it easily, just let me know where it should be moved.

Question: Should this hook be available at top level, so provider can just be "scm_versioning"?

### DIFF
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -1,5 +1,52 @@
 # Integrations
 
+## scikit-build / dynamic-metadata
+
+Any PEP 517 backend that supports
+[scikit-build/dynamic-metadata](https://github.com/scikit-build/dynamic-metadata)
+(such as scikit-build-core) can infer its version from VCS metadata using the
+`vcs_versioning.dynamic_metadata` provider shipped with vcs-versioning:
+
+```toml title="pyproject.toml"
+[build-system]
+requires = ["scikit-build-core", "vcs-versioning"]
+build-backend = "scikit_build_core.build"
+
+[project]
+name = "my-package"
+dynamic = ["version"]
+
+[tool.vcs-versioning]
+# normal vcs-versioning options go here, e.g.
+# local_scheme = "no-local-version"
+
+[[tool.dynamic-metadata]]
+provider = "vcs_versioning.dynamic_metadata"
+field = "version"
+```
+
+Version-scheme settings are read from `[tool.vcs-versioning]`. You may also pass
+options inline in the `[[tool.dynamic-metadata]]` table — any key besides `field`
+is forwarded as an override, so this is equivalent to the `local_scheme` above:
+
+```toml
+[[tool.dynamic-metadata]]
+provider = "vcs_versioning.dynamic_metadata"
+field = "version"
+local_scheme = "no-local-version"
+```
+
+Only the `version` field is supported. The provider also requests `vcs-versioning`
+as a build dependency via `get_requires_for_dynamic_metadata`, so listing it in
+`build-system.requires` is enough.
+
+!!! note
+
+    This targets the standalone dynamic-metadata package. scikit-build-core's
+    *built-in* metadata loader uses a different, older plugin signature and keeps
+    its own bundled provider; the `[[tool.dynamic-metadata]]` table shown here is
+    the standalone one.
+
 ## ReadTheDocs
 
 ### Avoid having a dirty Git index
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -1,10 +1,10 @@
 # Integrations
 
-## scikit-build / dynamic-metadata
+## dynamic-metadata
 
-Any PEP 517 backend that supports
-[scikit-build/dynamic-metadata](https://github.com/scikit-build/dynamic-metadata)
-(such as scikit-build-core) can infer its version from VCS metadata using the
+Any PEP 517 backend that supports the `tool.dynamic-metadata` array (as described in
+[scikit-build/dynamic-metadata](https://github.com/scikit-build/dynamic-metadata)),
+such as scikit-build-core 1.0+, can infer its version from VCS metadata using the
 `vcs_versioning.dynamic_metadata` provider shipped with vcs-versioning:
 
 ```toml title="pyproject.toml"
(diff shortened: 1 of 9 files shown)


vcs-versioning/src/vcs_versioning/_dynamic_metadata.py
```diff
@@ -0,0 +1,46 @@
+"""scikit-build ``dynamic-metadata`` provider for vcs-versioning.
```
RonnyPfannschmidt, line 1: Let's make this a private module unless we see a actual need to expose

vcs-versioning/src/vcs_versioning/dynamic_metadata.py
```diff
@@ -0,0 +1,46 @@
+"""scikit-build ``dynamic-metadata`` provider for vcs-versioning.
+
+Use this module as a provider for `scikit-build/dynamic-metadata
+<https://github.com/scikit-build/dynamic-metadata>`_ so any backend supporting
+that can fill in a VCS-derived ``version``::
+
+    [[tool.dynamic-metadata]]
+    provider = "vcs_versioning.dynamic_metadata"
```
RonnyPfannschmidt: Is there no entrypoint to register those?

vcs-versioning/src/vcs_versioning/_dynamic_metadata.py
henryiii: Okay, moved to an export in `__init__`. Doesn't trigger any additional imports (besides itself).

vcs-versioning/src/vcs_versioning/dynamic_metadata.py
henryiii: No, and that would be the wrong model since this is explicit, not discovered. Entrypoints are for automatic discovery. It follows the model of `build-system.build-backend`.

RonnyPfannschmidt: Then it's broken and leaks abstraction 

If its plugin based it ought to provide entrypoints

henryiii: Investigating...

henryiii: I think I do like it. PR at https://github.com/scikit-build/dynamic-metadata/pull/74. I'll probably merge it and test it out in scikit-build-core. It actually likely is a little nicer there, as we can separate new style (standardized) plugins and classic plugins based on if they are an entry point.

Downsides:

* Two plugins providing the same entry-point name is a hard error
    * Plugins are _strongly_ recommended to start their entry point with their package name.
* The loader that backends need to implement gets longer
* Local plugins now are an inline table, since they can't be an entry point, so reusing the string didn't make sense

Benefits:

* Can list installed plugins
* Can provide "did you mean" messages
* Plugins don't have to make anything exposed to user (your request)
* The name can be simpler than the path (`dynamic_metadata.plugin.regex` -> `dynamic_metadata.regex`)
* Plugins can switch to the object version without the user changing config

The last point is the selling point to me, it bothered me that wasn't possible before.

henryiii: 0.4 uses entry points. It works much better in scikit-build-core (mostly due to the fact we can now implement the backward-compatibility hook and the modern entry point with the same name).

I've updated to using them here. I'll try to post an example of scikit-build-core using this PR in a bit. This is the first time I've really interacted very much with a workspace, so seeing how to get this PR via pip/uv will be interesting.

vcs-versioning/src/vcs_versioning/_dynamic_metadata.py
RonnyPfannschmidt: thanks for introducing the entry-points in the downstream, not exporting it is so much cleanrer

RonnyPfannschmidt: nice work

vcs-versioning/src/vcs_versioning/dynamic_metadata.py
RonnyPfannschmidt: much better, thanks

docs/integrations.md
```diff
@@ -1,5 +1,40 @@
...
+
+The provider always populates `version`. Version-scheme settings are read from
+`[tool.vcs-versioning]`. You may also pass options inline in the
+`[[tool.dynamic-metadata]]` table — any key there is forwarded as an override, so
+this is equivalent to the `local_scheme` above:
+
+```toml
+[[tool.dynamic-metadata]]
+provider = "vcs_versioning"
+local_scheme = "no-local-version"
```
RonnyPfannschmidt: imho dynamic metadata should have a wy to pass in the root pyproject directly - and the vcs-versioning provider should use its own tool section for the configuration - creating additional placees for putting the same thing is against the zen of python as far as im concerned

henryiii: It does, hooks run in the same directory as build-system.build-backend so it's fine to read it. That's why the docs say both work. You can disable this (don't pass through config, require to be empty), but users will likely expect to have all the config in each tool.dynamic-metadata section. And if you use it multiple times (like building readme's out of fragments), you have to have it in each entry.

vcs-versioning/src/vcs_versioning/_dynamic_metadata.py
```diff
@@ -0,0 +1,49 @@
...
+    dist_name = project.get("name")
+    # dynamic-metadata runs hooks with cwd at the project root.
+    with vcs_versioning.overrides.GlobalOverrides.from_env(
+        "VCS_VERSIONING", dist_name=dist_name
+    ):
+        pyproject = vcs_versioning.PyProjectData.from_file("pyproject.toml")
+        version = vcs_versioning.infer_version_string(
+            dist_name=dist_name,
+            pyproject_data=pyproject,
+            overrides=dict(settings) or None,
```
henryiii, line 46: This is the line that makes `[[tool.dynamic-metadata]] <stuff>` work (other than provider, which is baked in).

vcs-versioning/src/vcs_versioning/_dynamic_metadata.py
```diff
@@ -0,0 +1,49 @@
...
+    project: Mapping[str, Any],
+) -> dict[str, Any]:
+    """Return the ``version`` field for a dynamic-metadata consumer."""
+
+    dist_name = project.get("name")
+    # dynamic-metadata runs hooks with cwd at the project root.
+    with vcs_versioning.overrides.GlobalOverrides.from_env(
+        "VCS_VERSIONING", dist_name=dist_name
+    ):
+        pyproject = vcs_versioning.PyProjectData.from_file("pyproject.toml")
```
henryiii, line 42: And this is the line that makes `tool.vcs-versioning` work :). Fully configurable. :)

pyproject.toml
```diff
@@ -14,6 +14,9 @@ dependencies = [
 [project.scripts]
 create-release-proposal = "vcs_versioning_workspace.create_release_proposal:main"
 
+[project.entry-points."dynamic_metadata.provider"]
+"vcs_versioning" = "vcs_versioning._dynamic_metadata"
```
henryiii: Ahh, I put this in the wrong pyproject.toml. Sorry! Will move soon. I wondered why I didn't see any other entry-points.

henryiii:
I did test this with scikit-build-core 1.0 some time ago, forgot to undraft.

RonnyPfannschmidt:
Merge plus release Monday?

PR #1470, opened 2026-07-09
fix: honor export-ignore for git submodules in file discovery (#1469)

### LINKED ISSUE #1469: git ls-files file discovery includes export-ignore submodules
Hi,

We used `git archive` as a reference on how to configure our repository for proper export.
The new `ls-files` discovery method introduced in PR #1405 tries to approximate its behavior, but unexpectedly adds `--recurse-submodules`, which was never how archive worked.

This caused the source distribution of our package, which we carefully managed what to include, to suddenly explode to x45 its size, and made some CI pipelines fail.

If including submodules is a desired divergence from archive, an option to disabled it and have it behave as previously (`ls-files` but without `recurse`), would be appreciated.

This behavior was specially surprising as we have all submodules under a shared `vendor` directory, which is specifically tag with `export-ignore`.

Perhaps a better fix would be to do separate `ls-files` calls for submodules, but only if they appear in their parent's repository `ls-files -- . ":(exclude,attr:export-ignore)"`. That way instead of no submodules or all submodules, it can be controlled what submodules to include with the `export-ignore` attribute.

apoorvdarshan:
Fixes #1469

## Root cause

`git ls-files` file discovery (introduced in #1405) is invoked with `--recurse-submodules`:

```python
run_git(
    ["ls-files", "-z", "--recurse-submodules", "--", ".", ":(exclude,attr:export-ignore)"],
    ...
)
```

`git archive` — the reference behavior this discovery method is meant to approximate — never descended into submodules. `--recurse-submodules` is a divergence: it lists files from each submodule's *own* index, and the `:(exclude,attr:export-ignore)` pathspec is evaluated against the submodule's own attributes, not against the parent's `export-ignore` marking on the gitlink path. As a result, a submodule placed under an `export-ignore`'d directory (e.g. a shared `vendor/`) is not excluded, and its entire contents get pulled into the sdist — which is what blew the reporter's sdist up to ~45x its intended size.

## Reproduction

```bash
# submodule source (contains sub_file.txt)
git init sub_src && (cd sub_src && git add . && git commit -m init)

# parent repo, submodule under an export-ignore'd vendor/ dir
git init parent && cd parent
printf '/vendor/ export-ignore\n' > .gitattributes
git add . && git commit -m init
git -c protocol.file.allow=always submodule add ../sub_src vendor/mysub
git commit -m 'add submodule under export-ignore vendor'
python -c "import vcs_versioning._file_finders as f, pprint; pprint.pprint(sorted(f.find_files('.')))"
```

**Before (wrong):** the export-ignore'd submodule leaks in —

```
['./.gitattributes', './.gitmodules', './parent_file.txt', './vendor/mysub/sub_file.txt']
```

`git archive HEAD | tar -t` on the same repo, by contrast, contains only `.gitattributes`, `.gitmodules`, `parent_file.txt`.

**After (correct):**

```
['./.gitattributes', './.gitmodules', './parent_file.txt']
```

## Fix

Drop `--recurse-submodules` from the `git ls-files` call in `vcs-versioning/src/vcs_versioning/_file_finders/_git.py`. This restores archive-equivalent behavior: submodules are listed as gitlinks and their contents are not descended into, so `export-ignore` on the submodule path is honored again. The change is a one-flag removal in a single discovery call; the `-z` NUL parsing and the `export-ignore` exclude pathspec that #1405 introduced are preserved, so the #662 fixes (staged-file visibility, safe parsing) are unaffected.

## Tests

- Added `test_git_export_ignore_submodule_excluded` (marked `@pytest.mark.issue(.../1469)`): a submodule under an `export-ignore`'d `vendor/` directory must not contribute any files to the listing. It fails on the unmodified code (`./vendor/mysub/sub_file.txt` present) and passes after the fix.
- Updated the former `test_git_submodule_files_listed` to `test_git_submodule_files_not_recursed`, asserting the restored archive-equivalent behavior (submodule contents are not recursed into). Both share a small `_make_submodule_source` helper.
- `uv run pytest vcs-versioning/testing_vcs/test_git.py` passes (the only failure, `test_git_getdate_signed_commit`, is a pre-existing GPG-signing environment issue that also fails on unmodified `main`). setuptools-scm regression/pyproject tests and the vcs file-finder/compat modules pass. `ruff check`, `ruff format`, `mypy --strict`, and `codespell` are clean.
- Added changelog fragment `vcs-versioning/changelog.d/1469.bugfix.md`.

Disclosure: prepared with AI assistance; reviewed and verified locally.

### DIFF
--- a/vcs-versioning/changelog.d/1469.bugfix.md
+++ b/vcs-versioning/changelog.d/1469.bugfix.md
@@ -1,0 +1,1 @@
+Stop recursing into git submodules during file discovery so that submodules marked ``export-ignore`` are excluded from sdists, restoring the pre-9.2 ``git archive`` behavior
--- a/vcs-versioning/src/vcs_versioning/_file_finders/_git.py
+++ b/vcs-versioning/src/vcs_versioning/_file_finders/_git.py
@@ -64,16 +64,18 @@ def _git_ls_files_and_dirs(
     toplevel: str, *, timeout: int | None = None
 ) -> tuple[set[str], set[str]]:
     # Use git ls-files with -z for NUL-separated output (safe parsing).
-    # --recurse-submodules lists files inside submodules with prefixed paths.
     # The exclude pathspec filters out files marked with the export-ignore
-    # gitattribute, matching the old git-archive behavior.
+    # gitattribute, matching the git-archive behavior.
     # "." is needed as positive pathspec for the exclude to apply against.
+    # Submodules are intentionally *not* recursed into: git archive never
+    # descended into submodules, and --recurse-submodules would bypass the
+    # parent's export-ignore attribute on the submodule path, wrongly pulling
+    # export-ignore'd submodules into the sdist (see #1469).
     # Uses run_git (--git-dir) to pin to the correct repository.
     res = run_git(
         [
             "ls-files",
             "-z",
-            "--recurse-submodules",
             "--",
             ".",
             ":(exclude,attr:export-ignore)",
--- a/vcs-versioning/testing_vcs/test_git.py
+++ b/vcs-versioning/testing_vcs/test_git.py
@@ -384,12 +384,8 @@ def test_git_export_ignore_with_staged_files(
     assert opj(".", "ignore.txt") not in found
 
 
-@pytest.mark.issue("https://github.com/pypa/setuptools-scm/issues/662")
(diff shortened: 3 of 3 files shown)


RonnyPfannschmidt:
Closing as ai slop

PR #1487, opened 2026-07-31
fix: correct typo in source comment

TheGreatApollyon:
Fixes: contanizered → containerized in _git.py


RonnyPfannschmidt (approved):
Nice catch

Issue #1422, opened 2026-06-22
Conflicting warnings regarding tag.strict


henriquegemignani:
With the latest version, if I don't have `tag.strict` set I get the new warning:
```
<string>:24: FutureWarning: tag.strict is not set. Currently defaults to False (permissive tag matching). In a future major version the default will change to True (require tags to contain a dot). Set tag.strict = true or tag.strict = false explicitly in your [tool.setuptools_scm] / [tool.vcs-versioning] config to silence this warning.
```

However, if I add either `tag.strict = true` or `tag.strict = false` I get the following warning instead

```
<string>:24: UserWarning: Both tag.prefix/tag.strict and scm.git.describe_command are set. The explicit describe_command takes precedence; tag.prefix and tag.strict will have no effect on the git describe match pattern.
```

The pyproject.toml in question: https://github.com/randovania/randovania/blob/38f1295393a973df21ac1cfae256733ebf7d7fb2/pyproject.toml#L209-L221

RonnyPfannschmidt:
Thanks for the report 

Seems i made a mistake with the logic there

PR #1494, opened 2026-08-09
fix: make describe command warnings coherent

### LINKED ISSUE #1429: Improve tag.strict and describe_command warning interaction
## Context

Reported in #1422: when a user has `scm.git.describe_command` set but `tag.strict` is not set, they get a `FutureWarning` telling them to set `tag.strict`. But when they do set it, they get a conflicting `UserWarning` saying `tag.strict` has no effect because `describe_command` takes precedence.

This creates an unwinnable situation for users.

## Desired behavior

The warnings should be coherent:

1. The `tag.strict` FutureWarning should **not** fire when `describe_command` is explicitly set (since strict has no effect in that case anyway).
2. The "both are set" UserWarning should only fire when the user explicitly opted into `tag.prefix`/`tag.strict` with the apparent intent of affecting describe behavior — not when they were just trying to silence warning 1.
3. Consider whether `tag.prefix` (when it differs from default) + `describe_command` should warn or just document the precedence.

## Current state

Both warnings are temporarily suppressed (downgraded to debug-level logging) in the fix for #1422 to avoid user confusion. This issue tracks re-introducing them with correct, non-conflicting logic.

Refs: #1422

yangfan-yf-yf:
Fixes #1429.

## Summary

- suppress the `tag.strict` future warning when a custom Git describe command supplies the matching behavior
- emit the conflict warning only for `tag.strict = true` or a non-empty `tag.prefix`
- evaluate deprecated `git_describe_command` values after normalization so both configuration forms behave consistently
- point configuration warnings at the external caller and document that `tag.prefix` still strips returned tags

## Validation

- hosted validation: 24 checks passed, covering Ubuntu and Windows on Python 3.8-3.13, PyPy 3.11, MSYS2, package builds, API stability, documentation, pre-commit, and changelog validation; 7 release-only jobs were conditionally skipped
- `vcs-versioning/testing_vcs/test_tag_config.py`: 33 passed
- core `vcs-versioning/testing_vcs`: 535 passed, 66 skipped, 1 xfailed, 2 failed
- the two failures require the POSIX `rm` and `ls` commands and reproduce unchanged on `aef602d8`
- Ruff 0.16.1 check and format check passed
- Mypy 2.3.0 strict check passed for the changed Python files
- codespell and `git diff --check` passed

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -84,7 +84,7 @@ strict = true      # require tags to contain at least one dot
 
     | Value   | Behavior |
     |---------|----------|
-    | unset (`None`) | Current permissive matching (`*[0-9]*`) with a `FutureWarning` advising you to set this explicitly. |
+    | unset (`None`) | Current permissive matching (`*[0-9]*`) with a `FutureWarning` advising you to set this explicitly, unless `scm.git.describe_command` supplies the describe command. |
     | `true` | **Strict** – tags must contain at least one dot (e.g. `*[0-9]*.*[0-9]*`).  Event-like tags such as `2026-event` are rejected. |
     | `false` | **Permissive** – matches any tag containing a digit (`*[0-9]*`), no warning. |
 
@@ -182,8 +182,10 @@ strict = true      # require tags to contain at least one dot
     !!! warning "Overrides tag.prefix / tag.strict"
 
         When `scm.git.describe_command` is explicitly set, `tag.prefix` and
-        `tag.strict` have no effect on the describe match pattern (a warning
-        is emitted).  The explicit command takes full precedence.
+        `tag.strict` have no effect on the describe match pattern. A non-empty
+        `tag.prefix` or `tag.strict = true` emits a warning; the explicit
+        permissive value `tag.strict = false` does not. `tag.prefix` is still
+        stripped from the tag returned by the custom command before parsing.
 
 `scm.git.pre_parse`
 :   A string specifying which git pre-parse function to use before parsing version information.
--- a/vcs-versioning/changelog.d/1429.bugfix.md
+++ b/vcs-versioning/changelog.d/1429.bugfix.md
@@ -1,0 +1,1 @@
+Avoid conflicting tag matching warnings when a custom Git describe command is configured.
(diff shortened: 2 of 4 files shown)


RonnyPfannschmidt:
Thanks for taking this on, and for the detailed validation notes.

I'm closing rather than iterating, because the parts I'd keep are smaller than the parts I'd drop.

**`_warning_stacklevel()` doesn't reach user code.** It only skips frames from `_config.py`, so every real entry point attributes the warning to library internals — `get_version()` reports `vcs_versioning/_get_version_impl.py:315`, an sdist build reports `vcs_versioning/_environment.py:250`. Before this change it pointed at `<string>:24`, the user's own frame, which is what #1422 quoted. It also breaks `filterwarnings(..., module=...)` for anyone trying to silence it. Widening the skip to the whole package doesn't rescue it either — it just lands one layer up in `setuptools_scm/_get_version.py`. The two-package layering means no local frame walk finds the user. This codebase uses plain `stacklevel=2` at every other `warnings.warn` and should keep doing so.

**Warning on a non-empty `tag.prefix` recreates #1422 for monorepo users.** With a custom describe command the prefix still does real work — stripping, as this PR's own `test_prefix_is_still_stripped_from_returned_tag` demonstrates — so `tag.prefix = "pkg-v"` plus `scm.git.describe_command` produces a warning with no escape hatch, unlike the explicit `tag.strict = false` you get for strictness. That's exactly point 3 of #1429, and the answer there should be "document, don't warn".

**Re-enabling the `tag.strict` FutureWarning is the biggest user-visible effect here** and the changelog fragment doesn't mention it. Released behaviour since 2.1.0 / 10.1.2 emits nothing; this brings the warning back for every user who hasn't set `tag.strict`. That needs to be a deliberate, separately announced call rather than a side effect of a bugfix.

Smaller things, for the record: Mercurial also consumes `tag.strict`/`tag.prefix`, so keying suppression off `scm.git.describe_command` silently drops the migration warning for hg users; `dataclasses.replace(config, ...)` in `_find_scm_in_parents` re-runs `__post_init__` and warns a second time; and the `strict=False` additions to the seven unrelated tests aren't needed — they pass without them.

#1429 stays open. The shape I'd merge: suppress the FutureWarning when a describe command is set (this PR gets that part right), warn only for `tag.strict = true`, keep `stacklevel=2`, and document that `tag.prefix` still strips.

Issue #1407, opened 2026-06-21
Deprecate and remove the global setuptools.file_finders entry point


RonnyPfannschmidt:
## Context

setuptools-scm registers a `setuptools.file_finders` entry point that is **always active** when the package is installed — even for projects that don't use setuptools-scm for versioning. This entry point calls `git_find_files()` → `_git_toplevel()`, which runs bare `git rev-parse` and can walk up to an outer/unrelated git repository.

This was the root cause of #353 (building a non-git package inside an outer git repo), and while the worst symptom (creating a `git archive` of the wrong repo) is gone, the entry point can still silently return wrong file lists.

## Current state

The modern setuptools-scm integration (`ScmEggInfoMixin`) already **bypasses** this entry point entirely when a workdir is available: it passes the validated workdir through to `list_tracked_files()`, which uses `--git-dir` pinning. So for projects that configure setuptools-scm, the entry point is redundant.

## Remaining exposure

The entry point still fires when:
1. setuptools-scm is installed in the build environment but the project being built doesn't configure it
2. The CLI's `--query files` path calls `find_files()` directly
3. Any other consumer of `walk_revctrl()` / `setuptools.file_finders`

## Proposed timeline

1. **Next minor release**: emit a deprecation warning when the `setuptools.file_finders` entry point is invoked and no setuptools-scm configuration is found for the project being built
2. **Following major release**: remove the entry point registration entirely

## Alternatives to full removal

- Keep the entry point but add a `samefile` guard in `_git_toplevel` so it returns `None` when the discovered toplevel doesn't match the requested path (prevents the outer-repo walk-up)
- Make the entry point a no-op unless an explicit opt-in is configured

Related: #353

PR #1504, opened 2026-08-19
Deprecate setuptools.file_finders for unconfigured projects; also emi…

karlhillx:
## Summary

* The first part of #1407: emit a `DeprecationWarning` when the global `setuptools.file_finders` entry point runs for a project that isn't configured to use setuptools-scm (`[tool.setuptools_scm]`, `setuptools-scm[simple]`, or `use_scm_version`)
* No change for configured projects; they already use the workdir API : (`ScmEggInfoMixin`) and won't warn.
* File discovery still works the same... the entry point stays registered for now and can be removed in a future major release

## Test plan

* [x] `setuptools-scm/testing_scm/test_file_finders.py`: warns when setuptools-scm is only a build dependency; no warning with `[tool.setuptools_scm]`, `use_scm_version`, or `setuptools-scm[simple]`
* [x] `uv run pytest -n12`
* [x] `pre-commit run --all-files`
* [ ] Confirm a real sdist for an unconfigured project with setuptools-scm installed still builds, includes tracked files, and emits the deprecation warning

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -380,11 +380,13 @@ These environment variables control setuptools-scm specific behavior.
 
 !!! warning "Setuptools File Finder Integration"
 
-    `setuptools-scm` automatically registers a setuptools file finder that includes all SCM-tracked files in source distributions. This behavior is **always active** when setuptools-scm is installed, regardless of whether you use it for versioning.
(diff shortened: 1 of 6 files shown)


RonnyPfannschmidt (commented):
this looks generally clean but i'd like to bikeshed a bit on the mechanism that picks when to warn

i'd like to ensure we have some way to just auto-disable the file finder hook in case the finalize options hooking succeeded

RonnyPfannschmidt:
thanks

PR #1529, opened 2026-09-14
skip if VersionInferenceNoOp

yunyiyun:
### Summary

When setuptools-scm is installed and `pyproject.toml` exists, `_infer_version_impl` unconditionally calls `_register_build_py_command(dist)` regardless of whether version inference is actually needed (`should_infer()` returns `False`). This wraps the user's custom `build_py` command class with `ScmVersionFileMixin` via `type()`, which can break the MRO and cause `check_package()` to fail when the user's class inherits from `distutils.build_py` instead of `setuptools.build_py`.

### Environment

- setuptools-scm: 10.3.0 (latest)
- setuptools: 82.0.1 / 83.0.0 / 84.0.0 (all affected)
- Python: 3.10 (also affects 3.11, 3.12, 3.13)

### Reproduction

**setup.py:**

```python
import os
from setuptools import setup
from setuptools.command.build_clib import build_clib
from setuptools.command.egg_info import egg_info

# Key: inherits from distutils.build_py, NOT setuptools.build_py
from distutils.command.build_py import build_py

BASE_DIR = os.path.dirname(os.path.realpath(__file__))

def create_dirs():
    for pkg in ["myproj", "mygen"]:
        os.makedirs(os.path.join(BASE_DIR, "build/packages", pkg), exist_ok=True)

class PythonPackageBuild(build_py, object):
    def run(self):
        create_dirs()                           # creates build/packages/myproj
        super().finalize_options()              # NOTE: calls finalize_options(), not run()

class EggInfoBuild(egg_info, object):
    def finalize_options(self):
        self.egg_base = os.path.relpath(os.path.join(BASE_DIR, "build/packages"))
        create_dirs()
        super().finalize_options()

setup(
    name="demo",
    version="1.0.0",
    packages=["myproj", "mygen"],
    package_dir={'': os.path.relpath(os.path.join(BASE_DIR, "build/packages"))},
    cmdclass={'build_py': PythonPackageBuild, 'egg_info': EggInfoBuild},
)
```

**pyproject.toml** (content is irrelevant — even an empty file triggers the issue):

```toml
[project]
name = "demo"
version = "1.0.0"
```

**Run:**

```bash
pip install setuptools-scm   # must be installed
python setup.py build        # fails
```

**Output:**

```
running build
running build_py
error: package directory 'build/packages/myproj' does not exist
```

**Uninstall setuptools-scm → no error:**

```bash
pip uninstall setuptools-scm -y
python setup.py build        # succeeds
```

#### 1. `_register_build_py_command` is called unconditionally

```python
# setuptools_scm/_integration/setuptools.py
def _infer_version_impl(dist, *, ...):
    pyproject_data = read_pyproject(...)          # succeeds if pyproject.toml exists
    result = _get_version_inference_config(...)   # may return VersionInferenceNoOp
    result.apply(dist)                             # may be a no-op

    _register_build_py_command(dist)   # ← unconditional, ignores result
    _register_egg_info_command(dist)
    _register_bdist_wheel_command(dist)
```

`should_infer()` can return `False` (no `[tool.setuptools_scm]` section, no `setuptools-scm[simple]` in `build-system.requires`), but `_register_build_py_command` is still called. The command wrapping is only useful when version files need to be written to the build directory, yet it happens regardless.

#### 2. `type()` wrapping creates MRO where `setuptools.build_py` precedes the user's class

```python
# _register_build_py_command
wrapped = type(
    "_SetuptoolsScmWrappedBuildPy",
    (ScmVersionFileMixin, project_build_py),  # project_build_py = PythonPackageBuild
    {},
)
dist.cmdclass["build_py"] = wrapped
```

The inheritance chains:

```
ScmVersionFileMixin  →  setuptools.build_py  →  distutils.build_py
PythonPackageBuild   →  distutils.build_py          (NOT setuptools.build_py)
```

C3 linearization produces:

```
[0] _SetuptoolsScmWrappedBuildPy
[1] ScmVersionFileMixin           ← run(): calls super().run()
[2] setuptools.build_py            ← run(): calls build_packages() → check_package(), does NOT call super().run()
[3] PythonPackageBuild             ← run(): creates directories — NEVER REACHED
[4] distutils.build_py
```

`ScmVersionFileMixin.run()` calls `super().run()`, which resolves to `setuptools.build_py.run()` (MRO position 2), not `PythonPackageBuild.run()` (MRO position 3). `setuptools.build_py.run()` calls `check_package()` which fails because `PythonPackageBuild.run()` (which creates the package directories) is never executed.

#### 3. Why `setuptools.build_py` comes before `PythonPackageBuild`

`setuptools.build_py` appears only in L1 (ScmVersionFileMixin's MRO), not in L2 (PythonPackageBuild's MRO). The C3 "good head" rule allows it to be taken freely from L1 before `PythonPackageBuild` is considered from L2.

If `PythonPackageBuild` inherited from `setuptools.build_py` instead of `distutils.build_py`, then `setuptools.build_py` would appear in L2's tail, blocking its removal from L1, and `PythonPackageBuild` would be taken first.

### Impact

Any project that:
1. Registers a custom `build_py` in `cmdclass`
2. Said class inherits from `distutils.command.build_py.build_py` (not `setuptools.command.build_py.build_py`)
3. Has `pyproject.toml` present (any content, even empty)
4. Has setuptools-scm installed in the environment

...will fail with `package directory '...' does not exist` during `build_py`, because the user's `run()` is skipped by the MRO.

This is not an edge case — inheriting from `distutils.build_py` is a common pattern in projects that predate setuptools' `build_py` extensions or that follow older tutorials.


RonnyPfannschmidt:
thansk for the pointer 

i did it slightyl different

ps you really shouldn't mix distutils and setuptools like that

PR #1537, opened 2026-09-15
fix: name the real env vars in the "jj not installed" error

hxperl:
## What is wrong

When `.jj/` is present but the `jj` command is not installed, `discover()` raises:

> Jujutsu (jj) repository detected at ... but the 'jj' command is not available. Install jj (https://jj-vcs.dev/docs/install), **set the `DISABLE_JJ=1` environment variable** to fall back to git, or remove the `.jj` directory if this is not a jj-managed workspace.

`DISABLE_JJ` is never read. `VcsEnvironment.from_env` calls `reader.read("DISABLE_JJ")`, and `EnvReader.read` only ever looks up `f"{tool}_{name}"` for each prefix in `tools_names` (plus the `_FOR_<DIST>` variants). There is no unprefixed fallback, so the variable the error tells you to set matches nothing.

The docs are right — `docs/config.md` and `docs/usage.md` both say `SETUPTOOLS_SCM_DISABLE_JJ` / `VCS_VERSIONING_DISABLE_JJ`. It is only this error message that is wrong, and it is the message a user sees at exactly the moment they need the correct name.

## Reproduction

On `main` (e218fb7). A colocated repo, tagged `v1.2.3`, with `jj` not on `PATH`:

```console
$ mkdir jjdemo && cd jjdemo
$ git init -q . && echo x > f.txt && git add f.txt && git commit -qm init && git tag v1.2.3
$ mkdir .jj
$ printf '[project]\nname = "demo"\n' > pyproject.toml
$ command -v jj || echo "jj: not on PATH"
jj: not on PATH
```

Following the error's advice does nothing; the documented name works:

```console
$ python -m setuptools_scm
('ERROR: no version found for', ...)

$ DISABLE_JJ=1 python -m setuptools_scm
('ERROR: no version found for', ...)

$ SETUPTOOLS_SCM_DISABLE_JJ=1 python -m setuptools_scm
1.2.3
```

And the message itself, from the same tree:

```console
$ python -c "
from pathlib import Path
from vcs_versioning._environment import VcsEnvironment
from vcs_versioning._backends._discover_vcs import discover
cfg = VcsEnvironment.from_env('SETUPTOOLS_SCM', env={}).build_config(root=Path('/tmp/jjdemo'))
try: discover(Path('/tmp/jjdemo'), config=cfg)
except LookupError as e: print(e)
"
Jujutsu (jj) repository detected at /tmp/jjdemo but the 'jj' command is not available. Install jj
(https://jj-vcs.dev/docs/install), set DISABLE_JJ=1 in the environment to fall back to git, or
remove the .jj directory if this is not a jj-managed workspace.
```

After this change, the same command prints:

```
Jujutsu (jj) repository detected at /tmp/jjdemo but the 'jj' command is not available. Install jj
(https://jj-vcs.dev/docs/install), set SETUPTOOLS_SCM_DISABLE_JJ=1 or VCS_VERSIONING_DISABLE_JJ=1
in the environment to fall back to git, or remove the .jj directory if this is not a jj-managed
workspace.
```

## The change

Derive the suggested names from `config.env.tool_names` rather than hardcoding them. That keeps `vcs-versioning` from naming `SETUPTOOLS_SCM` in its own source, and it means the message always lists exactly the prefixes this run would honour — including for a third integrator such as `HATCH_VCS`, which would otherwise be told the wrong name too.

The regression test extracts every `*_DISABLE_JJ=1` the message suggests and feeds each one back through `VcsEnvironment.from_env`, asserting it actually sets `disable_jj`. On `main` that test fails on the first assertion:

```
E       AssertionError: error suggests no env var: Jujutsu (jj) repository detected at ...
                        set the DISABLE_JJ=1 environment variable to fall back to git, ...
E       assert []
```

## What I verified, and what I did not

macOS arm64, CPython 3.14, `uv sync --all-packages --all-groups`:

- `uv run pytest -n12` — `12 failed, 822 passed, 74 skipped, 1 xfailed`. All 12 are in `setuptools-scm/testing_scm/` (`test_integration.py` build tests and one `test_setuptools_git.py` calver test) and all 12 fail identically on unmodified `main` in this environment — they are setuptools build-invocation failures (`Build from read-only source failed`), unrelated to `vcs-versioning`. Nothing in `vcs-versioning/testing_vcs/` fails.
- The new test passes with the change and fails without it (output above).
- `pre-commit run --files ...` on the three changed files — all hooks pass, `ruff` and strict `mypy` included.
- I do not have `jj` installed, so the jj tests that need a real jj binary were skipped here. The new test does not need one (it patches `has_command`, matching the existing `test_jj_missing_binary_errors`). CI will need to confirm with jj present, and on Linux/Windows.

## Judgement calls

- **Derived vs. hardcoded names.** The alternative is the literal string `SETUPTOOLS_SCM_DISABLE_JJ=1 (or VCS_VERSIONING_DISABLE_JJ=1)`, matching `docs/usage.md` word for word. That is a one-line diff and easier to read, at the cost of naming setuptools-scm from inside the standalone core and going stale for other integrators. I went with the derived form; say the word and I will switch it.
- **Wording.** `"set the X environment variable"` does not survive two names, so it became `"set X or Y in the environment"`. Open to any phrasing you prefer.
- **Scope.** I left the neighbouring `log.debug("jujutsu detected at %s but disabled via DISABLE_JJ", path)` alone. It is debug-only and reads as the name of the setting rather than an instruction, so it is not obviously wrong — but if you want it to match, it is a one-line follow-up.
- Changelog fragment filed under `vcs-versioning/` as `bugfix`, since that is the package that changed. `doc` was the other candidate; the string is user-facing behaviour, so `bugfix` seemed closer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019zkQEGBfqqomF5tanu383Q


RonnyPfannschmidt (approved):
good catch, thanks

